PR #16, opened 2015-10-19
Copy deployment/ to archive in distTar

gdearment:
- include the deployment/ directory in distTar to pick up other
  skylab based files if they exist


markelliot:
Could you add a check to the tests that ensures this is in the tar? (There should be something similar to what you need already...and I thought I actually did check the manifest made it.)

nmalaguti:
Should this exclude `manifest.yaml` since it would conflict with the generated version?

markelliot:
#22 adds the missing tests, so going to kill this.

PR #20, opened 2015-10-26
Convert `project.version` to a string when it is not a string

### LINKED ISSUE #18: distTar task fails when project.version is not a string
See https://github.com/nmalaguti/repro-distribution-flexversion for a repro. Running `./gradlew distTar` results in:

```
* What went wrong:
Execution failed for task ':createManifest'.
> No signature of method: java.lang.String.replaceAll() is applicable for argument types: (java.lang.String, com.palantir.gradle.versions.flexversioning.FlexVersion) values: [@serviceVersion@, master-1-g2c6266869ebb]
  Possible solutions: replaceAll(java.lang.String, java.lang.String), replaceAll(java.util.regex.Pattern, groovy.lang.Closure), replaceAll(java.util.regex.Pattern, java.lang.CharSequence), replaceAll(java.lang.CharSequence, java.lang.CharSequence), replaceAll(java.lang.CharSequence, groovy.lang.Closure), replace(char, char)
```

Since Gradle will convert the `project.version` object to a string when it needs to, I would expect the plugin to do the same.

nmalaguti:
Fixes #18

### DIFF
--- a/src/main/groovy/com/palantir/gradle/javadist/DistTarTask.groovy
+++ b/src/main/groovy/com/palantir/gradle/javadist/DistTarTask.groovy
@@ -27,7 +27,7 @@ class DistTarTask extends Tar {
 
     public void configure(DistributionExtension ext) {
         baseName = ext.serviceName
-        String archiveRootDir = ext.serviceName + '-' + project.version
+        String archiveRootDir = ext.serviceName + '-' + String.valueOf(project.version)
 
         from("${project.projectDir}/var") {
             into "${archiveRootDir}/var"
--- a/src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
+++ b/src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
@@ -62,10 +62,13 @@ class JavaDistributionPluginTests extends Specification {
 
         // try all of the service commands
         exec('dist/service-name-0.1/service/bin/init.sh', 'start') ==~ /(?m)Running 'service-name'\.\.\.\s+Started \(\d+\)\n/
+        sleep 5000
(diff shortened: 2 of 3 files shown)


src/main/groovy/com/palantir/gradle/javadist/DistTarTask.groovy
```diff
@@ -27,7 +27,7 @@ class DistTarTask extends Tar {
 
     public void configure(DistributionExtension ext) {
         baseName = ext.serviceName
-        String archiveRootDir = ext.serviceName + '-' + project.version
+        String archiveRootDir = ext.serviceName + '-' + project.version.toString()
```
markelliot: Prefer String,valueOf to toString calls as it deals with nulls correctly.

markelliot:
this is good to go but needs to be rebased (GitHub's branch protection requires each PR to be at HEAD of the merge target)

src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
```diff
@@ -62,10 +62,13 @@ class JavaDistributionPluginTests extends Specification {
 
         // try all of the service commands
         exec('dist/service-name-0.1/service/bin/init.sh', 'start') ==~ /(?m)Running 'service-name'\.\.\.\s+Started \(\d+\)\n/
+        sleep 5000
```
markelliot, line 65: I don't think these should be necessary.

nmalaguti, line 65: I tried to get the tests to pass 4 times on travis and they didn't pass until I added the sleeps. They aren't needed when running locally, but for some reason it keeps flaking on travis.

PR #22, opened 2015-10-29
Copy deployment/ to archive in distTar

stevenfan:
include the deployment/ directory in distTar to pick up other skylab based files if they exist

### DIFF
--- a/src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
+++ b/src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
@@ -159,6 +159,37 @@ class JavaDistributionPluginTests extends Specification {
         manifest.contains('productVersion: 0.1\n')
     }
 
+    def 'produce distribution bundle with files in deployment/' () {
+        given:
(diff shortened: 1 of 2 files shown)


src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
```diff
@@ -62,10 +62,13 @@ class JavaDistributionPluginTests extends Specification {
 
         // try all of the service commands
         exec('dist/service-name-0.1/service/bin/init.sh', 'start') ==~ /(?m)Running 'service-name'\.\.\.\s+Started \(\d+\)\n/
+        sleep 5000
```
markelliot, line 65: Have you actually found this to be flakey? The sleeps shouldn't be necessary as the shell execution waits for the program to actually run, and the program is just a long sleep anyway.

stevenfan, line 65: Yes: my first commit without this failed and looking at the Travis builds it looks like PR#20 went through the same thing.

PR #27, opened 2015-12-09
Feature/add additional jvm params at runtime through an optional wrapper.conf file

viveklak:
The location for the wrapper conf is currently in $SERVICE_HOME/bin. Not convinced on the location but that can be changed later.

Note - we may not actually need this as $JAVA_OPTS etc. can be set manually to add additional args. Leaving this here for archival purposes - in case we need to revive it later.

NB: I didn't add support for this in the windows startup scripts yet.


markelliot:
Going to decline this since I think this is covered by JVM_ARGS settings that override default parameters anyway.

PR #29, opened 2015-12-11
add `init.sh console` to run in foreground

bmoylan:
Addresses https://github.com/palantir/gradle-java-distribution/issues/26

### DIFF
--- a/src/main/resources/init.sh
+++ b/src/main/resources/init.sh
@@ -93,11 +93,24 @@ stop)
         exit 0
     fi
 ;;
+console)
+    service/bin/init.sh status > /dev/null 2>&1
+    if [[ $? == 0 ]]; then
+        echo "Process is already running"
+        exit 1
+    fi
+    trap "service/bin/init.sh stop &> /dev/null" SIGTERM EXIT
+    mkdir -p "$(dirname $PIDFILE)"
+
+    service/bin/$SERVICE $ARGS &
+    echo $! > $PIDFILE
+    wait
+;;
 restart)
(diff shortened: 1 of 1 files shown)


src/main/resources/init.sh
```diff
@@ -93,11 +93,14 @@ stop)
         exit 0
     fi
 ;;
+console)
+    service/bin/$SERVICE $ARGS
```
markelliot: I think you probably ought to check to see if the service is running and failing if it's running already. I also suspect you want to write a pid file and attach to the process rather than just execute the command, in particular because that allows start/stop to fail if the process is being run from console.

markelliot:
I think you'll need to find a way to test this using the test framework before we push this through.

bmoylan:
Thanks for the feedback, will do

bmoylan:
I just added logic to save the pid and use `init.sh stop` to exit, cleaning up. Working on tests, but having trouble with groovy/shell interaction to background the `console` process and kill it with `stop`, `kill`, etc.

markelliot:
Cool. Can you just assert here that you manually tested this and it's functional, and then I think we can merge.

Unit testing shell scripts is a pain, especially this kind of thing, I know.

bmoylan:
Totally. I manually tested killing with the following, and pidfile is cleaned up in all cases. `status` returns correct output and exit code while `console` is running in different terminal. 
- `Ctrl-C` in foreground
- `init.sh stop`
- `kill`

markelliot:
sgtm

PR #33, opened 2015-12-17
use jvm options in run task

bavardage:


bavardage:
(I did this in the browser, made a mistake :P)

PR #43, opened 2016-01-24
Support extra settings in service/bin

### LINKED ISSUE #42: Generate and source service/bin/config.sh as part of init.sh execution
- Generate service/bin/config.sh as part of script generation.
- Add `javaHome` config option and emit `JAVA_HOME="<javaHome>"` into config.sh as part of generation

markelliot:
fixes #42

### DIFF
--- a/src/main/resources/config.sh
+++ b/src/main/resources/config.sh
@@ -1,0 +1,5 @@
+# override the JAVA_HOME value with a known JRE location
+@javaHome@
+
+# use JVM_OPTS to set JVM settings, such as minimum and maximum heap
+#JVM_OPTS="-Xms1g -Xmx1g"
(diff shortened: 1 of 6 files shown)


src/main/resources/config.sh
```diff
@@ -0,0 +1,5 @@
+# override the JAVA_HOME value with a known JRE location
+@javaHome@
+
+# use JVM_OPTS to set JVM settings, such as minimum and maximum heap
+#JVM_OPTS="-Xms1g -Xmx1g"
```
bavardage, line 5: wait but how do we actually set the JVM_OPTS?

markelliot, line 5: The idea is that this file is easy to edit. If you want to set default JVM opts can do it through the normal config in gradle.

bavardage, line 5: aha yes - may well not be deploying in a managed way - could just be extracting the generated tgz

bavardage:
:+1:

PR #51, opened 2016-02-16
[WIP] Kill whole group, not just parent

### LINKED ISSUE #50: Stop command should kill whole process group
If the top-level (daemonized) service starts one or more long-running child processes, it has to ensure that those child processes get killed when the parent process is killed. This is somewhat painful to do in Java because it involves setting a shutdown hook, and it's repetitive work that's easy to get wrong. If done incorrectly, the child processes become zombies and have to be reaped manually by other means; they could take up system resources such as memory and ports.

Instead, if `init.sh` were to record the process group ID and `kill` the process group, we'd have a clean solution that works without additional work from the app.

Tagging @mingyukim, @tomkozlowski and @gdearment for comments.

punya:
Fixes #50.

### DIFF
--- a/src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
+++ b/src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
@@ -279,6 +279,38 @@ class JavaDistributionPluginTests extends Specification {
             .find({it.name.endsWith("-manifest-classpath-0.1.jar")})
     }
 
+    def 'kills children spawned by service'() {
+        given:
+        createUntarBuildFile(buildFile)
+
+        temporaryFolder.newFolder('src', 'main', 'java', 'test')
+        temporaryFolder.newFile('src/main/java/test/Test.java') << '''
+        package test;
+        public class Test {
+            public static void main(String[] args) throws Exception {
+                Runtime.exec("yes");
+                while(true);
+            }
+        }
+        '''.stripIndent()
+
+        when:
+        BuildResult buildResult = run('build', 'distTar', 'untar').build()
+
+        then:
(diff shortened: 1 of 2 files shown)


src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
```diff
@@ -279,6 +279,38 @@ class JavaDistributionPluginTests extends Specification {
...
+        buildResult.task(':distTar').outcome == TaskOutcome.SUCCESS
+        buildResult.task(':untar').outcome == TaskOutcome.SUCCESS
+
+        // check content was extracted
+        new File(projectDir, 'dist/service-name-0.1').exists()
+
+        // try all of the service commands
+        String output = exec('dist/service-name-0.1/service/bin/init.sh', 'start')
+        exec('dist/service-name-0.1/service/bin/init.sh', 'stop')
+        exec("ps aux | grep '[y]es'").trim() == ""
```
markelliot, line 311: the `[y]` here is to distinguish the grep command from finding itself?

punya, line 311: Yeah, I found this trick on a Stackoverflow page.

markelliot:
In principle I'm fine with this, but the tests are failing =/

markelliot:
@punya still interested in pushing this through?

punya:
Yes but I won't be able to work on it for a few more days.

markelliot:
@punya I'm going to close this for now, feel free to reopen if you'd like to contribute the change to init.sh. @uschi2000 is working on a go-based launcher to clean up #73 and I think we can account for process group as part of that.

Issue #52, opened 2016-02-16
Avoid using afterEvaluate { } to configure the distTar task


henryptung:
Some Gradle plugins depend, for example, on the `distTar` task output filename to resolve against artifacts or publish locations. (For example, pt-almanac.) Since `gradle-java-distribution` adds an `afterEvaluate` to configure the output name, any output filenames fetched during normal project evaluation are not reliable, and will cause those plugins to break in non-obvious ways at runtime (see HIGHLINE-1091).

The workaround, as it exists currently, is to move the publication resolution strategy (or the entire `almanacPublish` task) to an `afterEvaluate`, but I find this difficult to maintain for the same reason `_.defer` isn't desirable in JS code. If, for example, the Almanac task is placed in a shared Gradle script, depending on it or modifying its configuration would probably require a _nested_ `afterEvaluate` block.

I'd suggest refactoring the extension instead to mutate variables/properties synchronously, so that after the extension is reconfigured, the project is in a consistent/complete state and ready for further configuration. As an even stronger notion, the extension can be a call-me-once method, such that all necessary configuration is collected from the closure, and the run/dist tasks are created in immutable-state fashion from the one-time configuration closure.

markelliot:
Things used to work that way, but because of other ordering issues, and because the plugin doesn't receive state purely through injection, I'm reasonably sure we can't accommodate this.

markelliot:
Also, we can probably point you to several examples where you do not need to put your artifactoryPublish block into an afterEvaluate call, ping me directly for those.

henryptung:
Sorry, not really sure I understand the context of the other ordering issues/non-injected state you're speaking of, but let me try to prototype what I'm talking about and submit a PR.

Other examples along these lines would also be helpful (in particular, ones where the project name and the archive name differ, e.g. project `raygun-server`, application name `raygun`).

markelliot:
Fixed in 0.5.1.

PR #55, opened 2016-02-19
Configure tasks synchronously

henryptung:
Targets #52.

Test `build.gradle`:

``` groovy
distribution {
    serviceName "my-service"
    mainClass "dummy.service.MainClass"
    args "hello"
}

dependencies {
    compile "com.palantir.gradle.javadist:gradle-java-distribution:$javaDistributionVersion"
}

println "Class-Path: ${manifestClasspathJar.manifest.attributes["Class-Path"]}"
println "distTar output: ${distTar.outputs.files.singleFile}"

afterEvaluate {
    println "Class-Path (afterEvaluate): ${manifestClasspathJar.manifest.attributes["Class-Path"]}"
    println "distTar output (afterEvaluate): ${distTar.outputs.files.singleFile}"
}
```

With 0.5.0:

```
Class-Path: null
distTar output: /home/htung/dev/sandbox/build/distributions/sandbox.tgz
Class-Path (afterEvaluate): gradle-java-distribution-0.5.0.jar sandbox.jar
distTar output (afterEvaluate): /home/htung/dev/sandbox/build/distributions/my-service.tgz
```

After patch:

```
Class-Path: gradle-java-distribution-0.5.0-4-gd49c1de.dirty-sources.jar gradle-java-distribution-0.5.0-4-gd49c1de.dirty.jar sandbox.jar
distTar output: /home/htung/dev/sandbox/build/distributions/my-service.tgz
Class-Path (afterEvaluate): gradle-java-distribution-0.5.0-4-gd49c1de.dirty-sources.jar gradle-java-distribution-0.5.0-4-gd49c1de.dirty.jar sandbox.jar
distTar output (afterEvaluate): /home/htung/dev/sandbox/build/distributions/my-service.tgz
```

### DIFF
--- a/src/main/groovy/com/palantir/gradle/javadist/DistTarTask.groovy
+++ b/src/main/groovy/com/palantir/gradle/javadist/DistTarTask.groovy
@@ -44,8 +44,7 @@ class DistTarTask extends Tar {
         }
 
         into("${archiveRootDir}/service/lib") {
-            from(project.tasks.jar.outputs.files)
-            from(project.configurations.runtime)
+            from(ext.classpath)
         }
 
         if (ext.isEnableManifestClasspath()) {
--- a/src/main/groovy/com/palantir/gradle/javadist/DistributionConvention.groovy
+++ b/src/main/groovy/com/palantir/gradle/javadist/DistributionConvention.groovy
@@ -20,7 +20,7 @@ class DistributionConvention {
             throw new IllegalStateException("Cannot configure distribution twice!")
         }
         configured = true
-        DistributionExtension ext = new DistributionExtension()
+        DistributionExtension ext = new DistributionExtension(project)
         project.configure(ext, closure)
 
         // Specify classpath using pathing jar rather than command line argument on Windows, since
--- a/src/main/groovy/com/palantir/gradle/javadist/DistributionConvention.groovy
+++ b/src/main/groovy/com/palantir/gradle/javadist/DistributionConvention.groovy
@@ -1,0 +1,104 @@
+package com.palantir.gradle.javadist
+
+import org.gradle.api.Project
+import org.gradle.api.Task
+
+import java.nio.file.Paths
+
+class DistributionConvention {
+    private static final String GROUP_NAME = "Distribution"
+
+    private Project project
+    private boolean configured = false
+
+    DistributionConvention(Project project) {
+        this.project = project
+    }
+
+    void distribution(Closure<?> closure) {
+        if (configured) {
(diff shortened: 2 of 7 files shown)


markelliot:
According to https://discuss.gradle.org/t/whats-the-difference-between-plugin-conventions-vs-plugin-extensions/5523 conventions are no longer recommended. Perhaps there's a fix here which doesn't involve such a large change.

markelliot:
e.g. I suspect this whole thing is fixable by making the classpath computation lazy (by deferring it as a lambda)

henryptung:
Er, the classpath computation is lazy in the new impl, but that's not the primary problem here; the configuration of the distTar name is.

The reason conventions were deprecated is namespacing concerns, where method names could collide between plugins. In this case, the distribution extension already made `distribution` a closure-taking method, which could collide with other plugins; this implementation should have no more API collision surface area than the original.

markelliot:
why isn't this solved by just moving `distTar.configure(ext)` then?

henryptung:
Where would you suggest moving it in this case? I tried looking for a reasonable place to put it, but plugin application is too early (the `DistributionExtension` is not configured yet), and given the current API, I'm not sure you can schedule anything to run after the `distribution { }` closure automatically.

I chose instead to move code in bulk; there's very little actual change to the code; it just moves task creation somewhat later (plugin application time -> post-`distribution { }` time) and moves task configuration earlier (`afterEvaluate` -> post-`distribution { }` time), to minimize behavior change.

henryptung:
From a more abstract POV, Gradle configuration is moving towards [rule-based model configuration](https://docs.gradle.org/current/userguide/software_model.html). That model seems to be based around immutable state, with a dependency resolver that schedules execution of configuration blocks based on their state dependencies.

But without that resolver/scheduler (which doesn't seem to exist yet), there's not a way to automagically schedule my `almanacPublish` task configuration after `distTar` task configuration; at least, if `distTar` task configuration is done as early as possible (during project evaluation), I can just move my task declaration below that point, instead of moving it into `afterEvaluate` (which in turn would force anything depending on that task into `afterEvaluate`, etc.).

markelliot:
I'm still a bit puzzled why this shows up for you, in my almanacPublish blocks I'm just referencing `artifact distTar`, and I get the right names.

markelliot:
I'm going to close this in favor of #56

markelliot:
...if you want to contribute the configurable classpath thing you're welcome to do so, but obv. doesn't need the rest of the change on this PR.

PR #56, opened 2016-02-20
Ensure that the distTar artifact name is available without an afterEvaluate block

markelliot:


markelliot:
@henryptung

henryptung:
Semantically, it might make sense to override `setBaseName` to throw in this case, but in any case my build will work. :+1:

henryptung:
Link #52

Issue #59, opened 2016-03-16
Wait longer for process to stop


dpaq:
init.sh stop has `sleep 4` right now; instead could loop for awhile checking

dpaq:
Don't loop forever though!

markelliot:
Released in 0.5.2

PR #60, opened 2016-03-16
check every second up to a max count for the process to stop

dpaq:
Addresses issue #59


dpaq:
I wasn't sure how long we wanted to wait so I picked 240 seconds (approx; there is some slight overhead). Checks once a second to see if we are done and prints to the user every 5 seconds that it's still waiting.

PR #62, opened 2016-03-23
Add startDelay property to DistributionExtension

henryptung:
Configures wait delay before checking liveness of the application
process.

### DIFF
--- a/src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
+++ b/src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
@@ -23,6 +23,7 @@ class DistributionExtension {
     private List<String> defaultJvmOpts = []
     private boolean enableManifestClasspath = false
     private String javaHome = null
+    private int startDelay = 1
 
     public void serviceName(String serviceName) {
         this.serviceName = serviceName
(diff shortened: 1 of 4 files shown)


markelliot:
I think it'd be better to do this with a fixed maximum wait for start time, similar to the way Dan modified the fixed maximum wait for stop time. See #59.

henryptung:
Agreed, there's an easy gate for figuring out whether a process is stopped via PID existence. Do you have a similar analogue for startup in mind?

markelliot:
The is_process_active function used in the stop loop can also be used in the start loop, it checks the process table using the PID file, which is created before the current sleep call.

henryptung:
Ah, sorry, I should have explained the use case first, my bad.

This PR targets the case where service startup requires initializing or checking some backend, and may fail N seconds after the Java process is created.

If I understand correctly, `is_process_active` checks if the process exists, but I'm trying to check beyond "Java PID exists", to "was configuration parsed correctly" and "are backends alive + properly initialized".

Not sure if that's within scope of this script, but "process starts but exits very soon after" is probably fairly common (malformed configuration).

markelliot:
Ah, makes sense. Could we just add an optional 2nd check which checks the process lives for at least N seconds after it starts, then?

uschi2000:
Closing due to inactivity. Feel free to reopen.

PR #64, opened 2016-04-05
Fix Gradle IDEA task so that unit tests can be run in IntelliJ

nmiyake:


nmiyake:
This is mostly for convenience of deving. Applied this to my own Gradle project, and sharing in case there's interest. Would probably be cleaner as a plugin, but not sure if it's worth the overhead to write/distribute.

markelliot:
Is this just fixing dumb ordering in some dependencies block?

nmiyake:
Yeah, see https://discuss.gradle.org/t/nosuchmethoderror-in-testkit-after-2-9-2-10-transition/13505. This is a bug that the Gradle dev team has known about since 2.10, but is still not fixed. If 'gradle-core' is on the classpath before 'gradle-test-kit' in IDEA, debugging in IDEA will fail due to classpath issues unless you re-order the dependencies. However, because 'gradle-test-kit' is a test dependency, by default it will always come after 'gradle-core'.

Not a huge issue, but can be confusing to new devs on a project (unit tests fail with a "No such method" runtime exception in IDEA even though they succeed when run via Gradle), and the fix isn't obvious. Although it can be fixed by manually editing the classpath order in IDEA, the process is actually a bit of a pain due to the way the UI is set up.

PR #67, opened 2016-04-15
fix parsing jvm opts

### LINKED ISSUE #66: Incorrect default variable name in config.sh
Currently [this line](https://github.com/palantir/gradle-java-distribution/blob/develop/src/main/resources/config.sh#L5) mentions setting Java options using the `JVM_OPTS` environment variable, but the artifacts built using the plugin look for a variable named `JAVA_OPTS` (see [this line](https://github.com/palantir/gradle-java-distribution/blob/623cf9dc5753eced09a2c0f488035f9d33255324/gradlew#L161)).

vipshek:
Fixes #66

### DIFF
--- a/gradlew
+++ b/gradlew
@@ -6,7 +6,7 @@
 ##
 ##############################################################################
 
-# Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
+# Add default JVM options here. You can also use JVM_OPTS and GRADLE_OPTS to pass JVM options to this script.
 DEFAULT_JVM_OPTS=""
 
 APP_NAME="Gradle"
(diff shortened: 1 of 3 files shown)


rimig:
+1 was just hitting this as well

rimig:
Should we also change config.sh to:

`#export JVM_OPTS="-Xms1g -Xmx1g"`

I haven't been able to get this working without exporting in the distributions I'm using this in.

sixinli:
+1 on adding the `export`

markelliot:
The export won't do anything as `config.sh` is sourced from init.sh.

rimig:
@markelliot but then init.sh calls SERVICE_CMD which won't have the JVM_OPTS variable when it's actually the one that's launching java so it needs it

gradlew
```diff
@@ -6,7 +6,7 @@
 ##
 ##############################################################################
 
-# Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
+# Add default JVM options here. You can also use JVM_OPTS and GRADLE_OPTS to pass JVM options to this script.
```
markelliot, line 9: this should remain as `JAVA_OPTS` (it's a generated file)

PR #69, opened 2016-05-02
Upgrade gradle

nmiyake:


src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
```diff
@@ -341,7 +340,7 @@ class JavaDistributionPluginTests extends Specification {
 
     private GradleRunner run(String... tasks) {
         GradleRunner.create()
-            .withPluginClasspath(pluginClasspath)
+            .withPluginClasspath()
```
markelliot, line 343: oooh, much nicer

markelliot:
Thanks!

src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
nmiyake, line 343: Yup -- unfortunately it still requires a Gradle task to be run before it will work in an IDE (the `pluginUnderTestMetadata` task), but at least that task is provided by the official Gradle development plugin and there's less futzing in the actual code.

PR #70, opened 2016-05-02
Replace idea-test-kit-fix script with plugin

nmiyake:


markelliot:
Is this a dup of #65?

nmiyake:
Yup, this is a dupe -- forgot about that one :). Will close this one and update the other.

PR #71, opened 2016-05-18
Exclude var/data from distTar

sfackler:
Same rationale as #21 

This is assuming that `var/data` is the right place for the service to place persistent data, in contrast to `var/run` which is for stuff that should only persist until the service is stopped - is that right?

### DIFF
--- a/src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
+++ b/src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
@@ -25,6 +25,7 @@ class DistributionExtension {
     private List<String> defaultJvmOpts = []
     private boolean enableManifestClasspath = false
     private String javaHome = null
+    private List<String> excludeFromVar = []
 
     public void serviceName(String serviceName) {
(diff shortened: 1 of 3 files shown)


markelliot:
var/data is the right place for persistence, but I don't think can be default excluded -- what if a service needs to package an initial data state?

Instead, do you want to add a "excludeFromVar" config item?

sfackler:
Ah, good point - updated.

src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
```diff
@@ -25,6 +25,7 @@ class DistributionExtension {
     private List<String> defaultJvmOpts = []
     private boolean enableManifestClasspath = false
     private String javaHome = null
+    private List<String> excludeFromVar = ["log", "run"]
```
markelliot: Groovy Strings -> Strings. Also, I think we should consider _always_ not packaging these, so probably want to do the list adjustment here or leave existing excludes.

PR #72, opened 2016-05-18
resolve runtime configuration later

rjullman:
Configurations should be resolved during gradle's configuration phase, but were inadvertently being resolved in gradle's initialization phase.


rjullman:
I'm relatively new to gradle, but I think this fix is in the right direction.  

Another idea is to have another task that just resolves the runtime dependencies and then have everything else depend on the output of that task.  Then we can go back to having the code not run in doFirst blocks.

rjullman:
Also, the jars I added are just empty jars. I can probably just make them simple text files -- I bet gradle doesn't care what kind of dependencies they are.

tpetracca:
@markelliot I'm not sure who owns this.  Mind suggesting a reviewer/committer?

markelliot:
I'm the RP. Can you guys link me to the thing you're trying to solve? It seems weird that you should need to change the way this stuff works. Configuration already happens in an `afterEvaluate` block.

rjullman:
Trying to get https://github.com/palantir/atlasdb/pull/463 working, which will add atlas CLIs to an existing project that is already using this plugin.  We need to able to add additional runtime jars from the plugin and this is not allowed if the configuration is already resolved.  We let the user specify the atlas version so this needs to be done in our project after evaluation block which can run after this plugins afterEvaluate block.  I'm also working on a different solution that won't need this change.

markelliot:
Ok, the placement of the doFirst stuff seems pretty sketchy to me. @rfink, do you have thoughts here?

markelliot:
Separately, I think the way I've seen this solved elsewhere is to declare that runtime extends a sourceSet, and then to have the user declare the CLI version in their sourceSet.

rjullman:
I'm working on other work arounds here -- perhaps something like you describe will work.  

Perhaps a less sketchy way of doing the same thing I'm doing in this PR is to make a task that runs and evaluates the runtime and then using that tasks outputs in the other tasks's configuration (e.g. project.tasks.newtask.output instead of project.configurations.runtime).  Then the doFirst's would not be need.

src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
```diff
@@ -308,6 +307,52 @@ class JavaDistributionPluginTests extends Specification {
 
     }
 
+    def 'produce distribution with dependencies in services/lib'() {
+        given:
+        createUntarBuildFile(buildFile)
+        buildFile << sprintf('''
```
jmcampanini, line 313: why not just use a GString variable replacement instead of `sprintf`?

rjullman, line 313: I'm new to groovy so happy to learn the right way to do this, but from my understanding triple-ticked strings must be literals.  It felt clearer to use multiline strings instead of string replacement and concatenation.  Maybe there is a third way that is cleaner than both of these?

markelliot:
@rjullman do you still want to move forward with this?

rjullman:
Atlas CLIs have become a lower priority and have been handed off to the London team. This may have to be fixed in the future, but for now I'm happy to close this.

PR #74, opened 2016-05-23
default tmp location is var/tmp/files

bavardage:


src/main/groovy/com/palantir/gradle/javadist/DistTarTask.groovy
```diff
@@ -57,6 +57,8 @@ class DistTarTask extends Tar {
             }
         }
 
+        new File(project.projectDir, "var/tmp/files").mkdirs()
```
jared2501, line 60: why `var/tmp/files` and not `var/tmp`?

bavardage, line 60: discussion on internal thread - other product uses var/tmp

briantoth, line 60: sorry i didn't see this earlier.  i strongly recommend against using var/tmp/files.  responded in the thread

PR #75, opened 2016-05-23
var/tmp/files -> var/data/tmp

bavardage:


bavardage:
whoops, thought I could just do this in github ui but missed a bit :P will do locally properly

markelliot:
It's a bit weird that this is in data.

dekramer:
Slightly odd, but if people need more space this is where it needs to go for pcloud.

PR #77, opened 2016-06-08
Update init.sh

tpetracca:
Update SERVICE_HOME.

### DIFF
--- a/src/main/resources/init.sh
+++ b/src/main/resources/init.sh
@@ -22,7 +22,7 @@ PIDFILE="var/run/$SERVICE.pid"
 ARGS="@args@"
 
 # uses SERVICE_HOME when set, else, traverse up two directories respecting symlinks
-SERVICE_HOME=${SERVICE_HOME:-$(cd "$(dirname "$0")/../../" && pwd)}
+SERVICE_HOME=$(cd "$(dirname "$0")/../../" && pwd)
 cd "$SERVICE_HOME"
 
 source service/bin/config.sh


src/main/resources/init.sh
```diff
@@ -22,7 +22,7 @@ PIDFILE="var/run/$SERVICE.pid"
 ARGS="@args@"
 
 # uses SERVICE_HOME when set, else, traverse up two directories respecting symlinks
-SERVICE_HOME=${SERVICE_HOME:-$(cd "$(dirname "$0")/../../" && pwd)}
+SERVICE_HOME=$(cd "$(dirname "$0")/../../" && pwd)
```
markelliot, line 25: This doesn't respect $SERVICE_HOME when it's set before this runs. What's the issue you're running into?

tpetracca, line 25: I linked you a related internal ticket that describes the issue.

markelliot:
Submitting this is blocked on changing the layout spec contract (which requires respecting `SERVICE_HOME` when set). (@briantoth, not sure where we are on moving the layout spec to GH)

briantoth:
@markelliot thanks for tagging me on this.  current status is waiting for oss approval.  @tpetracca can you link me to the issue?

tpetracca:
I can't find which ticket that was that was tracking why I didn't want this.  Basically Orca was setting service_home for me based on the service whose command it was calling (my backup tool), but that command then chain calls a CLI from another service (cassandra's nodetool) which then didn't work because it was using the first service's home and not cassandra's own.

Obviously the chain calling of commands here is not ideal.  I think I also just find environment variables opaque.

markelliot:
We made this change (despite it's noncompliance) in #85.

PR #78, opened 2016-06-19
Fixing config.sh

### LINKED ISSUE #66: Incorrect default variable name in config.sh

vipshek:
See #67, fixes #66

### DIFF
--- a/src/main/resources/config.sh
+++ b/src/main/resources/config.sh
@@ -2,4 +2,4 @@
 @javaHome@
 
 # use JVM_OPTS to set JVM settings, such as minimum and maximum heap
-#JVM_OPTS="-Xms1g -Xmx1g"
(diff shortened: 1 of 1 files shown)


markelliot:
This also needs to address how `@java_home@` is substituted in the code, I think.

markelliot:
As of #85, config.sh no longer exists, so this is OBE.

PR #80, opened 2016-06-28
Minor spelling fixes to readme.md

j-esse:


markelliot:
Thanks, this looks great. Could I get you to sign Palantir's CLA so we can accept the contribution?

Fill out the an appropriate Contributor License Agreement and send it to opensource@palantir.com:
- [Individual Contributor License Agreement](https://github.com/palantir/auth-tokens/blob/develop/palantir_individual_cla.pdf?raw=true)
- [Corporate Contributor License Agreement](https://github.com/palantir/auth-tokens/blob/develop/palantir_corporate_cla.pdf?raw=true)

markelliot:
Ah sorry, just realized you're actually in the Palantir org (just doesn't say so on your profile).

PR #89, opened 2016-07-08
Fetch go-java-launcher from released distribution instead of checked-in binary

uschi2000:


src/main/groovy/com/palantir/gradle/javadist/JavaDistributionPlugin.groovy
```diff
@@ -31,18 +32,23 @@ class JavaDistributionPlugin implements Plugin<Project> {
         // force application of java
         project.plugins.apply('java')
 
+        project.configurations.create("goJavaLauncherBinaries")
+        project.dependencies {
+            goJavaLauncherBinaries "com.palantir.launching:go-java-launcher:1.0.0"
```
markelliot: is it worth allowing this to be overriden?

src/main/groovy/com/palantir/gradle/javadist/JavaDistributionPlugin.groovy
```diff
@@ -31,18 +32,23 @@ class JavaDistributionPlugin implements Plugin<Project> {
         // force application of java
         project.plugins.apply('java')
 
+        project.configurations.create("goJavaLauncherBinaries")
```
markelliot: single quotes, please

markelliot:
code looks okay but your tests are dying

src/main/resources/init.sh
```diff
@@ -27,10 +27,10 @@ pushd "`dirname \"$0\"`/../.." > /dev/null
 # Select launcher binary for this OS
 case "`uname`" in
   Linux*)
-    LAUNCHER_CMD=service/bin/javalauncher-linux-amd64
+    LAUNCHER_CMD=service/bin/linux-amd64/go-java-launcher
     ;;
   Darwin*)
-    LAUNCHER_CMD=service/bin/javalauncher-darwin-amd64
+    LAUNCHER_CMD=service/bin/darwin-amd64/go-java-launcher
```
markelliot, line 33: probably have to adjust the README to match this

src/main/groovy/com/palantir/gradle/javadist/JavaDistributionPlugin.groovy
uschi2000: probably not, but the code would be much easier if didn't explicitly allow it.

uschi2000: (using a gradle configuration also gives you caching for free which is quite handy)

src/main/groovy/com/palantir/gradle/javadist/JavaDistributionPlugin.groovy
uschi2000: done

src/main/resources/init.sh
uschi2000, line 33: done

uschi2000:
no ideas why tests died, trying again.

uschi2000:
no idea why this is failing, will look tomorrow

markelliot:
One thing we might want to consider here is this plugin bundling the go artifacts into the produced jars rather than requiring every downstream project to add the jcenter repo explicitly.

uschi2000:
yep, could do

PR #93, opened 2016-07-15
Add -XX:+PerfDisableSharedMem flag by default

uschi2000:
This change improves performance by avoiding GC pauses when writing stats to
a memory-mapped file under /tmp: http://www.evanjones.ca/jvm-mmap-pause.html.

Note that this behaviour can be overridden by users via '-XX:-PerfDisableSharedMem'.

### DIFF
--- a/src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
+++ b/src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
@@ -20,7 +20,7 @@ class DistributionExtension {
     private static final List<String> requiredJvmOpts = [
         '-Djava.security.egd=file:/dev/./urandom',
         '-Djava.io.tmpdir=var/data/tmp',
-        '-XX:+PerfDisableSharedMem'
+        '-XX:+PerfDisableSharedMem' // avoid memory-mapped IO during GC: http://www.evanjones.ca/jvm-mmap-pause.html
     ]
 
     private String serviceName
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
```diff
@@ -19,7 +19,8 @@ class DistributionExtension {
 
     private static final List<String> requiredJvmOpts = [
         '-Djava.security.egd=file:/dev/./urandom',
-        '-Djava.io.tmpdir=var/data/tmp'
+        '-Djava.io.tmpdir=var/data/tmp',
+        '-XX:+PerfDisableSharedMem'
```
markelliot: can we add a comment linking to the article describing impact of this?

uschi2000: it's in the commit message, but I can add it here too. good idea.

uschi2000: done

schlosna: Probably worth noting the side effect of this is that it will prevent some observability tooling from connecting to the JVM by default (e.g. the java process won't show up in `jps` & `jcmd`, and connecting to it via VisualVM is a bit more difficult). See http://www.evanjones.ca/jvm-mmap-pause.html as well as http://mail.openjdk.java.net/pipermail/hotspot-runtime-dev/2015-March/014324.html and http://bugs.java.com/view_bug.do?bug_id=6447182

markelliot:
Rob, feel free to cut a release.

PR #94, opened 2016-07-19
add an sls configuration

### LINKED ISSUE #92: add an sls configuration + artifact to the project
// + @markelliot + @bmoylan for SA + thoughts

i'd like to refer to projects that produce a java distribution tarball via Gradle dependencies. specifically, i'd like to do something like this:

``` groovy
dependencies {
    skylab project(path: ':product', configuration: 'sls')
}
```

this would give a clean way for a project to depend on the resulting `sls` artifact produced by this plugin.

it would involve adding something like this to the plugin:

``` groovy
configurations {
    sls
}

artifacts {
    sls distTar
}
```

i believe this would be non-disruptive to existing projects.

jmcampanini:
closes #92 

// + @bmoylan for SA

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -98,6 +98,17 @@ As part of package creation, this plugin will create three shell scripts:
 In addition to creating these scripts, this plugin will merge the entire
 contents of `${projectDir}/service` and `${projectDir}/var` into the package.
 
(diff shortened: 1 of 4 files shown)


jmcampanini:
ping

jmcampanini:
w00t! thank you! :)

PR #104, opened 2016-08-02
Remove special /dev/urandom java option

uschi2000:
Java8 uses urandom by default: https://docs.oracle.com/javase/8/docs/technotes/guides/security/enhancements-8.html
Java7 users can still manually set the option.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -79,9 +79,6 @@ concatenating the following list of hard-coded *required options* and the list o
 `distribution.defaultJvmOpts`:
 
 Hard-coded required JVM options:
-- `-Djava.security.egd=file:/dev/./urandom`: urandom is the
-  [preferred source of cryptographic randomness on modern Linux systems](http://www.2uo.de/myths-about-urandom/). Note
-  that some Java implementations override `/dev/urandom` internally; we thus set it to `/dev/./urandom`.
 - `-Djava.io.tmpdir=var/data/tmp`: Allocates temporary files inside the application installation folder rather than on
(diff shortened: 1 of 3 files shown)


uschi2000:
@mglazer @markelliot

mglazer:
👍

markelliot:
Why are we removing this?

uschi2000:
It's redundant in Java8, see https://docs.oracle.com/javase/8/docs/technotes/guides/security/enhancements-8.html .

markelliot:
Thanks. Do we need to mess with any other config settings to make sure we default to the non-blocking SR implementation?

uschi2000:
Mike?

mglazer:
According to this:

http://bugs.java.com/view_bug.do?bug_id=4705093

As long as you do new SecureRandom(), it uses /dev/urandom. If you do:

SecureRandom.getInstance("SHA1PRNG")

It does /dev/random..as expected.

@schlosna for his wisdom on reading JDK bug reports. 

I don't think this is strictly a required change (having it in there should be a no-op), but it cleans up the set of flags which someone has to ask: "Why this flag?"

markelliot:
What does the standard Jetty TLS implementation end up using? (Basically only place where we're worried about this.)

mglazer:
https://github.com/eclipse/jetty.project/search?utf8=%E2%9C%93&q=securerandom

From what I can tell, they do:

```
new SecureRandom()
```

everywhere.

markelliot:
Cool, fine with me, then.

uschi2000:
Rebased.

PR #108, opened 2016-08-18
Change artifact extension to sls.tgz

### LINKED ISSUE #105: File names must end in sls.tgz instead of tgz

bmoylan:
Fixes #105

---

This change is [Reviewable](https://reviewable.io/reviews/palantir/gradle-java-distribution/108)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -35,7 +35,7 @@ content of the package. The package will follow this structure:
 The `service/bin/` directory contains both Gradle-generated launcher scripts (`[service-name]` and `[service-name].bat`)
 and [go-java-launcher](https://github.com/palantir/go-java-launcher) launcher binaries.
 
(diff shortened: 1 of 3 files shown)


bmoylan:
need to update README and maybe add a test for this getting into the `pom.xml` created by `maven-publish`

markelliot:
@bmoylan could you update the readme as part of this PR, please?

bmoylan:
Yes, will do it right now. When are we planning on open-sourcing (making public) the spec repo so that we can link to it from the readme here? The `sls` seems kind of randomly thrown in without the context.

markelliot:
I'm not sure what we're waiting on w/r/t flipping the switch on the other repo, but this repo does describe its own intent " In particular, this plugin packages a project into a common deployment structure with a simple start script", so not too random.

bmoylan:
Updated filename in readme

PR #110, opened 2016-08-24
Update init.sh to exit 0 for "already running"

### LINKED ISSUE #109: init.sh start exits 1 for "already running", should exit 0
Per the spec here: https://refspecs.linuxbase.org/LSB_3.0.0/LSB-PDA/LSB-PDA/iniscrptact.html

> ...In addition to straightforward success, the following situations are also to be considered successful:
> - running start on a service already running

PR incoming

bmoylan:
Fixes #109

---

This change is [Reviewable](https://reviewable.io/reviews/palantir/gradle-java-distribution/110)

### DIFF
--- a/src/main/resources/init.sh
+++ b/src/main/resources/init.sh
@@ -48,7 +48,7 @@ start)
     service/bin/init.sh status > /dev/null 2>&1
     if [[ $? == 0 ]]; then
         echo "Process is already running"
-        exit 1
+        exit 0
     fi
     printf "%-50s" "Running '$SERVICE'..."
 
--- a/src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
+++ b/src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
@@ -61,6 +61,7 @@ class JavaDistributionPluginTests extends GradleTestSpec {
         // try all of the service commands
         exec('dist/service-name-0.1/service/bin/init.sh', 'start') ==~ /(?m)Running 'service-name'\.\.\.\s+Started \(\d+\)\n/
         file('dist/service-name-0.1/var/log/service-name-startup.log', projectDir).text.contains('Test started\n')
+        exec('dist/service-name-0.1/service/bin/init.sh', 'start') ==~ /(?m)Process is already running\n/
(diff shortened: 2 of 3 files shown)


bmoylan:
@markelliot

src/main/resources/init.sh
```diff
@@ -48,7 +48,7 @@ start)
     service/bin/init.sh status > /dev/null 2>&1
     if [[ $? == 0 ]]; then
         echo "Process is already running"
-        exit 1
+        exit 0
```
uschi2000, line 51: test?

src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
```diff
@@ -61,6 +61,7 @@ class JavaDistributionPluginTests extends GradleTestSpec {
         // try all of the service commands
         exec('dist/service-name-0.1/service/bin/init.sh', 'start') ==~ /(?m)Running 'service-name'\.\.\.\s+Started \(\d+\)\n/
         file('dist/service-name-0.1/var/log/service-name-startup.log', projectDir).text.contains('Test started\n')
+        exec('dist/service-name-0.1/service/bin/init.sh', 'start') ==~ /(?m)Process is already running\n/
```
uschi2000, line 64: this isn't verifying the return value, right?

uschi2000:
For reference: http://refspecs.linuxbase.org/LSB_3.1.1/LSB-Core-generic/LSB-Core-generic/iniscrptact.html requires `status` to return 0 iff "program is running or service is OK".

src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
bmoylan: Yeah I just ran the test on develop and it passed (so doesn't catch the bug), working on making `exec()` check the return code. I think there's an assumption throughout this suite that it returns 0 so I can bake in an assertion - you cool with that?

uschi2000, line 64: bake it into exec()? yeah that sounds ok.

bmoylan: done

uschi2000:
👍

PR #116, opened 2016-09-11
Clarify use of setters in DistributionExtension; make var/log and var/run excludes configurable

uschi2000:


uschi2000:
This might be controversial since it's a config-breaking change for anyone taking an upgrade. Rationale: simplify the semantics of the distribution object and make its behaviour more Groovy/Gradle standard. Example:

Old behaviour:

```
distribution {
  args 'foo'
  args 'bar'
}
assert distribution.args == ['bar'] // oops, probably unexpected
```

New behaviour:

```
distribution {
  args = ['foo']
  args = ['bar']
}
assert distribution.args == ['bar'] // expected as per above syntax

distribution {
  args = ['foo']
  args << 'bar' << 'boom'
}
assert distribution.args == ['foo', 'bar', 'boom'] // cool, can use standard groovy things

```

schlosna:
Rob, can you provide the context on this change?

I think you want to add a test case for the last example above as I believe that should include foo and not bar twice:

```
distribution {
  args = ['foo']
  args << 'bar' << 'boom'
}
assert distribution.args == ['foo', 'bar', 'boom'] // cool, can use standard groovy things
```

uschi2000:
thanks, I have a typo in my example. edited/fixed above.

uschi2000:
Context: I consider the current configuration mechanism a bit whacky and would like to remove tech debt.

markelliot:
I'm not in favor of this change -- it's more idiomatic Gradle to use the setters, IMO.

uschi2000:
that's not quite accurate, if I remember right. typically, gradle objects have setters (whose names start with `set`) and "accumulators" (for the lack of a better name). this would look like this:

```
distribution {
  args "foo"
  args "bar"
  // args is now [foo, bar]
  setArgs ["boom", "baz"]
  // args is now [boom, baz]
}
```

happy to support both here, but the current model is a little off IMO

markelliot:
Haven't seen the set method syntax -- this is a poor man's replication of the slightly fancier Action syntax the Gradle core stuff uses: https://github.com/gradle/gradle/blob/master/subprojects/publish/src/main/java/org/gradle/api/publish/PublishingExtension.java

uschi2000:
here's an example from DefaultSourceDirectorySet.

```
    public SourceDirectorySet srcDirs(Object... srcDirs) {
        for (Object srcDir : srcDirs) {
            source.add(srcDir);
        }
        return this;
    }

    public SourceDirectorySet setSrcDirs(Iterable<?> srcPaths) {
        source.clear();
        GUtil.addToCollection(source, srcPaths);
        return this;
    }
```

uschi2000:
will change this pr to adopt that convention.

markelliot:
Are there any examples of single-valued things? Not all of our parameters are iterable.

uschi2000:
yep, standard setter

```
    public void setSourceCompatibility(Object value) {
        srcCompat = JavaVersion.toVersion(value);
    }
```

markelliot:
sourceCompatibility is done using the old container method instead of extensions; I think we want to model what we're doing after extensions.

uschi2000:
rebased. to summarize:

Simplify the semantics of the distribution object and make its behaviour more Groovy/Gradle standard. Example:

Old behaviour:

```
distribution {
  args 'foo'
  args 'bar'
}
assert distribution.args == ['bar'] // oops, probably unexpected
```

New behaviour:

```
distribution {
  args 'foo'
  args 'bar'
}
assert distribution.args == ['foo', 'bar']

distribution {
  args 'foo'
  setArgs(['bar'])
}
assert distribution.args == ['bar']
```

markelliot:
lgtm. I think the `=` signs in your example of new behavior aren't supposed to be there, though.

readme.md
```diff
@@ -65,9 +65,9 @@ The `distribution` block offers the following options:
...
    classpath is omitted from the generated Windows start script and instead inferred
    from a JAR file whose MANIFEST contains the classpath entries.
- * (optional) `excludeFromVar` a list of directories (relative to `${projectDir}/var`) to exclude 
-   in addition to `log` and `run` (which are always excluded).
-   **Note**: this plugin will _always_ create `var/data/tmp` in the resulting distribution to
+ * (optional) `excludeFromVar` a list of directories (relative to `var`) to exclude from the distribution,
+   defaulting to `['log', 'run']`.
+   **Note**: this plugin will *always* create `var/data/tmp` in the resulting distribution to
    ensure the prescribed Java temp directory exists. Setting `data` for this option will still ensure
    nothing in `${projectDir}/var/data` is copied.
```
markelliot, line 72: you removed `${projectDir}` above, but not here?

Issue #114, opened 2016-09-07
Should possibly create the tmpdir by means other than creating it in source


j-baker:
Saw an internal product which forcibly deletes var/data before doing any docker tasks; this removed the tmpdir that gjd created. Gjd should consider creating the dir by some other mechanism, because the current method is liable to being tampered with and then leading to confusing runtime bugs.

markelliot:
Yeah, we can do something different here.

schlosna:
For Cassandra and ES you also need to set `jna.tmpdir` (similar to `java.io.tmpdir`), but might want them to be different, in which case you'd need to also create that a directory that is readable, writable, and executable (and isn't mounted with `noexec` as some `/tmp/` or `/var/tmp/` may be).

@tpetracca for SA 

https://docs.datastax.com/en/landing_page/doc/landing_page/troubleshooting/DSE/noWriteTmpDir.html and https://github.com/elastic/elasticsearch/issues/18272 for background

quellkot:
It's not clear to me that gjd should create var/data and var/data/tmp at all?
If we do create them, then do we agree that they should always be empty?

(this was uschi2000/robert with a different account)

schlosna:
Not sure what the state of SLS spec is, but I'd propose that GJD should create the minimum expected directory structure with appropriate permissioning, to include var/data/ and var/tmp/ (to be consistent with Linux filesystem hierarchy standards http://www.tldp.org/LDP/Linux-Filesystem-Hierarchy/html/var.html )

uschi2000:
Hmm. I think it's generally safer to be explicit in your application: if you're expecting to write to var/data then you should make sure the directory exist and optionally create it if not. `var` is guaranteed to exist as per SLS.

j-baker:
@uschi2000 you have to choose between either not setting java.io.tmpdir or making sure that the directory you set it to exists. Otherwise when any part of java tries to create a temporary file, it'll crash.

quellkot:
Ah, yes I forgot about the tmpdir setting. Cool you're right.

uschi2000:
fixed by #119

PR #117, opened 2016-09-11
Distribution contains /var/data/tmp even if tmp/ dir is cleared by user

uschi2000:


uschi2000:
DO NOT MERGE. Needs to get rebased onto develop once #116 merges.

uschi2000:
Fixes #114

src/main/groovy/com/palantir/gradle/javadist/tasks/DistTarTask.groovy
```diff
@@ -40,7 +40,18 @@ class DistTarTask extends Tar {
...
             }
 
-            new File(project.projectDir, "var/data/tmp").mkdirs()
+            // Create var/data/tmp
+            File.createTempDir().with { emptyTempDirBase ->
+                emptyTempDirBase.deleteOnExit()
+                emptyTempDirBase.toPath().resolve("data").resolve("tmp").toFile().mkdirs()
+                from(emptyTempDirBase) {
+                    into "${archiveRootDir}/var"
+                    // TODO(rfink) If users exclude 'data', then the temp dir will not be created. That seems wrong?
```
tpetracca, line 49: Per https://github.com/palantir/gradle-java-distribution/issues/114#issuecomment-246156725 I think we want to just have it be `var/tmp`, and then you don't have to worry about this.

markelliot: Unfortunately var/tmp is already used. There was a discussion about this previously.

markelliot:
I had taken a stab at this on Friday, both of these seem valid, but I had a smaller diff: #119

PR #119, opened 2016-09-12
Ensure tmp directory gets created

markelliot:


src/main/groovy/com/palantir/gradle/javadist/tasks/DistTarTask.groovy
```diff
@@ -43,7 +43,10 @@ class DistTarTask extends Tar {
                 }
             }
 
-            new File(project.projectDir, "var/data/tmp").mkdirs()
+            new File(project.buildDir, "distributions/var/data/tmp").mkdirs()
+            from ("${project.buildDir}/distributions/var/data") {
```
uschi2000: should add to readme that `excludeFromVar 'data'` has no effect.

markelliot: Can clarify in the readme, but the exclusion is copying `${projectDir}/var/data`, which this still respects.

uschi2000:
the coveralls integration seems borked?

uschi2000:
(not only on this pr)

markelliot:
yeah, we reset our 3rd party integrations, going to see if i can fix

markelliot:
Yeah, status API requests don't seem to be going through. Will hit up our admin team to see what's up.

PR #122, opened 2016-09-30
[BREAKING] manifest.yml keys are kebab-case instead of camelCase, fixed #116

uschi2000:
This change is [Reviewable](https://reviewable.io/reviews/palantir/gradle-java-distribution/122)

### DIFF
--- a/src/main/resources/manifest.yml
+++ b/src/main/resources/manifest.yml
@@ -1,3 +1,3 @@
-manifest-version: 1.0.0
+manifest-version: 1.0
 product-name: @serviceName@
 product-version: @serviceVersion@
--- a/src/main/resources/manifest.yml
+++ b/src/main/resources/manifest.yml
@@ -1,3 +1,3 @@
-manifestVersion: 1.0.0-alpha
-productName: @serviceName@
-productVersion: @serviceVersion@
+manifest-version: 1.0.0
+product-name: @serviceName@
+product-version: @serviceVersion@
--- a/src/main/resources/manifest.yml
+++ b/src/main/resources/manifest.yml
@@ -1,3 +1,4 @@
 manifest-version: 1.0
+product-group: @serviceGroup@
 product-name: @serviceName@
 product-version: @serviceVersion@
--- a/src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
+++ b/src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
@@ -13,7 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.palantir.gradle.javadist;
+package com.palantir.gradle.javadist
+
+import org.gradle.api.Project;
 
 class DistributionExtension {
 
@@ -29,6 +31,9 @@ class DistributionExtension {
             '-verbose:gc'
     ]
 
+    private final Project project
+
+    private String serviceGroup
     private String serviceName
     private String mainClass
     private List<String> args = []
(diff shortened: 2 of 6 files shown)


uschi2000:
@nmiyake , how far are you with parsing kebab-case manifest.yml files?

nmiyake:
Sorry, I know it's something that we've discussed, but I'm not actively tracking it.

I assume the question boils down to Skylab being able to parse kebab-case and mapping it to non-kebab-case fields (or vice versa)? If we crystalize what the request is and where it needs to be supported (`manifest.yml` only, or all `YML` files?), we can prioritize it and shouldn't be too bad to implement.

markelliot (commented):
Can we start getting group with this change as well?

Nick, this change is focused strictly on the manifest. If it helps, we could bump the version number and you can just have a version specific parser?

uschi2000:
Bumped version to 1.0.0 from 1.0.0-alpha

uschi2000:
spoke to @nmiyake , they can make the required change in the next release. think we're good to go with this change @markelliot

src/main/resources/manifest.yml
```diff
@@ -1,3 +1,3 @@
-manifestVersion: 1.0.0-alpha
-productName: @serviceName@
-productVersion: @serviceVersion@
+manifest-version: 1.0.0
+product-name: @serviceName@
```
markelliot, line 3: Also start including product group

markelliot (commented):
Fine with this, but think we should wait until upstream changes for parsing new format release before merging.

src/main/resources/manifest.yml
uschi2000: Will do in separate commit, I think

uschi2000: Re release/merge: happy to postpone the merge, or just not make a release before other products support this

uschi2000:
Updated `manifest-version: 1.0` to match https://github.com/palantir/sls-spec/pull/10
@bmoylan

src/main/resources/manifest.yml
```diff
@@ -1,3 +1,3 @@
-manifestVersion: 1.0.0-alpha
-productName: @serviceName@
-productVersion: @serviceVersion@
+manifest-version: 1.0
+product-name: @serviceName@
+product-version: @serviceVersion@
```
bmoylan, line 4: `product-group` and `product-type` are [required](https://github.com/palantir/sls-spec/blob/master/manifest.md) in Manifest 1.0, we should fill them in here too. The group we can get from the gradle project; for the type we might need another field on the DistributionExtension

EDIT: type is not required (we default to `service.v1`) but people should be able to override this with daemon or asset.

uschi2000:
Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

_[src/main/resources/manifest.yml, line 3 at r3](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/122#-KUstc4s79Q5dZFPrkVZ-r3-3:-KUsyh5rvdmbbr404kll:bz53s2k) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/2a2ccb3abb4a00f69389bc30d01d275a0684f729/src/main/resources/manifest.yml#L3)):_
<details><summary>

<i>Previously, bmoylan (Brad Moylan) wrote…</i></summary>



> `product-group` and `product-type` are [required](https://github.com/palantir/sls-spec/blob/master/manifest.md) in Manifest 1.0, we should fill them in here too. The group we can get from the gradle project; for the type we might need another field on the DistributionExtension
> 
> EDIT: type is not required (we default to `service.v1`) but people should be able to override this with daemon or asset.
> </details>
> added product-group field, and will add service-type in follow-up PR.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/122)_

bmoylan:
Reviewed 6 of 6 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/122)_

bmoylan:
:lgtm:

---

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/122#-:-KUt0uXkdW44vsYZINuM:bnfp4nl)_

PR #128, opened 2016-10-18
enableManifestClasspath should also affect linux and init.sh start scripts

jfosco:
You can now define a customClasspath for the distribution which will
override the expansion of all *.jar files in the runtime configuration.

---

This change is [Reviewable](https://reviewable.io/reviews/palantir/gradle-java-distribution/128)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -63,7 +63,7 @@ The `distribution` block offers the following options:
    no monitoring script will be generated.
  * (optional) `defaultJvmOpts` a list of default JVM options to set on the program.
  * (optional) `enableManifestClasspath` a boolean flag; if set to true, then the explicit Java
-   classpath is omitted from the generated Windows start script and instead inferred
+   classpath is omitted from the generated start scripts and instead inferred
    from a JAR file whose MANIFEST contains the classpath entries.
  * (optional) `excludeFromVar` a list of directories (relative to `${projectDir}/var`) to exclude from the distribution,
    defaulting to `['log', 'run']`.
--- a/src/main/groovy/com/palantir/gradle/javadist/tasks/LaunchConfigTask.groovy
+++ b/src/main/groovy/com/palantir/gradle/javadist/tasks/LaunchConfigTask.groovy
@@ -16,17 +16,19 @@
 
 package com.palantir.gradle.javadist.tasks
 
-import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
-import com.palantir.gradle.javadist.JavaDistributionPlugin
-import groovy.transform.EqualsAndHashCode
+import java.nio.file.Files
+
 import org.gradle.api.file.FileCollection
 import org.gradle.api.plugins.JavaPlugin
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
 
-import java.nio.file.Files
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
+import com.palantir.gradle.javadist.JavaDistributionPlugin
+
+import groovy.transform.EqualsAndHashCode
 
 class LaunchConfigTask extends BaseTask {
 
@@ -87,8 +89,13 @@ class LaunchConfigTask extends BaseTask {
         config.mainClass = distributionExtension().mainClass
         config.javaHome = distributionExtension().javaHome ?: ""
         config.args = args
-        config.classpath = relativizeToServiceLibDirectory(
-                project.tasks[JavaPlugin.JAR_TASK_NAME].outputs.files + project.configurations.runtime)
+        if (distributionExtension().isEnableManifestClasspath()) {
+            config.classpath = relativizeToServiceLibDirectory(
+                    project.tasks.getByName('manifestClasspathJar').outputs.files)
+        } else {
+            config.classpath = relativizeToServiceLibDirectory(
+                    project.tasks[JavaPlugin.JAR_TASK_NAME].outputs.files + project.configurations.runtime)
+        }
         config.jvmOpts = distributionExtension().defaultJvmOpts
         return config
     }
--- a/src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
+++ b/src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
@@ -15,13 +15,14 @@
  */
 package com.palantir.gradle.javadist
 
-import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
-import com.palantir.gradle.javadist.tasks.LaunchConfigTask
+import java.nio.file.Files
+
 import org.gradle.testkit.runner.BuildResult
 import org.gradle.testkit.runner.TaskOutcome
 
-import java.nio.file.Files
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
+import com.palantir.gradle.javadist.tasks.LaunchConfigTask
 
 class JavaDistributionPluginTests extends GradleTestSpec {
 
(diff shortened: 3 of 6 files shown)


src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
```diff
@@ -309,6 +310,47 @@ class JavaDistributionPluginTests extends GradleTestSpec {
...
+
+        then:
+        def actualStaticConfig = new ObjectMapper(new YAMLFactory()).readValue(
+            file('dist/service-name-0.1/service/bin/launcher-static.yml'), LaunchConfigTask.StaticLaunchConfig)
+        def expectedStaticConfig = new LaunchConfigTask.StaticLaunchConfig()
+        expectedStaticConfig.setConfigVersion(1)
+        expectedStaticConfig.setConfigType("java")
+        expectedStaticConfig.setJavaHome("")
+        expectedStaticConfig.setMainClass("test.Test")
+        expectedStaticConfig.setClasspath(['service/lib/manager.jar', 'service/lib/config.jar'])
```
uschi2000: can probably worry about this field only in the assertion

src/main/groovy/com/palantir/gradle/javadist/tasks/LaunchConfigTask.groovy
```diff
@@ -65,8 +67,12 @@ class LaunchConfigTask extends BaseTask {
         config.mainClass = distributionExtension().mainClass
         config.javaHome = distributionExtension().javaHome ?: ""
         config.args = args
-        config.classpath = relativizeToServiceLibDirectory(
-                project.tasks[JavaPlugin.JAR_TASK_NAME].outputs.files + project.configurations.runtime)
+        if (distributionExtension().getCustomClasspath().size() == 0) {
```
uschi2000: isEmpty

uschi2000 (changes requested):
(1) need to update readme
(2) what's the use case for this? the implementation seems fine module some nitpicks, but I don't quite understand the rationale.

uschi2000:
further, we should fix the classpath at compile time; so it shouldn't contain `*` and other run-time -evaluated things like `../` etc

src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
```diff
@@ -62,6 +65,16 @@ class DistributionExtension {
         this.checkArgs = checkArgs.toList()
     }
 
+    public void customClasspath(String... customClasspath) {
+        assert customClasspath.every { isAllowedClasspathChar(it) }
```
uschi2000: should give more useful error of the form "Found foo, expected bar"

readme.md
```diff
@@ -62,6 +62,8 @@ The `distribution` block offers the following options:
  * (optional) `checkArgs` a list of arguments to supply to the monitoring script, if omitted,
    no monitoring script will be generated.
  * (optional) `defaultJvmOpts` a list of default JVM options to set on the program.
+ * (optional) `customClasspath` a list of jars to put on the classpath instead of all jars in the
```
uschi2000: this is not quite complete enough. at the minimum, must say that these are file names relative to some path, that they must not contain certain characters, something about the classpath order

src/test/groovy/com/palantir/gradle/javadist/DistributionExtensionTest.groovy
```diff
@@ -59,12 +63,41 @@ class DistributionExtensionTest extends Specification {
...
+        ext.customClasspath == ['g', 'h']
+    }
+
+    def 'can not define classpath with path expansion'() {
+        given:
+        def ext = new DistributionExtension()
+
+        when:
+        ext.with {
+            setCustomClasspath(['*', '..', ':', '/'])
```
uschi2000: every single one of them should throw, right?

src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
```diff
@@ -309,6 +310,27 @@ class JavaDistributionPluginTests extends GradleTestSpec {
...
+            dependencies { compile files("external.jar") }
+            tasks.jar.baseName = "internal"
+            distribution {
+                customClasspath 'manager.jar', 'config.jar'
+            }
+        '''.stripIndent()
+        file('src/main/java/test/Test.java') << "package test;\npublic class Test {}"
+
+        when:
+        runSuccessfully(':build', ':distTar', ':untar')
```
uschi2000, line 332: i actually believe this should fail if the classpath files are not readable

uschi2000 (changes requested):

src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
jfosco: If I use the custom classpath to filter against the list of jars originally used in the LaunchConfigTask, do I still need to check for special characters?

uschi2000: IMO, it's fine to just check the custom ones

src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy
jfosco: Good point. I changed it to require that the customClasspath jars have to match actual dependency jars.

uschi2000:
I still think this is a little convoluted, sorry. (I should have been more specific here from the start.) The default behaviour is to to use the concatenation of the jar and the runtime dependencies. I would propose to make provide an optional `FileCollection classpath` configuration point such that: if it is set, then this FileCollection becomes the classpath, else the default is the classpath (see above).

---

Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

_[a discussion](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/128#-KUecffZPyIUEj0gNIo6:-KUecffZPyIUEj0gNIo7:b-c6aubj) (no related file):_
Further, the ManifestClasspathJarTask must be updated accordingly, so this is likely a slightly bigger refactor.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/128#-:-KUebVYXP3vjxwoW4rEj:boazdqi)_

uschi2000:
So this PR is now about enabling the classpath jar, rather than about allowing to customize the classpath, right?

---

Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/128#-:-KV2iyVxpcss5IU9_1kN:b9o7vqr)_

uschi2000:
Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

_[src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy, line 338 at r3](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/128#-KV2lBXmUkACP4ussLyq:-KV2lBXnHeEe2uFdyC0J:b-lu6uj7) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/561434821bd8cd74a7bf3f95ad9f9697eee89433/src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy#L338)):_

> ``` Groovy
>     }
> 
>     def 'produces manifest-classpath jar and windows start script with no classpath length limitations'() {
> ```

stale test name

---

_[src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy, line 362 at r3](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/128#-KV2lEcfuICOMRZtUXwT:-KV2lEcfuICOMRZtUXwU:b3ofc27) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/561434821bd8cd74a7bf3f95ad9f9697eee89433/src/test/groovy/com/palantir/gradle/javadist/JavaDistributionPluginTests.groovy#L362)):_

> ``` Groovy
>         String launcher = file('dist/service-name-0.1/service/bin/launcher-static.yml', projectDir).text
>         !launcher.contains("external.jar")
>     }
> ```

would like to see an end-to-end test for launching the service through the unix classpath jar

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/128)_

uschi2000:
Reviewed 5 of 7 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/128)_

jfosco:
Done, but producing a manifest classpath jar is busted until you merge in YY's commit fixing how the plugin reaches into the gradle object model

uschi2000:
Closing since there wasn't much  movement. Feel free to reopen if this is still an issue.

PR #130, opened 2016-10-19
Do not force dependency resolution during configuration phase

lee467:
For example, accessing project.configurations.runtime during
configuration phase (.../blob/0.9.0/.../ManifestClasspathJarTask.groovy#L36)
slows down all the other tasks in the project that uses this plugin.

Also, favour composition over inheritance by reusing Gradle's
Copy/Jar/Tar tasks instead of extending them when the new tasks
only reconfigure the tasks.

Lastly, tasks should not reach into the Gradle object model
(BaseTask.distributionExtenion()). Instead have the plugin supply
a method for tasks to use. This will probably need further
clean-up/refactoring.

### DIFF
--- a/src/main/groovy/com/palantir/gradle/javadist/JavaDistributionPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/javadist/JavaDistributionPlugin.groovy
@@ -34,16 +34,20 @@ class JavaDistributionPlugin implements Plugin<Project> {
             goJavaLauncherBinaries 'com.palantir.launching:go-java-launcher:1.0.1'
         }
 
+        project.ext.set ("distributionExtension", {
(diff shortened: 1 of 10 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-java-distribution](https://github.com/palantir/gradle-java-distribution), @lee467! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

uschi2000:
Can you please split refactoring changes and behaviour changes into different PRs?

lee467:
Sure thing. I'll separate the changes that stops dependency resolution during configuration phase and the rest of the changes.

PR #134, opened 2016-10-21
Tasks should not reach into the Gradle object model.

lee467:
```
Tasks should not reach into the Gradle object model.

Remove unnecessary inhearitance of BaseTask and duplicated distributionExtension
method. Instead, have each custom task declare specific input properties and the
JavaDistributionPlugin set them after task creation, so that custom tasks are completely
independent with clear correlation to only the relevant DistributionExtension data.

This also necessitates refactoring to reuse gradle tasks like Copy/Tar/JavaExec
instead of extending them, especially since the child tasks are only reconfiguring
them.

Change 'distTar artifact name is set during appropriate lifecycle events' test
since now the task classes don't have direct access to project's DistributionExtension.
If we allow plugin users to directly configure each task, we can assert the tar
name is the same before and after the project evaluation.
```

---

This change is [Reviewable](https://reviewable.io/reviews/palantir/gradle-java-distribution/134)


uschi2000:
Can you provide a rationale why this method would be preferred? I struggle a bit with the magical nature of the `project.ext` mechanism; for instance, it's where hard to find how this works in the Gradle source.

lee467:
1. This removes duplicated code and unnecessary inheritance just to give task classes access to the distributionExtension method.
2. In the previous approach, tasks classes are assumed to know the details of the project configurations (project.extensions.findByType(DistributionExtension)), which makes them less portable. While this change still doesn't make them completely portable, the plugin code provides a method for them to call by setting a closure in the project.ext, details of which is hidden from the task classes. Again, ideally task classes should be completely independent by having salient properties declared as inputs and the plugin code set them - this will also allow for better incremental builds. 
3. https://docs.gradle.org/current/dsl/org.gradle.api.plugins.ExtraPropertiesExtension.html
   https://github.com/gradle/gradle/blob/f001402c8b7f404c923d972f1bc524b2de7b3fcd/subprojects/core/src/main/java/org/gradle/api/plugins/ExtraPropertiesExtension.java

uschi2000:
Re 1: Agree that removing duplicated code is desirable.
Re 2: This PR is replacing one method of cross-task-project coupling (findByType) by another one (project.ext). The project.ext method is pure Groovy runtime magic (e.g., you don't get any help in your IDE); this is why I prefer the compile-time coupling via subclassing and findByType. I don't think it's quite possible to completely reduce the coupling since the project as the creator of tasks and extensions as the configurer of tasks have different life-cycles. So you need some variant of shared state.

---

Review status: 0 of 10 files reviewed at latest revision, all discussions resolved.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/134#-:-KUsiIDsnE5iups3XWih:b-kdixg4)_

uschi2000:
Reviewed 10 of 10 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/134)_

lee467:
I want to have JavaDistributionPlugin class set the inputs for tasks classes instead of having task classes reaching into project. That is, I will refactor the code further to pass project.distributionExtension().serviceName and project.distributionExtension().checkArgs to CreateCheckScriptTask class as arguments, instead of having the CreateCheckScriptTask class call project.distributionExtension() method from within its task action method. This way there is no tight coupling. This PR makes the initial changes to remove the duplicated code and unnecessary inheritance revolving around distributionExtension. I can combine these changes, but I feel these can be more or less separate which makes PRs simpler.

uschi2000:
@lee467 should we just talk in person? that might be faster.

jmcampanini:
another way to decouple them would be to have the task have properties that are set at creation or during an `afterEvaluate`, so it doesn't know anything about the extension, it just knows it needs values to be set.

lee467:
Bump. Also @mglazer, in case you have any input since I fiddled with your changes a bit.

src/main/groovy/com/palantir/gradle/javadist/JavaDistributionPlugin.groovy
```diff
@@ -34,16 +34,50 @@ class JavaDistributionPlugin implements Plugin<Project> {
             goJavaLauncherBinaries 'com.palantir.launching:go-java-launcher:1.1.0'
         }
 
+        project.ext.set ("distributionExtension", {
```
markelliot: This doesn't seem to be used anywhere, and so can be removed.

lee467: Sorry, forgot to take this out.

uschi2000:
generally happy with this, but i think we'd want to go all-in if we want to disentangle tasks and distribution extension.

---

Review status: 1 of 11 files reviewed at latest revision, 4 unresolved discussions.

---

_[src/main/groovy/com/palantir/gradle/javadist/JavaDistributionPlugin.groovy, line 49 at r3](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/134#-KVLZU352frJ6_ToRwhL:-KVLZU352frJ6_ToRwhM:bjc1cx7) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/a86e519cc0444681cc66eec7476d3add0362bf9b/src/main/groovy/com/palantir/gradle/javadist/JavaDistributionPlugin.groovy#L49)):_

> ``` Groovy
> 
>         LaunchConfigTask launchConfig = project.tasks.create('createLaunchConfig', LaunchConfigTask)
>         project.afterEvaluate {
> ```

conceptually, i feel like a task should know how to configure itself. thoughts on using a constructor-like `configure(mainClass, args, checkArgs, ...)` method on the task so that we can do `project.afterEvaluate { launchConfig.configure(...) }` ? this would also allow to make a task-private runtime check like `isConfigured` which fails unless configure() was called.

---

_[src/main/groovy/com/palantir/gradle/javadist/tasks/DistTarTask.groovy, line 26 at r3](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/134#-KVLwTc5fB2rO-_H9s9E:-KVLwTc5fB2rO-_H9s9F:b-z9j8mn) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/a86e519cc0444681cc66eec7476d3add0362bf9b/src/main/groovy/com/palantir/gradle/javadist/tasks/DistTarTask.groovy#L26)):_

> ``` Groovy
> class DistTarTask {
> 
>     public static Tar createDistTarTask(Project project, String taskName, DistributionExtension distributionExtension) {
> ```

shouldn't use distribution extension for symmetry, right?

---

_[src/main/groovy/com/palantir/gradle/javadist/tasks/RunTask.groovy, line 25 at r3](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/134#-KVLwwrZW-8Q2GZ8K_ZC:-KVLwwrZW-8Q2GZ8K_ZD:bpx5u5i) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/a86e519cc0444681cc66eec7476d3add0362bf9b/src/main/groovy/com/palantir/gradle/javadist/tasks/RunTask.groovy#L25)):_

> ``` Groovy
> class RunTask {
> 
>     public static JavaExec createRunTask(Project project, String taskName, DistributionExtension distributionExtension) {
> ```

shouldn't use distribution extension?

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/134#-:-KVLx09G2S1AZulmaAJL:bnh1f5t)_

lee467:
Review status: 1 of 11 files reviewed at latest revision, 4 unresolved discussions.

---

_[src/main/groovy/com/palantir/gradle/javadist/tasks/DistTarTask.groovy, line 26 at r3](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/134#-KVLwTc5fB2rO-_H9s9E:-KVVsNkwKw_slcLIzjMu:b-yg2l9t) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/a86e519cc0444681cc66eec7476d3add0362bf9b/src/main/groovy/com/palantir/gradle/javadist/tasks/DistTarTask.groovy#L26)):_
<details><summary>

<i>Previously, uschi2000 (Robert Fink) wrote…</i></summary>



> shouldn't use distribution extension for symmetry, right?
> </details>
> DistTarTask is not a task class, i.e. doesn't extend DefaultTask. It is just a script since we are reusing gradle's Tar task. I kept the separate class to make the JavaDistributionPlugin script simpler and more organized.

---

_[src/main/groovy/com/palantir/gradle/javadist/tasks/RunTask.groovy, line 25 at r3](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/134#-KVLwwrZW-8Q2GZ8K_ZC:-KVVsqner6BiB2ADSo_2:b-tln4ln) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/a86e519cc0444681cc66eec7476d3add0362bf9b/src/main/groovy/com/palantir/gradle/javadist/tasks/RunTask.groovy#L25)):_
<details><summary>

<i>Previously, uschi2000 (Robert Fink) wrote…</i></summary>



> shouldn't use distribution extension?
> </details>
> Same reason as for DistTar class.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/134)_

uschi2000:
The impetus for this PR is to decouple task internals from the user-facing
DistributionExtension config, right? I believe that's a worthwhile goal,
but don't think it's accomplished as long as the extension is still used in
those tasks, be it via subclassed or composed tasks.

lee467:
DistTar and RunTask classes are just scripts, not task classes. None of the tasks defined/being configured by this plugin has access to the DistributionExtension. But I agree that it's quite misleading, I'll change the code to have the plugin code pass individual properties instead of passing the DistributionExtension.

uschi2000:
Cool, thanks

uschi2000:
LGTM. Needs a rebase though.

---

Reviewed 1 of 10 files at r2, 9 of 9 files at r4.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/134#-:-KWEW2Yil9mRWiJcCkuI:brvtd6s)_

uschi2000 (approved):

uschi2000:
@lee467 can you rebase please?

lee467:
I did a week ago right after your comment. I should've commented - rebased and pushed again.

uschi2000:
Reviewed 3 of 3 files at r5.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/134)_

PR #137, opened 2016-10-26
Support env vars in static-launcher.yml

mglazer:
This change is [Reviewable](https://reviewable.io/reviews/palantir/gradle-java-distribution/137)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -52,6 +52,7 @@ program for a default run configuration:
         serviceName 'my-service'
         mainClass 'com.palantir.foo.bar.MyServiceMainClass'
         args 'server', 'var/conf/my-service.yml'
+        env 'KEY1': 'value1', 'KEY2': 'value1'
     }
 
 The `distribution` block offers the following options:
@@ -61,6 +62,9 @@ The `distribution` block offers the following options:
  * (optional) `args` a list of arguments to supply when running `start`.
  * (optional) `checkArgs` a list of arguments to supply to the monitoring script, if omitted,
    no monitoring script will be generated.
+ * (optional) `env` a map of environment variables that will be placed into the `env` block
+   of the static launcher config. See [go-java-launcher](https://github.com/palantir/go-java-launcher)
+   for details on the custom environment block.
  * (optional) `defaultJvmOpts` a list of default JVM options to set on the program.
  * (optional) `enableManifestClasspath` a boolean flag; if set to true, then the explicit Java
    classpath is omitted from the generated Windows start script and instead inferred
--- a/src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
+++ b/src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
@@ -39,6 +39,7 @@ class DistributionExtension {
     private List<String> args = []
     private List<String> checkArgs = []
     private List<String> defaultJvmOpts = []
+    private Map<String,String> env = [:]
     private boolean enableManifestClasspath = false
     private String javaHome = null
     private List<String> excludeFromVar = ['log', 'run']
@@ -91,6 +92,14 @@ class DistributionExtension {
         this.javaHome = javaHome
     }
 
+    public void setEnv(Map<String, String> env) {
+        this.env = env
+    }
+
+    public void env(Map<String, String> env) {
+        this.env.putAll(env)
+    }
+
     public void excludeFromVar(String... excludeFromVar) {
         this.excludeFromVar.addAll(excludeFromVar)
     }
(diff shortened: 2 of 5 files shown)


uschi2000:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

_[src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy, line 95 at r1](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/137#-KUz-ibY0z_40IfOwfD4:-KUz-ibY0z_40IfOwfD5:bgologx) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/1af95f281d03fab573aefc6949433a69464e902a/src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy#L95)):_

> ``` Groovy
>     }
> 
>     public void env(Map<String, String> env) {
> ```

similar to the other methods here, let's separate this into a builder-style method and a setter:

```
env(String variableName, String value);
setEnv(Map<String, String> env);
```

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/137)_

uschi2000:
Also needs a corresponding information in readme.md

---

Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/137#-:-KUz16x3o1D52URfoQg8:btijm1f)_

uschi2000:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/137)_

uschi2000:
Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

_[src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy, line 95 at r1](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/137#-KUz-ibY0z_40IfOwfD4:-KV0lF1eq3Ul-AjMTza2:b-o0pm8s) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/1af95f281d03fab573aefc6949433a69464e902a/src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy#L95)):_
<details><summary>

<i>Previously, uschi2000 (Robert Fink) wrote…</i></summary>



> similar to the other methods here, let's separate this into a builder-style method and a setter:
> 
> ```
> env(String variableName, String value);
> setEnv(Map<String, String> env);
> ```
> 
> </details>
> how do i invoke the setEnv vs env methods given all the groovy magic around setters? in particular, what's the result of
> 
> ```
> distribution {
>   env "foo", "bar"
>   env "boom", "baz"
> }
> ```

(there is a test case for the other variables that clarifies the behaviour around this)

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/137)_

uschi2000:
:lgtm:
@markelliot any concerns?

---

Review status: 2 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/137#-:-KV2hlwCf95Da0ejIiTj:bolfm1h)_

uschi2000:
Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/137)_

markelliot:
@mglazer ping me offline if you like, but would like to understand why these env vars aren't just in the init.sh script as fixed values (launcher-static is constant for any deployment, by hand or by machine). I'm a bit confused why we're using the launcher to set env vars...the machinery we have can set these env vars (and must always be capable of this because we deploy things we don't write that may need env vars set).

mglazer:
@markelliot I was looking to find where the init.sh file is generated, and it's actually generated from this project:

https://github.com/palantir/gradle-java-distribution/blob/develop/src/main/resources/init.sh

So I guess I'm still lost as to where I'm supposed to add these env-vars, since the only extension point I seemingly have available is the launcher-static.yml

mglazer:
So should I just template in the env-vars in `init.sh`? That seems like a less than great solution since the current templating solution wouldn't do a great job of escaping or ensuring the shell script is valid after fill ins.

markelliot:
Sorry, sure, merge this. I don't see any other way to accomplish this, but I think it's still a confusing place to offer control. Basically, not sure why absence of an environment variable isn't causing a correct default to be used, so inclusion in launcher-static seems…strange.

PR #153, opened 2016-12-06
Add a refresh command

sfackler:
Intended to support managed live reload

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/gradle-java-distribution/153)

### DIFF
--- a/src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
+++ b/src/main/groovy/com/palantir/gradle/javadist/DistributionExtension.groovy
@@ -38,6 +38,7 @@ class DistributionExtension {
     private String mainClass
     private List<String> args = []
     private List<String> checkArgs = []
+    private List<String> refreshArgs = []
     private List<String> defaultJvmOpts = []
     private Map<String,String> env = [:]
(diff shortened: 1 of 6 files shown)


markelliot:
I think we're trying to move away from a world where we add more options to `init.sh`. I assume you want the deployability software we run to have a configurable location to invoke, that could happen with another shell script calling go-java-launcher with launcher-refresh.yml, rather than invoking the init script.

markelliot:
For reference, we're currently talking about ways to remove the `check` option.

sfackler:
Yep, if the deployability software allowed the refresh command to be configured then this would be unnecessary.

markelliot:
Interesting, thought it was configurable. Worth filing a ticket with those folks, would prefer to hold on this until we get that feedback.

sfackler:
Filed an issue on that tracker and cc'd you on it.

uschi2000:
Can this be closed?

sfackler:
Has the upstream issue been resolved?

uschi2000:
I haven't done any investigation

sfackler:
Handled by #195

PR #155, opened 2016-12-07
Re-run LaunchConfigTask when classpath changes, fixes #144

### LINKED ISSUE #144: launcher-static.yml not being updated
While trying to run our ETE tests, if our commit hash changes, the launcher-static.yml isn't updated. This causes our ETE tests to fail because the tgz has the new jars but the launcher-static.yml is referencing the old ones.

@hsaraogi @jeremyk-91

uschi2000:
This change is [Reviewable](https://reviewable.io/reviews/palantir/gradle-java-distribution/155)

### DIFF
--- a/src/main/groovy/com/palantir/gradle/javadist/JavaDistributionPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/javadist/JavaDistributionPlugin.groovy
@@ -19,6 +19,7 @@ import com.palantir.gradle.javadist.tasks.*
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
+import org.gradle.api.plugins.JavaPlugin
 
 class JavaDistributionPlugin implements Plugin<Project> {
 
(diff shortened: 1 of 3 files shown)


uschi2000:
@lee467 @mglazer the plugin broke as a consequence of adding `@Input` annotations, see #144 . suggestions for how to best fix this? I provided a test case that shows the issue.

uschi2000:
OK, here's an implementation that fixes the bug. Let me know what you think, @lee467 @mglazer

mglazer:
One thing that's weird, is that this shouldn't have blocked anyone's release, to my knowledge, because gradlew clean distTar should have done the right thing no?

uschi2000:
With clean yes, without clean, no, I guess.

uschi2000:
Any comments on this fix?

mglazer:
I trust that it works. I'm not certain that the test is testing the right thing for #144 though. 

I also fundamentally believe that as a core principle, build scripts should run clean as part of their process, but it's good to at least fix this.

👍

PR #157, opened 2017-01-20
(WIP) Refactor existing code and add support for asset bundle generation.

darora:
Rough draft; need to look it over again. Does the following things--

- Refactors out a `BaseDistributionExtension` with properties common to all slsv2 product types. This is inherited by an `AssetDistr...` and an `ServiceDistr...`, the latter being the existing class
- Moves packages; `com.palantir.gradle.javadist` -> `com.palantir.gradle.dist.(service|asset)`
- Lastly, adds a `asset-distribution` plugin which accepts mappings of `sourceDir, relocatedDir` that will be used to copy static content into the final asset bundle. As with the java dist plugin, the final tarball is created at `${project.buildDir}/build/distributions/${serviceName}-${versionName}.sls.tgz`.

Note that I'm not renaming properties on the extension e.g. `serviceName` to be the more consistent `productName` (per slsv2) in the interest of not breaking existing users of the plugin. There is also an assumption here that we're not going to produce multiple slsv2 product type dists from the same gradle subproject (I re-use the `distTar` and `distribution` names), which should be ok as by convention we only have one `configuration.yml` per subproject anyway.

Happy to break it up as appropriate.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/gradle-java-distribution/157)

### DIFF
--- a/src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.groovy
+++ b/src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.groovy
@@ -1,0 +1,61 @@
+package com.palantir.gradle.dist
+
+import org.gradle.api.Project
+
+class BaseDistributionExtension {
+
+        private static final List<String> VALID_PRODUCT_TYPES = [
+                "service.v1",
+                "asset.v1",
+                "daemon.v1"
+        ]
+
+        private final Project project
+        private String serviceGroup
+        private String serviceName
+        private String productType = "service.v1"
+        private Map<String, Object> manifestExtensions = [:]
+
+        public BaseDistributionExtension(Project project) {
+                this.project = project;
+        }
+
+        public void serviceName(String serviceName) {
+                this.serviceName = serviceName
+        }
+
+        void serviceGroup(String serviceGroup) {
+                this.serviceGroup = serviceGroup
+        }
+
+        public void setManifestExtensions(Map<String, Object> manifestExtensions) {
+                this.manifestExtensions = manifestExtensions;
+        }
+
+        public void manifestExtensions(Map<String, Object> manifestExtensions) {
+                this.manifestExtensions.putAll(manifestExtensions)
+        }
+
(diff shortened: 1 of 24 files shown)


src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.groovy
```diff
@@ -0,0 +1,61 @@
...
+
+        public String getServiceGroup() {
+                return serviceGroup ?: project.group
+        }
+
+        public Map<String, Object> getManifestExtensions() {
+                return this.manifestExtensions;
+        }
+
+        public String getProductType() {
```
darora, line 51: make spacing consistent between files

src/main/groovy/com/palantir/gradle/dist/service/JavaDistributionPlugin.groovy
```diff
@@ -73,6 +82,7 @@ class JavaDistributionPlugin implements Plugin<Project> {
             manifest.configure(
                     distributionExtension.serviceName,
                     distributionExtension.serviceGroup,
+                    SERVICE_PRODUCT_TYPE,
```
darora, line 85: use extension value?

src/main/groovy/com/palantir/gradle/dist/service/tasks/CreateManifestTask.groovy
```diff
@@ -32,6 +32,9 @@ class CreateManifestTask extends DefaultTask {
     String serviceGroup
 
     @Input
+    String productType
+
+    @Input
     Map<String, Object> manifestExtensions
 
     CreateManifestTask() {
```
darora, line 40: can be moved out of service specific dir

uschi2000:
Can you separate the refactor from the feature PR, please? This is a bit
too bulky to review, sorry.

darora:
closing in favour of refactor in #158 ; will put up the feature after.

PR #159, opened 2017-01-23
Clean up stale PID files in init.sh and exit 1

JacekLach:
If the process identified by the pid file is dead, or was replaced by another process, we should consistently report the service as not running (1), and clean up the stale PID file.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/gradle-java-distribution/159)

### DIFF
--- a/src/main/resources/init.sh
+++ b/src/main/resources/init.sh
@@ -53,7 +53,7 @@ STATIC_LAUNCHER_CHECK_CONFIG="service/bin/launcher-check.yml"
 
 case $ACTION in
 start)
-    if service/bin/init.sh status > /dev/null 2>&1; then
+    if service/bin/init.sh status &> /dev/null; then
         echo "Process is already running"
         exit 0
     fi
@@ -95,7 +95,7 @@ status)
 ;;
 stop)
     printf "%-50s" "Stopping '$SERVICE'..."
-    if service/bin/init.sh status > /dev/null 2>&1; then
+    if service/bin/init.sh status &> /dev/null; then
         PID=$(cat $PIDFILE)
         kill $PID
         COUNTER=0
@@ -124,7 +124,7 @@ stop)
     fi
 ;;
 console)
-    if service/bin/init.sh status > /dev/null 2>&1; then
+    if service/bin/init.sh status &> /dev/null; then
         echo "Process is already running"
         exit 1
     fi
--- a/src/main/resources/init.sh
+++ b/src/main/resources/init.sh
@@ -18,7 +18,15 @@
 is_process_active() {
    local PID=$1
    ps $PID > /dev/null;
-   echo $?
+   return $?
+}
+
+is_process_service() {
+  local PID=$1
+  local SERVICE_NAME=$2
+  # trailing '=' prevents a header line
+  ps -o command= $PID | grep -q "$SERVICE"
+  return $?
 }
 
 # Everything in this script is relative to the base directory of an SLSv2 distribution
@@ -45,8 +53,7 @@ STATIC_LAUNCHER_CHECK_CONFIG="service/bin/launcher-check.yml"
 
 case $ACTION in
 start)
-    service/bin/init.sh status > /dev/null 2>&1
-    if [[ $? == 0 ]]; then
+    if service/bin/init.sh status > /dev/null 2>&1; then
         echo "Process is already running"
         exit 0
     fi
@@ -56,9 +63,10 @@ start)
     mkdir -p "var/log"
     mkdir -p "var/run"
     PID=$($LAUNCHER_CMD $STATIC_LAUNCHER_CONFIG $CUSTOM_LAUNCHER_CONFIG > var/log/$SERVICE-startup.log 2>&1 & echo $!)
+    # always write $PIDFILE so that `init.sh status` for a service that crashed when starting will return 1, not 3
+    echo $PID > $PIDFILE
     sleep 1
-    if [ $(is_process_active $PID) -eq 0 ]; then
-        echo $PID > $PIDFILE
+    if is_process_service $PID $SERVICE; then
         printf "%s\n" "Started ($PID)"
         exit 0
     else
@@ -70,15 +78,14 @@ status)
     printf "%-50s" "Checking '$SERVICE'..."
     if [ -f $PIDFILE ]; then
         PID=$(cat $PIDFILE)
-        if [[ $(is_process_active $PID) -eq 0 ]]; then
-            ps -o command $PID | grep -q "$SERVICE"
-            if [[ $? == 0 ]]; then
-                printf "%s\n" "Running ($PID)"
-                exit 0
-            else
-                printf "%s\n" "Warning, Pid $PID appears to not correspond to service $SERVICE"
-            fi
+        if is_process_service $PID $SERVICE; then
+          printf "%s\n" "Running ($PID)"
+          exit 0
+        elif is_process_active $PID; then
+          printf "%s\n" "Warning, Pid $PID appears to not correspond to service $SERVICE"
+          # fallthrough to generic 'process dead but pidfile exists'
         fi
+
         printf "%s\n" "Process dead but pidfile exists."
         exit 1
     else
@@ -88,12 +95,11 @@ status)
 ;;
 stop)
     printf "%-50s" "Stopping '$SERVICE'..."
-    service/bin/init.sh status > /dev/null 2>&1
-    if [[ $? == 0 ]]; then
+    if service/bin/init.sh status > /dev/null 2>&1; then
         PID=$(cat $PIDFILE)
         kill $PID
         COUNTER=0
-        while [ $(is_process_active $PID) -eq "0" -a "$COUNTER" -lt "240" ]; do
+        while is_process_service $PID $SERVICE && [ "$COUNTER" -lt "240" ]; do
             sleep 1
             let COUNTER=COUNTER+1
             if [ $((COUNTER%5)) == 0 ]; then
@@ -103,7 +109,7 @@ stop)
                 printf "%s\n" "Waiting for '$SERVICE' ($PID) to stop"
             fi
         done
-        if [[ $(is_process_active $PID) -eq "0" ]]; then
+        if is_process_service $PID $SERVICE; then
             printf "%s\n" "Failed"
             exit 1
         else
@@ -118,8 +124,7 @@ stop)
     fi
 ;;
 console)
-    service/bin/init.sh status > /dev/null 2>&1
-    if [[ $? == 0 ]]; then
+    if service/bin/init.sh status > /dev/null 2>&1; then
         echo "Process is already running"
         exit 1
     fi
--- a/src/main/resources/init.sh
+++ b/src/main/resources/init.sh
@@ -77,11 +77,11 @@ status)
                 exit 0
             else
                 printf "%s\n" "Warning, Pid $PID appears to not correspond to service $SERVICE"
-                exit 0
             fi
         fi
-        printf "%s\n" "Process dead but pidfile exists"
-        exit 1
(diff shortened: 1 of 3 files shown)


JacekLach:
Without this attempting to `./init.sh restart ...` a service will fail if process named by $PID is not killable by palantir, or kill some innocent process while thinking that it's killing $SERVICE

src/main/resources/init.sh
```diff
@@ -77,11 +77,11 @@ status)
...
             else
                 printf "%s\n" "Warning, Pid $PID appears to not correspond to service $SERVICE"
-                exit 0
             fi
         fi
-        printf "%s\n" "Process dead but pidfile exists"
-        exit 1
+        printf "%s\n" "Process dead but pidfile exists. Cleaning up."
+        rm -f $PIDFILE
+        exit 3
```
uschi2000: why this change?

uschi2000 (commented):
Mid-term, I think we need to test the init scripts :(

src/main/resources/init.sh
JacekLach: Mostly for consistency with the other 'service not running' branch below. Happy to change it back if the contract is meant to be `0 if running, non-0 otherwise`, though that leaves no way to message internal errors.

From a API perspective, it seems strange to expose whether the pid file existed or not; that's an implementation detail of the script, rather than anything actually useful to any caller.

uschi2000: OK, cool.

uschi2000: @bmoylan I assume you're cool with this?

uschi2000: On the actual change: It's not so clear that clean-up should be performed when calling `status`; `start` would seem to be a more natural place, no?

JacekLach: We don't want `init.sh stop` to attempt killing a PID that's not the service in question. `status` is called at the beginning of every other command; it seems like the most sensible spot unless we want to pull out pid file handling to a separate function completely

uschi2000: Ah, I forgot that status gets called at the beginning everywhere. Cool, convinced. Need to fix the build though.

JacekLach: Ack, will correct it tomorrow.

JacekLach: (looks like the test condition is inverted too: `java.lang.AssertionError: Expected command 'dist/foo/service/bin/init.sh status' to be successful expected:<3> but was:<0>` - expected and actual seem to be swapped)

bmoylan: 👎  the existing implementation is per the Linux Base Spec: https://refspecs.linuxbase.org/LSB_3.0.0/LSB-PDA/LSB-PDA/iniscrptact.html

> If the status action is requested, the init script will return the following exit status codes.
> 
> 0	program is running or service is OK
> 1	program is dead and /var/run pid file exists
> 2	program is dead and /var/lock lock file exists
> 3	program is not running
> 4	program or service status is unknown
> 5-99	reserved for future LSB use
> 100-149	reserved for distribution use
> 150-199	reserved for application use
> 200-254	reserved

JacekLach: Well, strictly speaking the pid file is not in `/var/run` anyway, but sure, can change this back to 1.

bmoylan: thanks, sorry to be pendantic but there's sometimes a difference between a crash and an intended stop and that return code distinguishes the two

JacekLach:
Requesting re-review. Since @bmoylan raised the point of keeping our return codes in line with the LSB, I've decided to:

1. Not clean up the pidfile in `status` calls, so that subsequent status queries return the same result
2. Always generate the pidfile in `start`, so that we report the right result in a `status` call following a failed start (crashed vs stopped successfully)
3. Consistently treat the service as running only when the pid exists *and* it matches `$SERVICE`

uschi2000:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/test/groovy/com/palantir/gradle/dist/service/JavaDistributionPluginTests.groovy, line 131 at r2](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/159#-KbGUooYD6PIFxt_iFRQ:-KbGUooYD6PIFxt_iFRR:b-re6p2v) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/25029dce29b4388ebf904d139d8d1703121654be/src/test/groovy/com/palantir/gradle/dist/service/JavaDistributionPluginTests.groovy#L131)):*
> ```Groovy
>         runSuccessfully(':build', ':distTar', ':untar')
>         createFile('dist/service-name-0.1/var/run/service-name.pid')
>         exec('/bin/bash', '-c', 'sleep 30 & echo $! > dist/service-name-0.1/var/run/service-name-0.1.pid')
> ```

I don't understand this test, can you explain please?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/159)*

JacekLach:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/test/groovy/com/palantir/gradle/dist/service/JavaDistributionPluginTests.groovy, line 131 at r2](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/159#-KbGUooYD6PIFxt_iFRQ:-KbGaXxB0cQbf-lFqHmy:betitf1) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/25029dce29b4388ebf904d139d8d1703121654be/src/test/groovy/com/palantir/gradle/dist/service/JavaDistributionPluginTests.groovy#L131)):*
<details><summary><i>Previously, uschi2000 (Robert Fink) wrote…</i></summary><blockquote>

I don't understand this test, can you explain please?
</blockquote></details>

Sure; the main subject here is what the init script does when it has a pidfile, but the process at that pid does not match what we expect. For that purpose, we generate a new process (`sleep 30`, which should be long enough for any testing we do), and put its pid into `$SERVICE.pid` to fake a pid reuse situation.

The init scripts then are validated that:
1. `init.sh status` notices that the process that is running is not `$SERVICE`
2. the status persists against runs (doesn't clean the pidfile itself)
3. `init.sh stop` cleans the pidfile and doesn't kill the clashing process ('service not running' vs 'Stopping $SERVICE... OK')
4. `init.sh status` returns 'not running' after `stop` has succeeded

Is anything else unclear?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/159)*

uschi2000:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/test/groovy/com/palantir/gradle/dist/service/JavaDistributionPluginTests.groovy, line 131 at r2](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/159#-KbGUooYD6PIFxt_iFRQ:-KbGdghxf5U0vk2RcJM3:b-1kyufa) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/25029dce29b4388ebf904d139d8d1703121654be/src/test/groovy/com/palantir/gradle/dist/service/JavaDistributionPluginTests.groovy#L131)):*
<details><summary><i>Previously, JacekLach (Jacek Lach) wrote…</i></summary><blockquote>

Sure; the main subject here is what the init script does when it has a pidfile, but the process at that pid does not match what we expect. For that purpose, we generate a new process (`sleep 30`, which should be long enough for any testing we do), and put its pid into `$SERVICE.pid` to fake a pid reuse situation.

The init scripts then are validated that:
1. `init.sh status` notices that the process that is running is not `$SERVICE`
2. the status persists against runs (doesn't clean the pidfile itself)
3. `init.sh stop` cleans the pidfile and doesn't kill the clashing process ('service not running' vs 'Stopping $SERVICE... OK')
4. `init.sh status` returns 'not running' after `stop` has succeeded

Is anything else unclear?
</blockquote></details>

OK, cool. Why is the expected message for `status` different in the first two calls?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/159)*

bmoylan:
I'm good with the init.sh changes but didn't look as closely at the tests. Thanks Jacek!

---

Reviewed 1 of 3 files at r2.
Review status: 1 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/main/resources/init.sh, line 29 at r2](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/159#-KbGhHz3BM0QRprUybu1:-KbGhHz3BM0QRprUybu2:b31d1fw) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/25029dce29b4388ebf904d139d8d1703121654be/src/main/resources/init.sh#L29)):*
> ```Shell
>   # trailing '=' prevents a header line
>   ps -o command= $PID | grep -q "$SERVICE"
>   return $?
> ```

nit: you don't need `return $?`, it will automatically return the status of the last command in the function

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/159#-:-KbGlJ6bKkptGfJ8RjvU:blcnjvs)*

JacekLach:
Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/main/resources/init.sh, line 29 at r2](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/159#-KbGhHz3BM0QRprUybu1:-KbKzTLSn1b_Gh-NqPvD:b-no1m10) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/25029dce29b4388ebf904d139d8d1703121654be/src/main/resources/init.sh#L29)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

nit: you don't need `return $?`, it will automatically return the status of the last command in the function
</blockquote></details>

I like doing this for clarity, to differentiate functions that return via exit code from functions that return via stdout.

---

*[src/test/groovy/com/palantir/gradle/dist/service/JavaDistributionPluginTests.groovy, line 131 at r2](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/159#-KbGUooYD6PIFxt_iFRQ:-KbKzfrZcXaQRk2R2C1g:b-3ddn4l) ([raw file](https://github.com/palantir/gradle-java-distribution/blob/25029dce29b4388ebf904d139d8d1703121654be/src/test/groovy/com/palantir/gradle/dist/service/JavaDistributionPluginTests.groovy#L131)):*
<details><summary><i>Previously, uschi2000 (Robert Fink) wrote…</i></summary><blockquote>

OK, cool. Why is the expected message for `status` different in the first two calls?
</blockquote></details>

They both should print both messages (unless >30s expire before the second one is ran and the `sleep 30` process terminates); I went with this over a regex because it was simpler to express. Groovy is a pain :P

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/159)*

JacekLach:
Rebased onto develop to fix merge conflicts; also removed unused test function `execIgnoreResult`

---

Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/159#-:-KbL4mpbBWBndkl6Go0I:bnno3kt)*

uschi2000:
Reviewed 1 of 3 files at r2, 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/159)*

uschi2000:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/gradle-java-distribution/159#-:-KbL6p9CBp7CQnnJw1Nw:bnfp4nl)*

uschi2000 (approved):

PR #165, opened 2017-01-25
gradle compatible pid status check

sandorw:
In our docker containers, `ps $PID` always returns successfully and just lists the same output as `ps`. This means that any docker based tests that check service status or restart services fail.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/gradle-java-distribution/165)


uschi2000 (commented):
I'm hesitant to take this change without tests, sorry. (Yes, even if this is currently broken in Docker.)

sandorw:
That's reasonable, I'll work on some.

PR #167, opened 2017-01-25
CreateManifestTask checks that project.version is valid SLS version s…

uschi2000:
…tring

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/gradle-java-distribution/167)

### DIFF
--- a/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
+++ b/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
@@ -16,6 +16,7 @@
 
 package com.palantir.gradle.dist.tasks
 
+import com.palantir.gradle.dist.SlsProductVersions
 import com.palantir.gradle.dist.service.JavaDistributionPlugin
 import groovy.json.JsonOutput
 import org.gradle.api.DefaultTask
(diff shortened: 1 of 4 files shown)


uschi2000:
FYI @aditkumar @briandorne

src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
```diff
@@ -44,7 +45,11 @@ class CreateManifestTask extends DefaultTask {
 
     @Input
     public String getProjectVersion() {
-        return String.valueOf(project.version)
+        def stringVersion = String.valueOf(project.version)
+        if (!SlsProductVersions.isValidVersion(stringVersion)) {
```
markelliot, line 49: Probably worth a noisy-ish non-blocking warning if this isn't also an orderable version.

markelliot (approved):
Your choice on solving my one comment.

src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
uschi2000: done

PR #172, opened 2017-01-30
Apply gradle-java-distribution --> sls-packaging renaming

uschi2000:
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/172)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,7 +31,7 @@ if (System.env.CIRCLE_TEST_REPORTS) {
     test.reports.junitXml.destination = new File(System.env.CIRCLE_TEST_REPORTS, getName())
 }
 
-group 'com.palantir.gradle.javadist'
+group 'com.palantir.gradle.sls-packaging'
 version gitVersion()
 
 task sourceJar(type: Jar) {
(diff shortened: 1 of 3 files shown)


markelliot:
you're missing the associated META-INF resource changes?

build.gradle
```diff
@@ -31,7 +31,7 @@ if (System.env.CIRCLE_TEST_REPORTS) {
     test.reports.junitXml.destination = new File(System.env.CIRCLE_TEST_REPORTS, getName())
 }
 
-group 'com.palantir.gradle.javadist'
+group 'com.palantir.gradle.sls-packaging'
```
markelliot, line 34: note that when this changes we'll need to publish to a different bintray destination and we'll need to get that whitelisted to jcenter.

uschi2000:
Yep, aware of the bintray thing

uschi2000:
Was going to do the meta INF thing in a separate PR. Artifact names and
plugin names are unrelated.

markelliot:
The META-INF thing maps plugin name to implementation. If you change the plugin name, the resource also needs to change.

Sent from my iPhone

On Jan 30, 2017, at 12:50 PM, Robert Fink <notifications@github.com<mailto:notifications@github.com>> wrote:

Was going to do the meta INF thing in a separate PR. Artifact names and
plugin names are unrelated.

uschi2000:
Ah, seems like the gradle plugin publishing was borked already after we introduced the second plugin. Any way, I reverted the name change for the plugin here.

build.gradle
uschi2000: yep

PR #183, opened 2017-02-05
Support for setting the productType to daemon.v1

SerialVelocity:
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/183)

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.groovy
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.groovy
@@ -7,6 +7,7 @@ class BaseDistributionExtension {
 
     private static final Set<String> VALID_PRODUCT_TYPES = [
             "service.v1",
+            "daemon.v1",
             "asset.v1"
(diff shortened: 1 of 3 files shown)


SerialVelocity:
I don't think any other changes are required for this to work, right? @uschi2000

uschi2000:
I missed this in the previous PR, unfortunately: How hard would it be to assert that the asset plugin produces asset.v1 only and the service plugin produces service&daemon only?

SerialVelocity:
I can look into that but the tests don't pass locally for me so it's hard to see if it is working properly.

uschi2000:
Are you building in Macos?

SerialVelocity:
Yes. Thinking about it more, it might be to do with me having my gradle set to be parallel by default in my `~/.gradle/gradle.properties`

I'm also using a hack to work around this for now:
```
distribution.class.superclass.superclass.metaClass.setAttribute(distribution, 'productType', 'daemon.v1')
```

uschi2000:
is this still needed? if so, please rebase onto develop.

SerialVelocity:
We're removing our daemon soon so no.

PR #186, opened 2017-02-08
Allow version matchers in service dependencies

uschi2000:
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/186)

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.groovy
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.groovy
@@ -59,7 +59,6 @@ class BaseDistributionExtension {
     }
 
     void serviceDependency(ServiceDependency dependency) {
-        dependency.verifyVersions()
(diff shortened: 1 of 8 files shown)


uschi2000:
Merging, let me know if anything's missing.

IanMair:
@uschi2000 - I could be missing something, but it looks like service dependencies added with the `void productDependency(Closure closure)` method on line [55 of BaseDistributionExtension.groovy](https://github.com/palantir/sls-packaging/blob/develop/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.groovy#L55) aren't going to get validated anymore? It's calling the constructor with no arguments (which obviously doesn't do the validation), but since `isValid` was moved into the constructor with arguments seems like we might be missing validation now - unless `ConfigureUtil` is somehow using the correct constructor?

Other than that, LGTM

uschi2000:
Ian, I shuffled this a bit and the check is now in createManifestTask. I guess it makes more sense in the extension, will revert to doing that.

PR #190, opened 2017-02-14
Assert that UTF8 works

uschi2000:
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/190)


ellisjoe:
think your system default is likely UTF-8 where the SO issue was related to the system default being ISO-8859-1

jmcampanini:
even if we can't simulate the error via test, is it worth setting the default value?

PR #191, opened 2017-02-17
Change VersionMatcher format

mnazbro:
Version matcher must:
 1. Contain at least one wildcard (x).
 2. Must not contain a wildcard in a larger position than a
    a non-wildcard version (e.g. x.1.2 is not valid)

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/191)


mnazbro:
@gdearment @uschi2000 @bmoylan

uschi2000:
Looks good. Need to make corresponding change to sls-spec too.

uschi2000:
is this still relevant, @mnazbro ?

mnazbro:
This change is still relevant. The sls-spec has already been changed accordingly. A single regex was used instead in the spec.

uschi2000 (approved):

uschi2000:
Looks good, but doesn't compile.

mnazbro:
Updated with fixed code. I tried modifying it in the GitHub UI and that caused the compilation error.

PR #202, opened 2017-03-14
Allow 1.2.3.dirty as non-orderable version

uschi2000:
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/202)

### DIFF
--- a/sls-versions/src/main/java/com/palantir/slspackaging/versions/SlsProductVersions.java
+++ b/sls-versions/src/main/java/com/palantir/slspackaging/versions/SlsProductVersions.java
@@ -6,7 +6,7 @@ import java.util.regex.Pattern;
  * Utility methods for checking whether version strings are valid SLS version strings.
  */
 public class SlsProductVersions {
-    private static final Pattern NON_ORDERABLE_VERSION = Pattern.compile("^[0-9]+\\.[0-9]+\\.[0-9]+((-[a-z0-9-.]*)|(\\.dirty))$");
+    private static final Pattern NON_ORDERABLE_VERSION = Pattern.compile("^[0-9]+\\.[0-9]+\\.[0-9]+(-[a-z0-9-]+)?(\\.dirty)?$");
     private static final Pattern[] ORDERABLE_VERSION = new Pattern[]{
             Pattern.compile("^[0-9]+\\.[0-9]+\\.[0-9]+-[0-9]+-g[a-f0-9]+$"),
             Pattern.compile("^[0-9]+\\.[0-9]+\\.[0-9]+$"),
--- a/sls-versions/src/main/java/com/palantir/slspackaging/versions/SlsProductVersions.java
+++ b/sls-versions/src/main/java/com/palantir/slspackaging/versions/SlsProductVersions.java
@@ -6,7 +6,7 @@ import java.util.regex.Pattern;
  * Utility methods for checking whether version strings are valid SLS version strings.
  */
 public class SlsProductVersions {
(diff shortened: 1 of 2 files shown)


uschi2000:
fyi @derenrich

sls-versions/src/test/java/com/palantir/slspackaging/versions/SlsProductVersionsTest.java
```diff
@@ -23,11 +23,15 @@ public void testOrderableVersionDetection() {
...
         assertThat(SlsProductVersions.isNonOrderableVersion("2.0.0-beta1")).isTrue();
+        assertThat(SlsProductVersions.isNonOrderableVersion("2.0.0-beta1.dirty")).isTrue();
         assertThat(SlsProductVersions.isNonOrderableVersion("2.0.0-rc1")).isTrue();
+        assertThat(SlsProductVersions.isNonOrderableVersion("2.0.0-rc1.dirty")).isTrue();
         assertThat(SlsProductVersions.isNonOrderableVersion("2.0.0-beta1")).isTrue();
+        assertThat(SlsProductVersions.isNonOrderableVersion("2.0.0-beta1.dirty")).isTrue();
         assertThat(SlsProductVersions.isNonOrderableVersion("2.0.0-foo")).isTrue();
+        assertThat(SlsProductVersions.isNonOrderableVersion("2.0.0-foo.dirty")).isTrue();
         assertThat(SlsProductVersions.isNonOrderableVersion("2.0.0-foo-g20-gaaaaaa")).isTrue();
         assertThat(SlsProductVersions.isNonOrderableVersion("2.0.0-foo-g20-gaaaaaa.dirty")).isTrue();
```
markelliot: How about a test that covers 1.0.0.dirty?

sls-versions/src/main/java/com/palantir/slspackaging/versions/SlsProductVersions.java
```diff
@@ -6,7 +6,7 @@
  * Utility methods for checking whether version strings are valid SLS version strings.
  */
 public class SlsProductVersions {
-    private static final Pattern NON_ORDERABLE_VERSION = Pattern.compile("^[0-9]+\\.[0-9]+\\.[0-9]+-[a-z0-9-.]*$");
+    private static final Pattern NON_ORDERABLE_VERSION = Pattern.compile("^[0-9]+\\.[0-9]+\\.[0-9]+((-[a-z0-9-.]*)|(\\.dirty))$");
```
markelliot: Surprised this works -- would expect the or at the end to allow either but not both, and we explicitly want to allow both. Maybe clearer to replace "|(\\.dirty)" with "(\\.dirty)?" ?

uschi2000: note that `[a-z0-9-.]` accepts `.dirty` already

sls-versions/src/test/java/com/palantir/slspackaging/versions/SlsProductVersionsTest.java
uschi2000: added

sls-versions/src/main/java/com/palantir/slspackaging/versions/SlsProductVersions.java
markelliot: ah, I missed that. I think I'd prefer to remove `.` support -- we added it just to support `.dirty` anyway.

markelliot: ie we want to allow `1.0.0.dirty` and `1.0.0-1-gabcdef.dirty` but did not really want to allow `1.0.0-rc.1`

PR #204, opened 2017-03-21
Make gcJvmOpts configurable on JavaServiceDistributionExtension

iamdanfox:
@uschi2000 I've hit some problems using sls-packaging 2.0.0 on circle:

```
Unrecognized VM option 'CrashOnOutOfMemoryError'
Did you mean 'OnOutOfMemoryError=<value>'?
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

This seems to arise because Circle's `oracklejdk8` is not new enough to support this option.  To enable a workaround, I've made these gc options configurable.  This PR doesn't change the defaults :)

cc @skedida

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/204)


markelliot:
I'd prefer not to accept this. Some alternatives available to you:
 - use Circle's customizations to upgrade your Java dist
 - run tests that run go-java-launcher in a container

markelliot (changes requested):
(feedback left on main thread)

uschi2000:
that was my feedback too.

PR #205, opened 2017-03-22
Jc/switch to runtime classpath

### LINKED ISSUE #201: gradle 3.4 java-library support using runtimeClasspath
> When resolving the runtime classpath for Java applications, Gradle will now use the runtimeClasspath configuration instead of the runtime configuration. If you previously attached resolution rules to runtime, you should apply them to runtimeClasspath instead or apply them to all configurations.

in Gradle 3.4 ([release notes][1]), they've moved to using the `runtimeClasspath` as the configuration rather than `runtime`. i believe this means enhancing this [bit of code][2] in the `DistTarTask` to include the `runtimeClasspath`.

i believe you can check if the configuration exists by name and then add it conditionally, which should let it work with pas versions of gradle.

would a PR for this make sense?

[1]: https://docs.gradle.org/3.4/release-notes.html#runtimeclasspath-used-instead-of-runtime
[2]: https://github.com/palantir/sls-packaging/blob/develop/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/DistTarTask.groovy#L61-L64

jmcampanini:
fixes #201

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/205)

### DIFF
--- a/gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/ServiceDistributionPluginTests.groovy
+++ b/gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/ServiceDistributionPluginTests.groovy
@@ -599,6 +599,68 @@ class ServiceDistributionPluginTests extends GradleTestSpec {
         result.output.contains("The plugins 'com.palantir.sls-asset-distribution' and 'com.palantir.sls-java-service-distribution' cannot be used in the same Gradle project.")
     }
 
+    def 'uses the runtimeClasspath so api and implementation configurations work with java-library plugin'() {
+        given:
+        helper.addSubproject('parent', '''
+            plugins {
+                id 'com.palantir.sls-java-service-distribution'
+                id 'java'
+            }
+            version '0.0.1'
+            distribution {
+                serviceName "service-name"
+                mainClass "dummy.service.MainClass"
+                args "hello"
+                enableManifestClasspath true
+            }
+            repositories { jcenter() }
+            dependencies {
+                implementation project(':child')
+                compile 'org.mockito:mockito-core:2.7.22'
+            }
+            // most convenient way to untar the dist is to use gradle
+            task untar (type: Copy) {
+                from tarTree(resources.gzip("${buildDir}/distributions/service-name-0.0.1.sls.tgz"))
+                into "${projectDir}/dist"
+                dependsOn distTar
+            }
+        ''')
+
+        helper.addSubproject('child', '''
+            plugins {
+                id 'java-library'
+            }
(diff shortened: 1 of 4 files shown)


gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/ServiceDistributionPluginTests.groovy
```diff
@@ -497,8 +497,10 @@ class ServiceDistributionPluginTests extends GradleTestSpec {
         def classpathJar = file('dist/service-name-0.0.1/service/lib/').listFiles()
                 .find({ it.name.endsWith("-manifest-classpath-0.0.1.jar") })
         classpathJar.exists()
-        readFromZip(classpathJar, "META-INF/MANIFEST.MF")
-                .contains('Class-Path: guava-19.0.jar produces-manifest-') // etc
+        String fromZip = readFromZip(classpathJar, "META-INF/MANIFEST.MF")
+        fromZip.contains('Class-Path:')
+        fromZip.contains('guava-19.0.jar')
+        fromZip.contains('produces-manifest-')
```
uschi2000: why did this change?

jmcampanini: the old version required that `guava-19.0.jar` was the first entry:

```
 .contains('Class-Path: guava-19.0.jar produces-manifest-')
```

when using `runtimeClasspath` it changed to say `main main guava-19.0.jar`, which i assume means that the main directories were included in the classpath.

which is technically correct that it's the full classpath, but if you were expecting to only have the JAR files, then we'd need to filter the file collection to remove certain entities.

uschi2000: but there is no `main` directory in the distribution, right? it's just the jars.

jmcampanini:
@uschi2000 updated it to find all configurations found within the classpath. this should:

1. mimic the existing behavior by enumerating only configurations found in the `runtimeClasspath`
1. leave the tests the same
1. remove the `main` folders we were seeing
1. work with a `java-library` example

also added a test to ensure it works. lemme know if there's anything else necessary or if this isn't a good approach

jmcampanini:
silly me, there is a configuration called `runtimeClasspath`, no need to pull it from sourceSets. updated

uschi2000 (approved):

uschi2000:
This looks much better, thanks Javier!

PR #207, opened 2017-03-25
Add jar.archivePath to classpath

leeavital:
This will create a run task which includes the application jar in the
classpath. This is important if the application reads informatino from
its own jar manifest like UserAgents.fromClass() in http-remoting

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/207)

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/RunTask.groovy
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/RunTask.groovy
@@ -26,7 +26,7 @@ class RunTask {
             t.dependsOn(project.tasks.jar)
             t.group = JavaServiceDistributionPlugin.GROUP_NAME
             t.description = "Runs the specified project using configured mainClass and with default args."
-            classpath project.files(project.tasks.jar.archivePath, project.sourceSets.main.runtimeClasspath)
+            t.classpath project.files(project.tasks.jar.archivePath, project.sourceSets.main.runtimeClasspath)
         }
     }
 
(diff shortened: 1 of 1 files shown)


markelliot (approved):

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/RunTask.groovy
```diff
@@ -23,9 +23,10 @@ class RunTask {
 
     static JavaExec createRunTask(Project project, String taskName) {
         return project.tasks.create(taskName, JavaExec) { t ->
+            t.dependsOn(project.tasks.jar)
```
uschi2000, line 26: is this line really necessary? ( i think you could just use `project.tasks.jar` below to get this behavior, no?)

leeavital: That would get you the path to the jar, but wouldn't guarantee that the jar exists at task execution time.  This line will tell gradle to ensure the project's jar exists before execution.

uschi2000: I would expect that the task dependency gets added automatically when one depends on the task output (not the `archivePath`), at least this is how it works in gradle-docker, see https://github.com/palantir/gradle-docker/blob/develop/src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy#L463 . Have you verified the behavior?

leeavital: Which behavior? I've tested this against one of the projects which uses this plugin and leaving out the `dependsOn(jar)` broke it

uschi2000: OK, nvm.

PR #209, opened 2017-04-07
Make expected Java tmp directory by default

markelliot:
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/209)

### DIFF
--- a/gradle-sls-packaging/src/main/resources/init.sh
+++ b/gradle-sls-packaging/src/main/resources/init.sh
@@ -60,9 +60,8 @@ start)
     fi
     printf "%-50s" "Running '$SERVICE'..."
 
-    # ensure log and pid directories exist
-    mkdir -p "var/log"
-    mkdir -p "var/run"
+    # ensure log, pid and tmp directories exist
+    mkdir -p "var/log" "var/run" "var/data/tmp"
     PID=$($LAUNCHER_CMD $STATIC_LAUNCHER_CONFIG $CUSTOM_LAUNCHER_CONFIG > var/log/$SERVICE-startup.log 2>&1 & echo $!)
     # always write $PIDFILE so that `init.sh status` for a service that crashed when starting will return 1, not 3
     echo $PID > $PIDFILE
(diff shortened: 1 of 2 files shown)


gradle-sls-packaging/src/main/resources/init.sh
```diff
@@ -60,9 +60,8 @@ start)
     fi
     printf "%-50s" "Running '$SERVICE'..."
 
-    # ensure log and pid directories exist
-    mkdir -p "var/log"
-    mkdir -p "var/run"
+    # ensure log, pid and tmp directories exist
+    mkdir -p "var/log" "var/run" "var/data/tmp"
```
uschi2000, line 64: how does this relate to the tmp stuff in DistTarTask?

uschi2000: and this:
```
    def 'produce distribution bundle and check var/data/tmp is created and used for temporary files'() {
        given:
        createUntarBuildFile(buildFile)
        file('src/main/java/test/Test.java') << '''
        package test;
        import java.nio.file.Files;
        import java.io.IOException;
        public class Test {
            public static void main(String[] args) throws IOException {
                Files.write(Files.createTempFile("prefix", "suffix"), "temp content".getBytes());
            }
        }
        '''.stripIndent()

        when:
        runSuccessfully(':build', ':distTar', ':untar')

        then:
        execAllowFail('dist/service-name-0.0.1/service/bin/init.sh', 'start')
        file('dist/service-name-0.0.1/var/data/tmp').listFiles().length == 1
        file('dist/service-name-0.0.1/var/data/tmp').listFiles()[0].text == "temp content"
    }
```

markelliot: can likely remove the distTar steps. The issue with depending on distTar is that it can only make this directory during install, but we need this directory exist for every application run.

uschi2000:
OK.

gradle-sls-packaging/src/main/resources/init.sh
uschi2000: are you doing that in this PR or a separate one?

markelliot: I can do it here, but will have to come back to it.

markelliot: done

uschi2000 (approved):

uschi2000:
Turns out this is broken for anyone calling into the go-java-launcher directly (e.g., docker). cf https://github.com/palantir/go-java-launcher/issues/32

markelliot:
yeah, I think our best bet is actually fix in go-java-launcher and pull tmp directory config up to the yaml.

Issue #208, opened 2017-03-30
both jar and classes contained in tgz


pnepywoda:
Using gradle 3.4.1, running `distTar` writes both the jar and classes to the sls.tgz:

```
build-runner-0.0.7-4-g8dd04f9/service/lib/build-runner-0.0.7-4-g8dd04f9.jar
build-runner-0.0.7-4-g8dd04f9/service/lib/com/
build-runner-0.0.7-4-g8dd04f9/service/lib/com/palantir/
build-runner-0.0.7-4-g8dd04f9/service/lib/com/palantir/foundry/
build-runner-0.0.7-4-g8dd04f9/service/lib/com/palantir/foundry/build2/
build-runner-0.0.7-4-g8dd04f9/service/lib/com/palantir/foundry/build2/runner/
build-runner-0.0.7-4-g8dd04f9/service/lib/com/palantir/foundry/build2/runner/store/
build-runner-0.0.7-4-g8dd04f9/service/lib/com/palantir/foundry/build2/runner/store/generated/
build-runner-0.0.7-4-g8dd04f9/service/lib/com/palantir/foundry/build2/runner/store/generated/BuildGroupConsistentPropertiesTable$1.class
build-runner-0.0.7-4-g8dd04f9/service/lib/com/palantir/foundry/build2/runner/store/generated/BuildGroupConsistentPropertiesTable$2.class
```

My ultimate goal here is to use `getClass().getPackage().getImplementationVersion()` metrics-relay and I think having both the jar and the classes in the tgz is probably messing this up.

uschi2000:
Looks like only generated code/classes are in the tgz?

pnepywoda:
non-generated things are definitely there
```build-runner-0.0.12-9-g06ed3bf.dirty/service/lib/com/palantir/foundry/build2/runner/store/ImmutableBuildInfo.java```

**Comment by uschi2000** [comment]
That's generated by annotation-processors.

**Comment by pnepywoda** [comment]
Ok we're getting sidetracked here. `getClass().getPackage().getImplementationVersion()` doesn't work for me...do you think it's because of some other issue?

**Comment by pnepywoda** [comment]
We also have non-generated code in the tgz
```
build-runner-0.0.18-3-ge94a520/service/lib/com/palantir/foundry/build2/runner/store/JobInfo.class
build-runner-0.0.18-3-ge94a520/service/lib/com/palantir/foundry/build2/runner/store/JobInfoPersister.class
[...]

PR #211, opened 2017-04-17
Add test assuring no class files in service/lib

markelliot:
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/211)


markelliot:
demonstrates #208 isn't caused by this library

@pnepywoda and I separately debugged and found the root-cause was including `files` based dependencies in his build scripts

uschi2000 (approved):

PR #215, opened 2017-05-08
Jc/missed fix for java lib plugin

### LINKED ISSUE #201: gradle 3.4 java-library support using runtimeClasspath

jmcampanini:
fixes #201 

this should cover migrating all `.runtime` to `.runtimeClasspath`.

this PR covers:

* `<service-name>` script (BAT script has the classpath omitted)
* `launcher-static.yml`
* `launcher-check.yml`

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/215)

### DIFF
--- a/gradle-sls-packaging/build.gradle
+++ b/gradle-sls-packaging/build.gradle
@@ -33,7 +33,3 @@ pluginBundle {
 publishPlugins.onlyIf {
     project.version ==~ /[0-9]+(\.[0-9]+)+(-rc[0-9]+)?(-alpha[0-9]+)?/
 }
-
-test {
-    include '**/ServiceDistributionPluginTests.class'
-}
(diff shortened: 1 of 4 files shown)


markelliot (approved):

jmcampanini:
w00t! ty! :)

PR #216, opened 2017-05-09
Add ability to specify mavenCoord for version range and recommended version

ghost:
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/216)

PR #217, opened 2017-05-09
Add ProductDependency#templateMaximumVersionFromMinimumVersion

ghost:
I realized that the previous attempt in #216 has a number of problems:

1. The common use case for the option added is to specify the api maven coord as the first parameter to get the group + name. Unfortunately, the name of the api maven coord has "-api" on the end of it, thus making the whole endpoint useless.
2. The fact the maximumVersion of the product dependency has a default behaviour that's implemented using the getter makes it impossible to set an null maximumVersion.
3. The product dependency is also not editable if using the added endpoint since you don't get a change to configure it like you would if you passed in a closure.

For all of these reasons, this PR:

- Reverts the commit introduced in #216.
- Implements a simpler method that can be called in a productDependency closure

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/217)


ghost:
Closing out since we can just use distribution maven coords, and code the recommended in another file

ghost:
@darora

gracew:
+1, I think the configuration that #216 requires is rather unreadable:
```
productDependency "<group>:<name>:${dependencyRecommendations.getRecommendedVersion('<group>', '<name of api>')}", dependencyRecommendations.getRecommendedVersion('<group>', '<name>')
```

it's unclear (a) what to pass as the two args and (b) what they are transformed into by the extension.

in contrast to:
```
productDependency {
        productGroup = '<group>'
        productName = '<name>'
        minimumVersion = dependencyRecommendations.getRecommendedVersion('<group>', '<api name>')
        maximumVersion = templateMaximumVersionFromMinimumVersion()
        recommendedVersion = dependencyRecommendations.getRecommendedVersion('<group>', '<dist name>')
    }
}
```

markelliot:
Why isn’t this just this:

productDependency(project.recommend(‘group:name’, ‘versions.props’), project.recommend(‘group:name’, ‘docker.props’))

ghost:
Few reasons:
0. I want to use API deps for the min/max and dists  for the recommended
1. Don't want to split versions across properties files.
2. With dependency recommender you have to apply and and name a properties
file. You can't just specify a file path, you actually have to have the
versions be part of your global recommendations.

markelliot:
Then it seems like we should simply make no changes here?

ghost:
Is maxVersion a required field? Current world does no allow it to be empty

ghost:
If you leave it empty it gets filled in for you - thought that might be
confusing. If we're fine wth that then I'll close this out and make a
corresponding readme change

markelliot:
It seems like if you don’t specify it that it shouldn’t be rendered.

ghost:
This changes that behaviour to not render it. I can also propose a change
which keeps the thing I added and has the behaviour you desrcribe above.

uschi2000:
is this still needed? if so, please rebase.

PR #220, opened 2017-05-16
Fix init.sh variable reference and output truncation

vipshek:
It appears that the is_process_service function defined in init.sh is improperly using the globally defined `$SERVICE` variable when identifying whether a given PID corresponds to a service, instead of the local `$SERVICE_NAME` variable as intended.

Additionally, running the `ps | grep` command failed for me because the output from `ps` was being truncated on my operating system (I'm adapting this script to run on IBM AIX). Adding the `ewww` flag prevents output truncation, so the `grep` can correctly find the name of the service in the `ps` output.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/220)

### DIFF
--- a/gradle-sls-packaging/src/main/resources/init.sh
+++ b/gradle-sls-packaging/src/main/resources/init.sh
@@ -25,7 +25,7 @@ is_process_service() {
   local PID=$1
   local SERVICE_NAME=$2
   # trailing '=' prevents a header line
-  ps -o command= $PID | grep -q "$SERVICE"
+  ps ewww -o command= $PID | grep -q "$SERVICE_NAME"
   return $?
 }
 


palantirtech:
Thanks for your interest in [palantir/sls-packaging](https://github.com/palantir/sls-packaging), @vipshek! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

gradle-sls-packaging/src/main/resources/init.sh
```diff
@@ -25,7 +25,7 @@ is_process_service() {
   local PID=$1
   local SERVICE_NAME=$2
   # trailing '=' prevents a header line
-  ps -o command= $PID | grep -q "$SERVICE"
+  ps ewww -o command= $PID | grep -q "$SERVICE_NAME"
```
uschi2000, line 28: is ewww supported on mac and linux?

vipshek: I've tested this on Mac OS X Sierra, CentOS 6.8, and IBM AIX 7.1 and the `ewww` flag prevents truncation of results as expected.

uschi2000:
sorry for dropping the ball here; if this is still needed, then please rebase onto develop, thanks!

iamdanfox:
Closing out due to inactivity.  Also it looks like develop's `init.sh` includes the SERVICE_NAME fix and modified args to ps:

```
ps -ww -o command= $PID | grep -q "$SERVICE_NAME"
```

PR #230, opened 2017-07-17
Ensure that minimumVersion and maximumVersion are not equal

leeavital:
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/230)

This prevents a known antipattern where services declare themselves to require a lockstep upgrade. 

For example:

Service A 1.5 depends service B with the range [2.0, 2.0]

Now B's version number cannot change to anything other than 2.0 without invalidating the dependency of service A.

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependency.groovy
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependency.groovy
@@ -15,8 +15,8 @@ class ProductDependency {
 
     ProductDependency() {}
 
-    ProductDependency(String productGroup, String productName, String minimumVersion,
-                      String maximumVersion, @Nullable String recommendedVersion) {
+    ProductDependency(String productGroup, String productName, @Nullable String minimumVersion,
+                      @Nullable String maximumVersion, @Nullable String recommendedVersion) {
         this.productGroup = productGroup
         this.productName = productName
         this.minimumVersion = minimumVersion
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependency.groovy
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependency.groovy
@@ -50,5 +50,10 @@ class ProductDependency {
                         "minimumVersion and recommendedVersions must be valid SLS versions: " + it)
             }
         }
+
+        if (minimumVersion == maximumVersion) {
+            throw new IllegalArgumentException("minimumVersion and maximumVersion must be different "
+                + "in product dependency on " + this.productName)
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


uschi2000:
What's the rationale for this? (Can you add it to the commit message, too, please?)

szunami:
@uschi2000 some context here: today a service depends on an asset, and we have considered that asset to have private APIs (i.e. we haven't been careful about versioning the asset). To guarantee the correct version of the asset, the service declares a dep on the asset with min==max.

I think there are active conversations about how best to resolve this situation, but for some reason lockstep upgrades are out of the picture (I could use some more context here myself).

markelliot:
@sszuflita we're basically saying that level of coupling is not allowed.

leeavital:
Also random logistical note -- do any maintainers know what's up with circle? It's not building my PR for some reason.

gracew:
FYI, I'm planning on performing this category of checks in an internal gradle-preflight plugin. Beyond checking minimumVersion != maximumVersion, it will also check that maximumVersion >= recommendedVersion >= minimumVersion.

Not opposed to merging this here (defensive programming!), but I expect gradle-preflight to be the place where we centralize all requirements for a service to be Apollo-deployable, including producing certain roles like `sls`, `sls-status` in a product's configuration.yml.

gracew:
re: circle, I think that the 'enable builds for PRs from forks' setting is not enabled. @uschi2000 (or another admin) will have to enable it and then close/open this PR to retrigger the build.

uschi2000:
"Your build was not run - reason code (:github-failure)."

uschi2000:
PR builds are actually enabled.

uschi2000:
added you as a collaborator, lee. will have to re-open PR, sorry.

gracew:
Actually, after a second thought, I think the check I'm writing should live in this plugin, to make sure we don't generate invalid manifests to begin with. We'd need to move this repository to internal github though (something I believe we've discussed in the past), since I'm using the internal sls-versions lib.

Will proceed with having the check in gradle-preflight for now in the interest of getting something working, but I'll aim for doing the external -> internal move + housekeeping later this week.

leeavital:
Sorry for the delay. This should be ready to go. 

This is a break in that you can no longer omit both maximumVersion and minimum version.  However,  that's already dictated in the spec, and also enforced elsewhere.

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependency.groovy
```diff
@@ -15,8 +15,8 @@ class ProductDependency {
 
     ProductDependency() {}
 
-    ProductDependency(String productGroup, String productName, @Nullable String minimumVersion,
-                      @Nullable String maximumVersion, @Nullable String recommendedVersion) {
+    ProductDependency(String productGroup, String productName, String minimumVersion,
```
uschi2000: would expect something to change in the code if these are no longer optional?

leeavital: Nice catch. These are still in-fact optional.  The just can't *both* be optional at the same time because that would technically make min and max version equal. I could hack around that specific case in this PR, but I figured it wasn't worth it since min version will become mandatory in #237

leeavital:
I added the `@Nullable` checks back. This should be ready for another round.

uschi2000 (approved):

PR #233, opened 2017-07-24
Fix is_process_service

sfackler:
It happened to work before since `$2` was set to `$SERVICE`.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/233)


uschi2000 (approved):

uschi2000:
Thanks!

uschi2000:
(Just to clarify, this doesn't fix a current bug, but it fixes a future bug if SERVICE gets renamed to something else in the main scope. Correct?)

sfackler:
Yep, the behavior with and without this chance is the same.

Issue #235, opened 2017-08-02
JAVA_HOME in windows startup script?


jcvitanovic:
Shouldn't the generated `.bat` script set JAVA_HOME to the path configured in ``distribution`` ? I see from the readme that  ``javaHome`` this is only applied to `init.sh` at the moment, but this would be very useful (having another optional param, ``javaHomeWin`` e.g.)

(e.g. ``set "JAVA_HOME=service\bin\jre\...``)

PR #236, opened 2017-08-02
Configure JAVA_HOME for Windows startup script

jcvitanovic:
Add an option to configure java jre path for Windows startup script

Adresses #235 

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/236)

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.groovy
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.groovy
@@ -27,6 +27,7 @@ class JavaServiceDistributionExtension extends BaseDistributionExtension {
     private Map<String, String> env = [:]
     private boolean enableManifestClasspath = false
     private String javaHome = null
+    private String javaHomeWin = null
     private List<String> excludeFromVar = ['log', 'run']
 
     JavaServiceDistributionExtension(Project project) {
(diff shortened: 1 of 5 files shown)


palantirtech:
Thanks for your interest in [palantir/sls-packaging](https://github.com/palantir/sls-packaging), @jcvitanovic! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

uschi2000:
I think we want the java path to be a relative path (relative to where the command is run from): this is only to support bundled JREs. does that constraint work with your use case?

jcvitanovic:
This is exactly the use case I need this for. It is already relative to where the command is run from, I tested this branch  and I got my service to start up with java home set to service/lib/jre path, am I missing something?

uschi2000:
OK. If it's relative, then we should be able to use just one javaPath of the form `foo/bar/baz` and then translate it to something windows-looking behind the scenes?

iamdanfox:
@jcvitanovic looks like there's just a couple more commits to get this merged - would you be able to re-use the existing `javaHome` parameter (as it seems unlikely to need one dist with a pre-specified unix `javaHome` to also work on windows)?

jcvitanovic:
@iamdanfox  -> yea agreed.
sorry, this fell of my radar, I'll push the changes as soon as I get a chance

iamdanfox:
Closing out stale PR

PR #241, opened 2017-08-14
Upgradle to 4.1

jared2501:
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/241)


jared2501:
@uschi2000 - looks like the circle build failed since it ran out of memory.. I need this change in an upcoming PR to allow min/max version to be declared by API producers.

PR #242, opened 2017-08-14
Add ability to generate product-dependencies from published recommendations

jared2501:
Adds the ability for the CreateManifestTask to infer the min/max/recommended versions for product dependencies by reading "pdep" files that API publishers can include alongside their published jar files. The "prdep" file includes a list of product dependencies that a consumer of the jar should take.

This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/242)


jared2501:
@uschi2000 @gracew @jmcampanini @markelliot - this is part 1 of 2 of what we discussed. Part 2 will add a task to generate pdep files.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Tue Jan 24 11:44:08 CET 2017
+#Mon Aug 14 15:48:50 EDT 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.4.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-bin.zip
```
jared2501: needed to upgradle so that the GradleDependencyGenerator worked

markelliot (commented):
So I think we don't need pdep files and instead can put this info in jar manifests -- when the jars show up on the classpath, one should only need to include the jars rather than also explicitly list the group/name parts.

jared2501:
@markelliot - Hmm yeah, I discussed this option with @uschi2000 and decided against it since pdep files can work with non-java languages as well. No strong prefs either way though.

jmcampanini:
having the information in the POM doesn't limit it to jvm-languages - it just limits it to a maven layout + repository, which is almost a requirement if we're using `pdeps` that are being published (technically you could publish `pdeps` anywhere, like an asset server).

you can access the POM data from artifactory with just a link: 

```
https://artifactory.server/artifactory/my-repo-name/com/palantir/package/artifact//0.1.0/artifact-0.1.0.pom
```

you just end up with an XML file, which may be more tedious to parse in a non-jvm language (more so than JSON or just plaintext or property file)

i'd say the pros/cons for `pdeps` over `POM` would be:

* PRO: the ability to publish + host from a different server/repo, no longer requiring a maven repo
* PRO: may be slightly easier parsing of file
* PRO: provides a non-invasive path for adding info about already-published artifacts (additive vs replacing existing POM)
* CON: POM files are already the canonical source of dependency information. arguably this is a different kind of dependency, so this is less important

---

apologies, just realized no one said POM. either case, i'm in favor of pdeps

jared2501:
@jmcampanini - some other cons of POMs:

- They don't allow you to store structured information (since you can only use the properties portion of the pom for custom info afaik).
- They are relatively complex when you start to consider parent-poms.
- Gradle does not offer you a way to transitively fetch all POMs that are involved in a resolution, so the code becomes more complex.

jared2501:
Also, using POM scopes for this information is probably an anti-pattern since the behaviour if a dependency is in multiple scopes is super confusion (see https://discuss.gradle.org/t/gradle-1-9-dependency-resolution-issue-if-artifact-present-in-multiple-scopes-within-the-pom/2053/6).

build.gradle
```diff
@@ -12,6 +12,9 @@ buildscript {
         classpath 'gradle.plugin.com.palantir.configurationresolver:gradle-configuration-resolver-plugin:0.1.0'
     }
 }
+plugins {
+    id 'org.inferred.processors' version '1.1.3'
```
jmcampanini: swap to buildscript method? (to follow existing convention)

circle.yml
```diff
@@ -3,7 +3,7 @@ machine:
     version: oraclejdk8
   environment:
     TERM: dumb
-    GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryError"'
+    GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx1024m -XX:+HeapDumpOnOutOfMemoryError"'
```
jmcampanini, line 6: curious: why half the size for this PR?

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/GradleTestSpec.groovy
```diff
@@ -16,34 +16,45 @@
...
         println("Build directory: \n" + projectDir.absolutePath)
     }
 
     protected run(String... tasks) {
         return GradleRunner.create()
+                .forwardOutput()
                 .withProjectDir(projectDir)
-                .withArguments(tasks)
+                .withArguments(ImmutableList.<String> builder()
+                .addAll(Arrays.asList(tasks))
```
jmcampanini: nit: indent is funky/deceiving

uschi2000:
I think we explicitly don't want transitive resolution?

Sent from Email+ secured by MobileIron

-------- Original Message --------

markelliot:
You do, if you pick up an API jar via a library dependency, you should pick up the product dependency. 

My vote was for the literal jars and to look at the classpath, rather than manage through POMs. While this won't work for all API deps of all languages, for Java services it should be fine, and this is primarily a packager for Java services. 

We should talk through the assets case separately, as we likely want to use package.json to be more npm-centric (but you kind of get the thing I'm going for).

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
```diff
@@ -1,10 +1,159 @@
...
+    def 'Fail on missing recommended product dependencies'() {
+        setup:
+        generateDependencies()
+        buildFile << '''
+            plugins {
+                id 'com.palantir.sls-java-service-distribution'
+            }
+
+            repositories {
+                maven {url "file:///{{mavenRepo}}"}
```
jmcampanini: if you use `"""` instead of `'''` you can use string interpolation and switch this to `${mavenRepo.getAbsolutePath()}`

uschi2000:
It's likely a separate topic, but I suspect that shared libraries with extensive dependencies on service jars are an antipattern.

In today's reality (=100% of services are in Java), the JAR manifest thing could work. Not super future-proof,  but likely the solution that's gets you near-term wins most quickly.

Sent from Email+ secured by MobileIron

-------- Original Message --------

jared2501:
In the jar manifest world, what would you do about asset/asset bundles dependencies? I.e. spark module -> spark dist asset bundle dependencies?

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
```diff
@@ -1,10 +1,159 @@
...
+                runtime 'a:a:1.0'
+            }
+
+            task testCreateManifest(type: com.palantir.gradle.dist.tasks.CreateManifestTask) {
+                serviceName = "serviceName"
+                serviceGroup = "serviceGroup"
+                productType = "service"
+                manifestExtensions = [:]
+                productDependencies = []
+                productDependenciesConfig = configurations.runtime
```
jmcampanini, line 42: could we not assume this will always be runtime? at least set it by default?

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
```diff
@@ -1,10 +1,159 @@
...
+                ignoredProductIds = [
+                    new com.palantir.gradle.dist.ProductId("group:name"),
```
jmcampanini, line 81: can these be just strings and have the plugin do the conversion to `ProductId` ?

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
```diff
@@ -1,10 +1,159 @@
...
+                runtime 'a:a:1.0'
+            }
+
+            task testCreateManifest(type: com.palantir.gradle.dist.tasks.CreateManifestTask) {
+                serviceName = "serviceName"
+                serviceGroup = "serviceGroup"
+                productType = "service"
+                manifestExtensions = [:]
+                productDependencies = [
+                    new com.palantir.gradle.dist.ProductDependency("group", "name"),
```
jmcampanini, line 118: 1. could we use strings here instead as well?
2. could this use a dependency configuration: `slsProduct` or something similar? (so it would use `runtime` + `slsProduct`, or `slsProduct` extends runtime)

jmcampanini (commented):
not a full review, attempting to understand the tests first

circle.yml
jared2501: the circle builds were OOMing since free version only gives you 1GB .<

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/GradleTestSpec.groovy
jared2501: intellij auto-formats as such, will change though

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
jared2501: runtime is created by the java plugin, asset bunldes don't have the java plugin applied

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
jared2501: we could, but it appears the convention of this code base is to have the task take the typed things and do the conversion in the extension - happy to change this though

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
jared2501: see comment above

jared2501: 1. see comment above
2. think productDependencies should not move over to a configuration since you may want to explicitly list the ProductDependencies and override the suggested min/max/recommended versions

build.gradle
jared2501: done

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/GradleTestSpec.groovy
jared2501: done

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
jared2501: done

jared2501:
@jmcampanini - addressed your comments.

@uschi2000 / @markelliot - happy to change to using the jar manifest, but would like an ACK that we're OK forgoing the ability for spark module asset bundles (which do use gradle + maven) to use this feature.

uschi2000:
I have a slight preference for not using the JAR approach *if we don't need transitives*. Mark, if you think transitives are required, then I'm fine with the JAR thing too.

jared2501:
Note that the pdep file approach in this PR works just as well as using jars with transitive dependencies since we attempt to fetch a pdep for every jar in the resolved configuration.

uschi2000:
OK, misinterpreted how you resolve&fetch the pdep files.

uschi2000:
Would be interested in understanding how this will interface with authoring conjure APIs?

jared2501:
I haven't done that part yet, this part only consumes the pdep files. I imagine you'd want that part to look like this:

```
// stemma-api/build.gradle
apply plugin "sls-packaging-recommended-product-dependencies"

recommendedProductDependencies {
  recommendedProductDependency {
    productGroup = "com.palantir.stemma"
    productName = "stemma"
    minVersion = "1.0.0"
    maxVersion = "1.X.X"
    recommendedVersion = "1.2.3"
  }
}
```

jared2501:
Ping here - is this OK to move forward with?

jared2501:
Talked with @markelliot offline and decided that the asset <-> asset bundle case can probably be handled with jar manifests, and the upside of the manifest approach is that changing what people publish is harder than shoving stuff in their jars.

Will change this PR to use the jar manifests instead of pdep files.

jared2501:
@jmcampanini - made the change to jar manifests, ready for your review :)

build.gradle
```diff
@@ -5,13 +5,17 @@ buildscript {
...
         classpath 'com.gradle.publish:plugin-publish-plugin:0.9.7'
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.5'
+        classpath 'com.netflix.nebula:nebula-publishing-plugin:4.4.4'
+        classpath 'com.palantir.gradle.gitversion:gradle-git-version:0.5.1'
         classpath 'gradle.plugin.com.palantir.configurationresolver:gradle-configuration-resolver-plugin:0.1.0'
+        classpath 'gradle.plugin.org.inferred:gradle-processors:1.2.15'
     }
 }
+plugins {
+    id 'org.inferred.processors' version '1.1.3'
```
jmcampanini: would switch to an apply statement

.gitignore
```diff
@@ -32,3 +31,5 @@ fileeditor
 *.iws
 .idea/
 out/
+
+**/generated_src/
```
jmcampanini, line 35: nit: add test as well, to help future-you

gradle-sls-packaging/build.gradle
```diff
@@ -1,15 +1,18 @@
...
+    processor 'org.immutables:value:2.5.6'
+
     compile project(':sls-versions')
     compile gradleApi()
     compile 'com.fasterxml.jackson.core:jackson-databind:2.3.2'
     compile 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.3.2' // match Gradle's version
 
     testCompile gradleTestKit()
-    testCompile "com.netflix.nebula:nebula-test:4.0.0" // for better temp directory junit rule only
+    testCompile "com.netflix.nebula:nebula-test:6.1.2"
```
jmcampanini: single quotes

gradle/wrapper/gradle-wrapper.properties
jmcampanini: nit: use `-all` version of the distribution for gradle plugins, so the source code is present locally

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/GradleTestSpec.groovy
```diff
@@ -16,34 +16,45 @@
 
 package com.palantir.gradle.dist
 
-import com.energizedwork.spock.extensions.TempDirectory
 import groovy.transform.CompileStatic
 import groovy.transform.TypeCheckingMode
 import nebula.test.multiproject.MultiProjectIntegrationHelper
+import org.gradle.internal.impldep.com.google.common.collect.ImmutableList
```
jmcampanini: just-comment: in test where we control the classpath, less necessary to use a shaded version of a guava class

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
```diff
@@ -1,10 +1,156 @@
...
+import org.gradle.testkit.runner.BuildResult
+import org.gradle.testkit.runner.TaskOutcome
 
-class CreateManifestTaskTest extends Specification {
+import java.nio.file.Files
+import java.nio.file.StandardCopyOption
+
+class CreateManifestTaskTest extends GradleTestSpec {
+
+    def jsonMapper = new ObjectMapper()
```
jmcampanini: nit: can use explicit type `ObjectMapper` here

jmcampanini: can we ask the `CreateManifestTask` class to create an objectmapper to not duplicate the code?

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
```diff
@@ -1,10 +1,156 @@
...
+            }
+        """.stripIndent()
+
+        when:
+        BuildResult buildResult = run(':testCreateManifest').buildAndFail()
+
+        then:
+        buildResult.task(':testCreateManifest').outcome == TaskOutcome.FAILED
+        buildResult.output.contains("The following products are recommended as dependencies but do not appear in the " +
+                "product dependencies or product dependencies ignored list: [group:name2, group:name]")
```
jmcampanini, line 52: how does it know it needs to depend on `group:name2` and `group:name`?

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
```diff
@@ -1,10 +1,156 @@
...
+            }
+
+            task testCreateManifest(type: com.palantir.gradle.dist.tasks.CreateManifestTask) {
+                serviceName = "serviceName"
+                serviceGroup = "serviceGroup"
+                productType = "service"
+                manifestExtensions = [:]
+                productDependencies = []
+                productDependenciesConfig = configurations.runtime
+                ignoredProductIds = [
```
jmcampanini, line 80: why not use: `ignoredProductDependency(String productGroup, String productName)`?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependency.groovy
```diff
@@ -1,11 +1,24 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore
 import com.palantir.slspackaging.versions.SlsProductVersions
+import groovy.transform.EqualsAndHashCode
+import groovy.transform.ToString
 import org.gradle.api.Nullable
 
-class ProductDependency {
+@ToString
+@EqualsAndHashCode
+class ProductDependency implements Serializable {
```
jmcampanini, line 11: curious: why not switch to immutables?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductId.groovy
```diff
@@ -0,0 +1,31 @@
+package com.palantir.gradle.dist
+
+import groovy.transform.EqualsAndHashCode
+
+@EqualsAndHashCode
+class ProductId implements Serializable {
```
jmcampanini, line 6: why not use immutables?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependencies.java
```diff
@@ -0,0 +1,25 @@
...
+import org.immutables.value.Value;
+
+import java.util.Set;
+
+@Value.Immutable
+@JsonSerialize(as = ImmutableRecommendedProductDependencies.class)
+@JsonDeserialize(as = ImmutableRecommendedProductDependencies.class)
+public interface RecommendedProductDependencies {
+
+    @JsonProperty("recommended-product-dependencies")
```
jmcampanini, line 15: curious: why kebob?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.groovy
```diff
@@ -77,15 +77,18 @@ class JavaServiceDistributionPlugin implements Plugin<Project> {
             checkScript.configure(distributionExtension.serviceName, distributionExtension.checkArgs)
         }
 
+        distributionExtension.productDependenciesConfig = project.configurations.getByName("runtime")
```
jmcampanini, line 80: if done when applying plugin, then the configuration must exist, meaning the java plugin must have already added it. if there is a way to delay this or evaluate it lazily, that would be ideal. maybe move to right above line 84, in the afterEvaluate?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
```diff
@@ -76,23 +193,30 @@ class CreateManifestTask extends DefaultTask {
...
         this.serviceGroup = serviceGroup
         this.productType = productType
+        this.productDependencies = productDependencies
         this.manifestExtensions = manifestExtensions
+        this.productDependenciesConfig = productDependenciesConfig
+        this.ignoredProductIds = ignoredProductIds
+    }
+
+    static String identifierToCoord(ModuleVersionIdentifier identifier) {
+        return "$identifier.group:$identifier.name:$identifier.version"
```
jmcampanini: nit: use `${}` notation, to make the variable interpolation easier to read

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
```diff
@@ -66,6 +85,104 @@ class CreateManifestTask extends DefaultTask {
...
+        def jsonMapper = new ObjectMapper()
+                .setSerializationInclusion(JsonInclude.Include.NON_NULL)
+                .setPropertyNamingStrategy(new KebabCaseStrategy())
+
+        Set<RecommendedProductDependency> allRecommendedProductDeps = []
+        Map<String, Set<RecommendedProductDependency>> allRecommendedDepsByCoord = [:]
+        Map<String, String> mavenCoordsByProductIds = [:]
+
+        productDependenciesConfig.resolvedConfiguration.resolvedArtifacts.each { artifact ->
+            def coord = identifierToCoord(artifact.moduleVersion.id)
```
jmcampanini: nit: explicit type for `coord`

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
```diff
@@ -66,6 +85,104 @@ class CreateManifestTask extends DefaultTask {
...
+
+            if (pdeps == null) {
+                logger.debug("No pdeps found in manifest for artifact '{}', file '{}'", coord, artifact.file)
+                return
+            }
+
+            def recommendedDeps = jsonMapper.readValue(pdeps, RecommendedProductDependencies.class)
+
+            if (!allRecommendedDepsByCoord.containsKey(coord)) {
+                allRecommendedDepsByCoord.put(coord, new HashSet<RecommendedProductDependency>())
```
jmcampanini, line 119: nit: multimap may be useful here to minimize boilerplate

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
```diff
@@ -66,6 +85,104 @@ class CreateManifestTask extends DefaultTask {
...
+            }
+
+            allRecommendedProductDeps.addAll(recommendedDeps.recommendedProductDependencies())
+            allRecommendedDepsByCoord.get(coord).addAll(recommendedDeps.recommendedProductDependencies())
+
+            recommendedDeps.recommendedProductDependencies().each { recommendedDep ->
+                def productId = "${recommendedDep.productGroup()}:${recommendedDep.productName()}".toString()
+                if (mavenCoordsByProductIds.containsKey(productId)) {
+                    def othercoord = mavenCoordsByProductIds.get(productId)
+                    throw new GradleException("Duplicate product dependency recommendations found for " +
```
jmcampanini, line 129: 1. no test for this?
2. this would happen if 2 different JARs recommended the same product, correct? if so, how does that work with 2 products having a shared dependency (diamond)

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
```diff
@@ -66,6 +85,104 @@ class CreateManifestTask extends DefaultTask {
...
+                            "'${productId}' in '${coord}' and '${othercoord}'")
+                }
+                mavenCoordsByProductIds.put(productId, coord)
+            }
+        }
+
+        Map<String, RecommendedProductDependency> recommendedDepsByProductId = [:]
+        allRecommendedProductDeps.each { recommendedDep ->
+            def productId = "${recommendedDep.productGroup()}:${recommendedDep.productName()}".toString()
+            recommendedDepsByProductId.put(productId, recommendedDep)
```
jmcampanini: can this be done next to line 132?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
```diff
@@ -66,6 +85,104 @@ class CreateManifestTask extends DefaultTask {
...
+                                    productDependency.productGroup,
+                                    productDependency.productName,
+                                    recommendedProductDep.minimumVersion(),
+                                    recommendedProductDep.maximumVersion(),
+                                    recommendedProductDep.recommendedVersion()),
+                            Map))
+
+                } catch (IllegalArgumentException e) {
+                    def mavenCoordSource = mavenCoordsByProductIds.get(productId)
+                    throw new GradleException("IllegalArgumentException encountered when generating " +
```
jmcampanini, line 162: why would this happen?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
```diff
@@ -66,6 +85,104 @@ class CreateManifestTask extends DefaultTask {
...
+                            "ProductDependency from recommended dependency for ${productId} from ${mavenCoordSource}",
+                            e)
+                }
+            }
+
+            seenRecommendedProductIds.add(productId)
+        }
+
+        def unseenProductIds = new HashSet<>(recommendedDepsByProductId.keySet())
+        seenRecommendedProductIds.each { unseenProductIds.remove(it) }
```
jmcampanini, line 172: i think you can do set subtraction in groovy: https://stackoverflow.com/questions/5713665/opposite-of-intersect-in-groovy-collections

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/ServiceDistributionPluginTests.groovy
```diff
@@ -500,7 +500,7 @@ class ServiceDistributionPluginTests extends GradleTestSpec {
                 .find({ it.name.endsWith("-manifest-classpath-0.0.1.jar") })
         classpathJar.exists()
         readFromZip(classpathJar, "META-INF/MANIFEST.MF")
-                .contains('Class-Path: guava-19.0.jar produces-manifest-') // etc
+                .contains('Class-Path: guava-19.0.jar root-project-manifest-') // etc
```
jmcampanini, line 503: is the reason for this change the `settingsFile` addition to `GradleTestKit` ?

jmcampanini (commented):
* you can mix both java + groovy code in same code base. we could evolve this to using java for non-tests slowly
* would like to see some of the developer ergonomics improved (fully qualified names in consumer's build scripts) but not functionality related

build.gradle
jared2501: woops forgot to remove this, done

.gitignore
jared2501: what test would I add for this? not sure if it's worth future proofing git ignore file

gradle-sls-packaging/build.gradle
jared2501: done

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/GradleTestSpec.groovy
jared2501: woops, IJ auto import failure

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
jared2501: done

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
jared2501: The runtime configuration brings in `a:a:1.0`, the test makes it so that `a:a:1.0` depends on `b:b:1.0` as well and both `a:a:1.0` and `b:b:1.0` have stuff jammed in the MANIFEST (I checked it into the test repo)

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependency.groovy
jared2501: outside the scope of this change :)

jared2501: Also this is more groovy!

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductId.groovy
jared2501: I didn't need JSON ser/de on it

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependencies.java
jared2501: things that are in config are generally kebab case, i also coppied "manifest-extensions" - no pref here though, can change

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.groovy
jared2501: The JavaServiceDistributionPlugin applies the java plugin

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
jared2501: done (note IJ does type inference on defs)

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
jared2501: Then I'd need to pull in guava as a dependency :/ have had issue in the past when plugins have done this... your call

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
jared2501: @jmcampanini -

1. Yeah no test, can add one if you want, but just got bored of writing tests...
2. Only API jars should recommend product dependencies. Gradle will have to conflict resolve the API jars to a single jar anyway, so this should never be a problem

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
jared2501: yup, done

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
jared2501: if the RecommendedProductDependency was invalid (i.e. didn't have min/max/recommended all set)

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
jared2501: hmm yeah i wasn't quite sure if the returned set was mutable, and the line below needs to call toString before set subtraction, your call here

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
jared2501: done

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/ServiceDistributionPluginTests.groovy
jared2501: the previous version of nebula-test this repo was on didn't generate maven repo when using DependencyGraph with the version of gradle it was on. Therefore, I bumped nebula-test to latest, but the latest version of nebula-test used the crazy temporary folder annotation this projected used before. I switched to junit's TemporaryFolder, but it generates the folder as a UUID, so i had to set rootProject.name in settings.gradle so set it to somethign sane (root-project) and hence this chage...
phew

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
jmcampanini: i meant why not call the method on the extension to add things to the `Set` rather than setting the `Set` explicitly

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependencies.java
jmcampanini: figured it's worth lining up with the manifest conventions. dunno what that is atm though :)

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
jmcampanini: yup. i understand guava being a source of pain on the buildscript classpath

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
jmcampanini: ah yes, it's after the resolution. 👍

jmcampanini (approved):
lgtm. only nit is `ignoredProductDependency` is unused

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
jared2501: ah - cause this test was testing just the task separately (i.e. this test creates it's own task "testCreateManifest" instead of using the conventional "createManifest")... I could change all the tests do use "createManifest" and then invoke the task and use "-x downStreamTask1 -x downStreamTask2, etc", but decided against cause im lazy >.<

jared2501:
@jmcampanini when I have part 2 of this change (the task that jams the recommended product dependencies in the jar manifest) I can write an end to end test that uses the extension if you fancy? :D

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
jmcampanini: cool.

assuming there only 1 manifest per project, i'd say that providing an extension to override is simpler/nicer than customizing a task. but that's ergonomic related, not functionality related, so can be a separate PR if desired

jmcampanini:
👍  lgtm. would def like an ete test that uses both parts. and i am fancy ;)

jared2501:
Awesome! @jmcampanini - do you think you can approve + merge this? Will get a part 2 up shortly.

jmcampanini (approved):

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.groovy
```diff
@@ -83,6 +94,14 @@ class BaseDistributionExtension {
         productDependencies.add(dependency)
     }
 
+    void setProductDependenciesConfig(Configuration productDependenciesConfig) {
+        this.productDependenciesConfig = productDependenciesConfig
+    }
+
+    void ignoredProductDependency(String productGroup, String productName) {
```
iamdanfox, line 101: @jared2501 could you mention this in your docs PR? I was pretty surprised a 2.4 -> 2.7 upgrade broke my build due to a pdep from MP!

PR #248, opened 2017-09-08
no longer match with full path, since mac os uses different path strings

jmcampanini:
when on a mac, we're searching for:

```
distTar: /var/folders/1n/lylm6fr54vn1mbhc_7rlgx95080n7q/T/junit3615108091981279385/junit2160697864594598480/build/distributions/my-service.sls.tgz
```

and the output contains:
```
distTar: /private/var/folders/1n/lylm6fr54vn1mbhc_7rlgx95080n7q/T/junit3615108091981279385/junit2160697864594598480/build/distributions/my-service.sls.tgz
```

so the leading `/private` is causing the `distTar: /var/...` string to not match. this change now looks for the trailing part of the folder AND a suffix of ` - distTar` this should be sufficiently equal for the project folder and avoid the private prefix issue

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/248)

### DIFF
--- a/gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/ServiceDistributionPluginTests.groovy
+++ b/gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/ServiceDistributionPluginTests.groovy
@@ -533,15 +533,19 @@ class ServiceDistributionPluginTests extends GradleTestSpec {
             }
 
             afterEvaluate {
-                println "from test: ${distTar.outputs.files.singleFile} - distTar"
+                String actualTarballPath = distTar.outputs.files.singleFile.absolutePath
+                String expectedTarballPath = project.file('build/distributions/my-service.sls.tgz').absolutePath
+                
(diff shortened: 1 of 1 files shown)


markelliot:
pretty sure this is a correct test that's indicative of a bug

jmcampanini:
my understanding based on the test name and running `:task` is that it's testing the artifact is named based on the configuration. it's still odd that the paths don't match up, even when i add `.absolutePath` to them... going to dig more.

assuming the test is to determine the artifact name, then this PR should still leave the test in a good state

jmcampanini:
seems that both exist on mac. when printed from inside an integration test (being ran by gradle test kit), it shows up with `/private`. and from the test code, `projectDir` which is generated by `TemporaryFolder` returns it without the `/private` prefix.

i could also rewrite the test build script to fail the build if they don't match, so the file references will be coming from the same environment

markelliot:
That’d work.

uschi2000 (approved):

PR #250, opened 2017-09-12
Force unlimited width in ps output

ctshepherd:
`ps -o command` has an undefined maximum output line length if -w isn't specified as an argument (see https://github.com/mmalecki/procps/blob/fe4c4a7314f32907b9f558ad0d8b8d0ff1cc76be/ps/ps.1#L868). This means that although the process is correctly started `init.sh status` will incorrectly fail to find the process and conclude it failed to start correctly because the service name was truncated. Specifying -ww forces unlimited output line width which fixes this problem. Seen in the wild with an older version of ps on Centos 6.6.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/250)

### DIFF
--- a/gradle-sls-packaging/src/main/resources/init.sh
+++ b/gradle-sls-packaging/src/main/resources/init.sh
@@ -25,7 +25,7 @@ is_process_service() {
   local PID=$1
   local SERVICE_NAME=$2
   # trailing '=' prevents a header line
-  ps -o command= $PID | grep -q "$SERVICE_NAME"
+  ps -ww -o command= $PID | grep -q "$SERVICE_NAME"
   return $?
 }
 


palantirtech:
Thanks for your interest in [palantir/sls-packaging](https://github.com/palantir/sls-packaging), @ctshepherd! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

uschi2000:
What's the behavior on MacOS?

ctshepherd:
Looks like it already does the right thing:

`     -w      Use 132 columns to display information, instead of the default which is your window size.  If the -w option is specified more than once, ps will use as many columns as
             necessary without regard for your window size.  When output is not to a terminal, an unlimited number of columns are always used.`

uschi2000 (approved):

PR #256, opened 2017-10-20
Change sleep value to 5 seconds in init.sh

leeavital:
If a server has a particularly long startup sequence (or the JVM just takes a while to start), this script will spuriously report sucess for `init.sh start`. Our infrastructure will then asyncronously call this later and realize that the service is in fact down. This results in the service flipping on and off instead of failing fast. Bumping this value to 5 seconds does not solve the problem, but  does give services a longer window in which to fail on startup.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/256)


jared2501 (approved):

jonsyu1:
LGTM

uschi2000 (approved):

PR #263, opened 2018-01-26
Add task to create configuration tarballs

andybradshaw:
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/263)

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/ConfigTarTask.groovy
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/ConfigTarTask.groovy
@@ -22,6 +22,8 @@ import org.gradle.api.tasks.bundling.Compression
 import org.gradle.api.tasks.bundling.Tar
 
 class ConfigTarTask {
+    private static final String PRODUCT_TYPE_REGEX = "^\\w+?\\.v\\d+\$"
+
     static Tar createConfigTarTask(Project project, String taskName, String productType) {
         project.tasks.create(taskName, Tar) { p ->
             p.group = JavaServiceDistributionPlugin.GROUP_NAME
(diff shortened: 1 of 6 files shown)


iamdanfox:
@andybradshaw code looks nice and clean but what's the use case for this `configTar` task?  Does it need a README addition too?

bmoylan:
Yeah we should add relevant docs to the readme. @iamdanfox there's an internal spec PR that explains what this is for.

@andybradshaw do we need to wire up any publishing or do consumers do that themselves in their own gradle buildscripts?

andybradshaw:
I thought leaving the publishing up to consumers was fine. Will update the README.

andybradshaw:
I added the task and a little blurb about it to the readme. I'm not sure if it's clear what I mean by `deployment configuration`. Could probably elaborate a little bit more on that.

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/ConfigTarTask.groovy
```diff
@@ -0,0 +1,55 @@
...
+
+class ConfigTarTask {
+    static Tar createConfigTarTask(Project project, String taskName, String productType) {
+        project.tasks.create(taskName, Tar) { p ->
+            p.group = JavaServiceDistributionPlugin.GROUP_NAME
+            p.description = "Creates a compressed, gzipped tar file that contains the sls configuration files for the product"
+            // Set compression in constructor so that task output has the right name from the start.
+            p.compression = Compression.GZIP
+            // The extension is the product type without the version
+            // service.v1 -> .service.config.tgz
```
uschi2000, line 34: should assert that the type has this format and fail if it doesn't

andybradshaw: I'm fine adding a check for this, but it looks like the distribution extension already enforces that it comes from a predefined list of types: 
```java
void productType(String type) {
    if (!VALID_PRODUCT_TYPES.contains(type)) {
        throw new IllegalArgumentException("Invalid product type '${type}' specified; supported types: ${VALID_PRODUCT_TYPES}.")
    }
    this.productType = type
}
```

uschi2000: kind of brittle, but ok

andybradshaw:
Spent a little bit of time looking into why this is flaking so much. It flakes on my machine with some IO errors for `:testCreateManifest`, but with lower frequency.

iamdanfox (approved):

PR #264, opened 2018-01-31
Disable optimization which can lead to empty stacktraces

a-k-g:
We recently had this issue where a `ClassCastException` was occuring repeatedly without a stacktrace. Turns out that it was because of this JVM optimization. I don't think we gain much value and given the pain it can cause in debugging, I think we should disable it. This optimization seems useful if you're using exceptions as control flow, but that is anyway an anti-pattern. 

@rfink, @schlosna thoughts?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/sls-packaging/264)

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
@@ -43,6 +43,10 @@ class LaunchConfigTask extends DefaultTask {
             '-verbose:gc'
     ]
 
+    static final List<String> optimizationsToDisableJvmOpts = [
+            '-XX:-OmitStackTraceInFastThrow'
+    ]
+
     static final List<String> dirs = ['var/data/tmp']
 
     @Input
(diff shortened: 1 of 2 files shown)


schlosna:
If we’re seeing stacktraceless exceptions, that typically means that there is a bug on a very hot code path. I’m a bit hesitant to apply this across all services using this plugin, especially without having any data on performance impact of different services’ workloads. 

I’m also a bit worried that something is using exceptions for control flow which is both a mess from code maintainability, but also bad for performance (see e.g. https://github.com/jersey/jersey/pull/3577 and https://github.com/google/guava/commit/e6003692500352e1ba59eee5c93bc21acb3f46c0 ). Unfortunately it is hard to avoid completely as the JDK has a number of places where exceptions are explicitly used for control flow, especially ClassCastException which are thrown heavily during classloading to fall back on other classloaders. If you’re seeing a CCE bubble out, that indicates a bug that needs to be fixed, and that service specifically could disable the stacktrace generation suppression optimization (say that five times fast).

schlosna:
Overall I'd prefer we don't set `-XX:-OmitStackTraceInFastThrow` by default for all services.

See also https://bugs.openjdk.java.net/browse/JDK-8046503 and https://bugs.openjdk.java.net/browse/JDK-8042127 for some notes on the potential impacts of disabling OmitStackTraceInFastThrow.

a-k-g:
Thanks for the comments! The perf concerns make sense as various libraries/the JDK itself could be exceptions for control flow.

I guess in my case it was just hard to debug because I wasn't aware why there wasn't any stacktrace. Once I discovered that it was due to a JVM optimization it was straightforward.

PR #269, opened 2018-03-13
serviceName is actually required

tpetracca:
When running `distTar` on a project with the `sls-java-service-distribution`
```
FAILURE: Build failed with an exception.

* What went wrong:
A problem was found with the configuration of task ':proejct:createCheckScript'.
> No value has been specified for property 'serviceName'.
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -92,8 +92,7 @@ A sample configuration for the Service plugin:
 
 And the complete list of configurable properties:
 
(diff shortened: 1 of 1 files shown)


gracew:
what version are you using? It shouldn't be required after https://github.com/palantir/sls-packaging/pull/228, which is in sls-packaging 2.5.0+.

tpetracca:
womp. I'm on 2.4.0.

PR #273, opened 2018-03-21
set sun.net.inetaddr.ttl to 20s by default

jnelson15:
Tested RDS failover on an internal product with with 60s as per email with Dan and team and it took the product about 70s to re-establish valid connections (likely due to DNS timeout + connection pool timeouts), which I think may be a little too long given that these can happen somewhat frequently -- e.g. AWS maintenance. Setting to 20s here but open to recommendations if people have opinions on a more appropriate value.

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
@@ -40,7 +40,8 @@ class LaunchConfigTask extends DefaultTask {
             '-XX:GCLogFileSize=10M',
             '-XX:NumberOfGCLogFiles=10',
             '-Xloggc:var/log/gc-%t-%p.log',
-            '-verbose:gc'
+            '-verbose:gc',
+            '-Dsun.net.inetaddr.ttl=20'
     ]
 
     static final List<String> dirs = ['var/data/tmp']
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
@@ -40,7 +40,10 @@ class LaunchConfigTask extends DefaultTask {
             '-XX:GCLogFileSize=10M',
             '-XX:NumberOfGCLogFiles=10',
             '-Xloggc:var/log/gc-%t-%p.log',
-            '-verbose:gc',
+            '-verbose:gc'
+    ]
+
+    static final List<String> dnsJvmOpts = [
             '-Dsun.net.inetaddr.ttl=20'
     ]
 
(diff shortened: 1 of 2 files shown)


jnelson15:
\+ @iamdanfox @schlosna @aggie05engineer

uschi2000:
What's the relationship with `networkaddress.cache.ttl` and why do you suggest we modify `sun.net.inetaddr.ttl`?

jnelson15:
@uschi2000 according to https://stackoverflow.com/questions/1256556/any-way-to-make-java-honor-the-dns-caching-timeout-ttl, the former is a security property vs latter is system property. The `networkaddress.cache.*` ones can be modified in the `java.security` file or in code. As flagged by @aggie05engineer, we can't really use the system property for the *negative* TTL since our `java.security` files explicitly set a `networkaddress.cache.negative.ttl` which is given priority over the system property. However, a value is not set for `networkaddress.cache.ttl` so we're safe to do pass in the `sun.net.inetaddr.ttl` value.

jnelson15:
@iamdanfox this change make sense to you?

iamdanfox (approved):
Yeah 20s sounds sensible to me - it's also reassuring that we can still override this on a per-product basis using the `java.security.properties` approach if necessary!

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
```diff
@@ -40,7 +40,8 @@ class LaunchConfigTask extends DefaultTask {
             '-XX:GCLogFileSize=10M',
             '-XX:NumberOfGCLogFiles=10',
             '-Xloggc:var/log/gc-%t-%p.log',
-            '-verbose:gc'
+            '-verbose:gc',
+            '-Dsun.net.inetaddr.ttl=20'
```
iamdanfox: Actually could we put this in a different place than `gcJvmOpts`?

iamdanfox (changes requested):

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
jnelson15: Yeah is there another list of options or just create a new variable?

iamdanfox: I think a new variable like `dnsJvmOpts` would be OK, they all get squashed together further down:

```
         writeConfig(createConfig(getArgs(), tmpdirJvmOpts + gcJvmOpts + defaultJvmOpts), getStaticLauncher())
```

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
```diff
@@ -43,6 +43,10 @@ class LaunchConfigTask extends DefaultTask {
             '-verbose:gc'
     ]
 
+    static final List<String> dnsJvmOpts = [
+            '-Dsun.net.inetaddr.ttl=20'
```
iamdanfox, line 47: Are we certain this is in seconds and not milliseconds?

jnelson15: Yeah: https://docs.oracle.com/javase/8/docs/technotes/guides/net/properties.html

iamdanfox (approved):

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
uschi2000: would be good to clarify this in a comment

jnelson15: https://github.com/palantir/sls-packaging/pull/274

PR #275, opened 2018-03-28
package libyjpagent.so in dist tars

jnelson15:
Copied logic for check/init scripts for providing the x86 yourkit agent in distributions. Are we okay with checking in the file into our resources in this repo? YourKit only seems to publish their big zips (88mb, including all the client code) in https://www.yourkit.com/java/profiler/ while we just need the ~3mb .so file, and not a good versioning story as well.

Next step following this PR will be adding config in the java-go-launcher to set the appropriate java args, potentially doing architecture check as well if that's feasible.

\+ @iamdanfox @vinooganesh @tstearns


jnelson15:
Tested on internal product:
```
jeffn-mbp:distributions jeffn$ find <service-name>/service/lib -type f | grep -v jar
<service-name>/service/lib/linux-x86-64/libyjpagent.so
```

jnelson15:
Actually nevermind, seems to be copying the file incorrectly, will dig in a bit more

jnelson15:
Okay this ^ one works. Just confirmed can attach Yourkit client to service built with this branch of sls-packaging.

iamdanfox:
@jnelson15 if we're checking in the .so file, I'd love to have a bash script so that we can easily upgrade it later (e.g. curl, untar, delete most files)!

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/CopyYourkitAgentTask.groovy
```diff
@@ -0,0 +1,30 @@
...
+
+import com.palantir.gradle.dist.service.JavaServiceDistributionPlugin
+import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.TaskAction
+
+import java.nio.file.Files
+import java.nio.file.Path
+
+class CopyYourkitAgentTask extends DefaultTask {
```
iamdanfox, line 12: This looks great so far! it would be great to have a unit test that verifies this file gets written (and is approximately the right size).

Hopefully should be able to use a similar structure to `ConfigTarTaskTest`. It uses spock rather than junit, so it all looks a bit alien, but you should be able to run your task from the unit test.

One key thing to assert is the incremental behaviour - we want to make sure this task understands it's up-to-date if the file already exists:
```
buildResult.task(':copyYourkitAgent').outcome == TaskOutcome.UP_TO_DATE
```

jnelson15:
Okay will add. Any thoughts on how we can version it? Imagine we'll want to take occasional upgrades as users get newer clients. Maybe just in a readme?

iamdanfox:
Actually I think we really want to make sure this `.so` file is very clearly auditable, because it'll end up in every single running palantir service.  I bet a cheeky little opaque binary file would actually make AppSec folks pretty uncomfortable.

Maybe we should download it within the sls-packaging CircleCI build, verify it against some known checksum, then add the .so file as a resource?

jnelson15:
How do we do that? Would we have to download the whole zip and extract every time that way?

iamdanfox:
Yeah I think so, although I hope it will be pretty quick on CI.  I'm sure we've downloaded 88mb of jar dependencies elsewhere :) this one is really just a zip dependency we happen to adjust.

jnelson15:
Cool do you have an example of something like that I can base off of?

iamdanfox:
OK so I think there's a gradle plugin that we can add to sls-packaging's top level build.gradle that can actually help us here: https://github.com/michel-kraemer/gradle-download-task#verify-task

```gradle
task downloadFile(type: Download) {
    src 'http://www.example.com/index.html'
    dest buildDir
}

task verifyFile(type: Verify, dependsOn: downloadFile) {
    src new File(buildDir, 'index.html')
    algorithm 'SHA-256'
    checksum '<some-checksum>'
}
```

We can figure out the actual expected checksum by running `sha256sum`. We'd just need to add another gradle exec task on the end to actually unzip and pull the right .so file out!

jnelson15:
Okay will see if I can get that to work, thanks

markelliot:
The download really should use Gradle deps so that it’s cached — see how gradle-baseline does that as an example.

________________________________

jnelson15:
Talked to @iamdanfox and he thinks the the gradle-download-task approach makes more sense here for sake of simplicity and so we can checksum the file but it should work similarly in either case if people have strong preferences otherwise.

@qinfchen for SA

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/CopyYourkitAgentTask.groovy
jnelson15: Still have to do this. Will follow-up with some tests if the general approach makes sense now.

gradle-sls-packaging/build.gradle
```diff
@@ -37,10 +38,30 @@ pluginBundle {
     }
 }
 
+task downloadYourkitDist(type: Download) {
```
iamdanfox: might need to fully qualify `Download` here...

jnelson15: Seems to work as-is with the plugin applied above

gradle-sls-packaging/build.gradle
```diff
@@ -37,10 +38,30 @@ pluginBundle {
...
+    src 'https://www.yourkit.com/download/YourKit-JavaProfiler-2017.02-b75.zip'
+    dest buildDir
+}
+
+task verifyYourkitDist(type: Verify, dependsOn: downloadYourkitDist) {
+    src new File(buildDir, 'YourKit-JavaProfiler-2017.02-b75.zip')
+    algorithm 'MD5'
```
iamdanfox: Could we use SHA-256 just for paranoia's sake?

jnelson15: Yep sure

iamdanfox:
This approach looks gr8 to me.  Let's pop in a single smoke test to verify that the .so resource is not empty and then ship it

circle.yml
```diff
@@ -29,5 +30,6 @@ deployment:
     owner: palantir
     commands:
       - ./gradlew -i bintrayUpload
+      - ./gradlew -i :gradle-sls-packaging:extractYourkitAgent
```
iamdanfox: rather than explicitly adding these here, could we just add a few lines to one of our build.gradles so that the task is run automatically?

```
tasks.idea.dependsOn extractYourkitAgent
tasks.compileJava.dependsOn extractYourkitAgent
```

iamdanfox (dismissed):

jnelson15:
Let's hold off on merging until getting +1 from @markelliot and/or @schlosna

iamdanfox:
@jnelson15 can you also confirm whether we got official approval to bundle this `.so` file?  My own cursory read of https://www.yourkit.com/purchase/eula.pdf suggests it's probably OK as all the license types allow:

>  enable profiling in any number of Java applications simultaneously, running them with the profiler agent
(yjpagent.dll or libyjpagent.so or libyjpagent.jnilib);

(Obviously would prefer to get someone with actual legal knowledge to sign this off!)

jnelson15:
Yeah we got approval to distribute this specific file (the licenses are for the clients, not the agent -- legal confirmed with YourKit support), but will still want to confirm with @markelliot @schlosna or @briandorne  about the other security things I mentioned in the email. Ping me if you want more info on those issues.

jnelson15:
@iamdanfox we have a +1 from Aki (see email) so I think should be good to go here if it looks good to you. Just added the license in the distributions as well.

iamdanfox (approved):

iamdanfox:
Thanks for pushing this through @jnelson15!

PR #280, opened 2018-05-31
Append to startup log rather than overwrite

ellisjoe:


iamdanfox:
@ellisjoe for long-lived installations, won't this lead to the startup log growing infinitely?  Nothing is set up to compress/truncate the startup log...

ellisjoe:
yeah I don't actually know if this is the right solution. the issue was that jvms were OOMing which dumps some diagnostic logs to stdout, then our deployment infra was restarting the service automatically and overwriting the startup log. open to other ideas here

markelliot:
we'd need to pipe to a real log rotator -- the startup stuff here could use another pass

dansanduleac:
gone stale

fryz:
Filed https://github.com/palantir/sls-packaging/issues/286 to keep track of this.

PR #282, opened 2018-06-12
Update copyrights

dansanduleac:


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/asset/AssetDistributionExtension.groovy
```diff
@@ -1,11 +1,11 @@
 /*
- * Copyright 2016 Palantir Technologies
+ * Copyright 2018 Palantir Technologies, Inc. All rights reserved.
```
markelliot: where we have an existing year I think we want to retain it.

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.groovy
```diff
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2018 Palantir Technologies, Inc. All rights reserved.
```
markelliot: I think we're missing the `(c)`

dansanduleac:
@markelliot updated - I'll add baseline in another PR to enforce the format

iamdanfox (approved):

PR #284, opened 2018-06-13
Set env MALLOC_ARENA_MAX=4 by default

carterkozak:
Given some versions of glibc, this can provide a significant
reduction in virtual memory usage.

https://jkutner.github.io/2017/04/28/oh-the-places-your-java-memory-goes.html
https://github.com/prestodb/presto/issues/8993
https://issues.apache.org/jira/browse/HADOOP-7154

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
@@ -51,6 +51,11 @@ class LaunchConfigTask extends DefaultTask {
 
     static final List<String> dirs = ['var/data/tmp']
 
+    // Reduce memory usage for some versions of glibc.
+    // Default value is 8 * CORES.
+    // See https://issues.apache.org/jira/browse/HADOOP-7154
+    static final Map<String, String> defaultEnvironment = Collections.unmodifiableMap(['MALLOC_ARENA_MAX':'4'])
+
     @Input
     String mainClass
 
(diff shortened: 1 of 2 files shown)


carterkozak:
@schlosna fysa

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
```diff
@@ -51,6 +51,8 @@ class LaunchConfigTask extends DefaultTask {
 
     static final List<String> dirs = ['var/data/tmp']
 
+    static final Map<String, String> defaultEnvironment = Collections.unmodifiableMap(['MALLOC_ARENA_MAX':'4'])
```
schlosna: Might be worth including a comment referencing https://issues.apache.org/jira/browse/HADOOP-7154 for some of the background on why this is being set

schlosna (commented):
Overall LGTM, minor nit to add comment for context

schlosna (approved):

schlosna:
@uschi2000 and @markelliot would like to get your eyes on this and a release if possible. Thanks!

See also http://www.evanjones.ca/java-native-leak-bug.html

uschi2000:
I have no experience or knowledge of the internals here, so defer to you.

uschi2000 (approved):

schlosna:
Thanks @uschi2000 , can you please merge and tag a release as I don't have permissions on this repo.

markelliot:
Is it possible to verify from a running process that this setting gets respected? The tail of the linked Hadoop issue seems to indicate not all versions of glibc respect the var.

schlosna:
@markelliot yes we could find a way to determine glibc version on the system, but not sure it is worth it at this point as the issue hits glibc shipped with CentOS 6+.

PR #285, opened 2018-06-13
Add baseline 0.21.0, use common publish-jar

dansanduleac:
Depends on #282 otherwise it won't pass checkstyle


iamdanfox (approved):
Looks good!

Issue #279, opened 2018-05-21
Allow consumers to resolve conflicts between recommended dependencies for manifests


gregakinman:
Situation:

- Project A directly depends on library B.
- Library B directly depends on version 1 of library C and declares this as a recommended product dependency in its JAR manifest.
- Library B directly depends on library D, which has declared version 2 of library C as a recommended product dependency in its JAR manifest.
- As a result, there are now two versions of library C recommended to project A.

sls-packaging should provide a way to allow the consumer to reconcile these if they are different - currently it just lets them slide if they're the same (https://github.com/palantir/sls-packaging/pull/244).

dansanduleac:
You can short circuit the conflict detection logic now by
1) replacing the distribution extension's `productDependenciesConfig` with an empty/bogus one:
    ```
    distribution {
        productDependenciesConfig = configurations.detachedConfiguration()
    }
    ```
2) Declaring the min/max/recommended versions of your own product dependencies manually, [like so](https://github.palantir.build/foundry/foundry/blob/126da1f3202d09940ed0bd613900ea670d97ee36/foundry-fsclient/build.gradle#L90-L96), but unlike in the example, nest these `productDependency` blocks under your `distribution` block.

However, #293 should address the issues where the dependency ranges differ but are otherwise non-disjoint.

gregakinman:
Nice, thanks for the PR @dansanduleac. I'd consider this fixed.

PR #293, opened 2018-06-14
Merge RecommendedProductDependency

dansanduleac:
Currently, [`CreateManifestTask`](https://github.com/palantir/sls-packaging/blob/develop/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy) will fail hard (with no recourse) if multiple `RecommendedProductDependency` come from two transitive dependencies and they are not exactly equal (i.e. they have the same minimum, maximum and recommended versions).
This is bad, and causes people to think about workarounds such as #279.
That change might be useful in itself as well (as a last recourse, to get people unblocked when transitive deps have declared unnecessarily strict version ranges in their dependencies).

However, it would be better if we make an attempt to reconcile two conflicting `RecommendedProductDependency` by constructing the most accommodating such dependency entry that respects the min/max versions of the conflicted dependency entries.

Simplified example:
```
{min=1.0.0, max=1.6.x}
{min=1.3.5, max=1.x.x}
-->
{min=1.3.5, max=1.6.x}
```


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependencyMerger.java
```diff
@@ -0,0 +1,96 @@
...
+                .of(
+                        OrderableSlsVersion.valueOf(dep1.getRecommendedVersion()),
+                        OrderableSlsVersion.valueOf(dep2.getRecommendedVersion()))
+                .filter(version -> comparator.compare(version, minimumVersion) >= 0
+                        && maximumVersion.map(maxVer -> maxVer.matches(version)).orElse(true))
+                .max(comparator)
+                .orElse(minimumVersion);
+
+        RecommendedProductDependency result = new RecommendedProductDependency();
+        result.setMetaClass(dep1.getMetaClass());
```
dansanduleac: not sure if I'm doing the right thing here, this is a weird groovy class...

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependencyMerger.java
```diff
@@ -0,0 +1,99 @@
...
+                Arrays.asList(
+                        OrderableSlsVersion.valueOf(dep1.getMinimumVersion()),
+                        OrderableSlsVersion.valueOf(dep2.getMinimumVersion())),
+                comparator);
+
+        Optional<SlsVersionMatcher> maximumVersion = Stream
+                .of(
+                        SlsVersionMatcher.valueOf(dep1.getMaximumVersion()),
+                        SlsVersionMatcher.valueOf(dep2.getMaximumVersion()))
+                .filter(Objects::nonNull)
```
iamdanfox: nonNull earlier?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.groovy
```diff
@@ -20,13 +20,12 @@ import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.core.type.TypeReference
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.databind.PropertyNamingStrategy
-import com.palantir.gradle.dist.ProductDependency
-import com.palantir.gradle.dist.ProductId
-import com.palantir.gradle.dist.RecommendedProductDependencies
-import com.palantir.gradle.dist.RecommendedProductDependency
+import com.palantir.gradle.dist.*
```
iamdanfox: pls no

dansanduleac: yep sorry - it was intellij :)

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/MaximumVersion.java
```diff
@@ -0,0 +1,72 @@
...
+import com.palantir.sls.versions.OrderableSlsVersion;
+import com.palantir.sls.versions.SlsVersionMatcher;
+import com.palantir.sls.versions.VersionComparator;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+
+abstract class MaximumVersion implements Comparable<MaximumVersion> {
```
ferozco: why do we need a class hierarchy?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/MaximumVersion.java
```diff
@@ -0,0 +1,72 @@
...
+    public abstract <T> T fold(
```
ferozco: please leave a comment

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/MaximumVersion.java
dansanduleac: explained in docstring

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/MaximumVersion.java
```diff
@@ -0,0 +1,91 @@
...
+ * {@link OrderableSlsVersion}, doesn't handle all the cases that the latter supports. For instance, it can't represent
+ * non-release versions (see version types enum {@link SlsVersionType}).
+ */
+abstract class MaximumVersion implements Comparable<MaximumVersion> {
+    /**
+     * Functional abstraction for the visitor pattern.
+     * Handles both possible states of this MaximumVersion (where it is represented by an {@link OrderableSlsVersion}
+     * or an {@link SlsVersionMatcher}), without requiring the caller to know in advance which one it is.
+     */
+    abstract <T> T fold(
```
iamdanfox: I know this is a cool way to achieve pattern matching-style stuff, but it's pretty confusing to read.  Please could you just use the normal MaximumVersionVisitor thing instead so that it's clear there's no magic here?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/MaximumVersion.java
```diff
@@ -0,0 +1,91 @@
...
+        return Stream
+                .<Supplier<Optional<MaximumVersion>>>of(
+                        () -> OrderableSlsVersion.safeValueOf(version).map(VersionMaximumVersion::new),
+                        () -> SlsVersionMatcher.safeValueOf(version).map(MatcherMaximumVersion::new))
+                .map(Supplier::get)
+                .flatMap(MaximumVersion::optionalToStream)
+                .findFirst()
+                .orElseThrow(() -> new SafeIllegalArgumentException(
+                        "Couldn't parse version as an OrderableSlsVersion or an SlsVersionMatcher",
+                        UnsafeArg.of("version", version)));
```
iamdanfox: I think the super functional approach here actually obscures readability... could we do this imperatively?

```java
    static MaximumVersion valueOf(String version) {
        Optional<OrderableSlsVersion> maybeOrderable = OrderableSlsVersion.safeValueOf(version);
        if (maybeOrderable.isPresent()) {
            return new VersionMaximumVersion(maybeOrderable.get());
        }

        Optional<SlsVersionMatcher> maybeMatcher = SlsVersionMatcher.safeValueOf(version);
        if (maybeMatcher.isPresent()) {
            return new MatcherMaximumVersion(maybeMatcher.get());
        }

        throw new SafeIllegalArgumentException(
                "Couldn't parse version as an OrderableSlsVersion or an SlsVersionMatcher", 
                UnsafeArg.of("version", version));
    }
```

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/MatcherMaximumVersion.java
```diff
@@ -0,0 +1,39 @@
...
+package com.palantir.gradle.dist;
+
+import com.palantir.sls.versions.OrderableSlsVersion;
+import com.palantir.sls.versions.SlsVersionMatcher;
+import java.util.function.Function;
+
+/**
+ * {@link MaximumVersion} implemented in terms of a {@link SlsVersionMatcher}.
+ */
+final class MatcherMaximumVersion extends MaximumVersion {
```
iamdanfox: Possibly worth making these two variants inner classes of MaximumVersion?  I think the javadoc should actually just give an example rather than repeating classnames, e.g.:

> Matchers may contain wildcards, e.g. 1.X.X

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/MaximumVersionComparator.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Comparator;
+import java.util.OptionalInt;
+
+enum MaximumVersionComparator implements Comparator<MaximumVersion> {
+    INSTANCE;
+
+    static final Comparator<OptionalInt> EMPTY_IS_GREATER =
+            Comparator.comparingInt(num -> num.isPresent() ? num.getAsInt() : Integer.MAX_VALUE);
+
+    static final Comparator<SlsVersionMatcher> MATCHER_COMPARATOR = Comparator
```
iamdanfox: This seems like a really nice pure function for dealing with version ranges.  Can we just put it in sls-versions now that it's open source?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependency.groovy
```diff
@@ -59,7 +61,7 @@ class RecommendedProductDependency {
             }
         }
         [minimumVersion, recommendedVersion].each {
-            if (it && !SlsProductVersions.isValidVersion(it)) {
+            if (it && !SlsProductVersions.isOrderableVersion(it)) {
```
iamdanfox, line 64: 🤦‍♂️ nice

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependencyMerger.java
```diff
@@ -0,0 +1,96 @@
...
+    private RecommendedProductDependencyMerger() { }
+
+    public static RecommendedProductDependency mergeRecommendedProductDependencies(
+            RecommendedProductDependency dep1, RecommendedProductDependency dep2) {
+        // Ensure they are valid
+        Arrays.asList(dep1, dep2).forEach(RecommendedProductDependency::isValid);
+        checkArgument(
+                dep1.getProductGroup().equals(dep2.getProductGroup()),
+                "Product groups differ",
+                SafeArg.of("dep1ProductGroup", dep1.getProductGroup()),
```
iamdanfox: I'm not sure how useful the safe-logging preconditions here are, as we're just going to use gradle's SLF4J implementation... I think this means we need `{}` tokens in the message (or maybe `%s` I can never remember)

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependencyMerger.java
```diff
@@ -0,0 +1,96 @@
...
+        VersionComparator versionComparator = VersionComparator.INSTANCE;
+        OrderableSlsVersion minimumVersion = Collections.max(
+                Arrays.asList(
+                        OrderableSlsVersion.valueOf(dep1.getMinimumVersion()),
+                        OrderableSlsVersion.valueOf(dep2.getMinimumVersion())),
+                versionComparator);
+
+        MaximumVersionComparator maximumVersionComparator = MaximumVersionComparator.INSTANCE;
+        Optional<MaximumVersion> maximumVersion = Stream
+                .of(dep1.getMaximumVersion(), dep2.getMaximumVersion())
```
iamdanfox: kinda funky to use Stream.of here but Collections.max(Arrays.asList()) above... pick one and run with it?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependencyMerger.java
```diff
@@ -0,0 +1,96 @@
...
+                .filter(Objects::nonNull)
+                .map(MaximumVersion::valueOf)
+                .min(maximumVersionComparator);
```
iamdanfox: mega +1 to encapsulating all this logic in nice clean comparators.  Possibly just inline the variable as the class is already nicely named?

versions.props
```diff
@@ -1,4 +1,5 @@
 com.fasterxml.jackson.*:* = 2.8.11
```
iamdanfox, line 1: btw is the `b-duplicate-different-versions-1.0.jar` necessary?

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/RecommendedProductDependencyMergerTest.groovy
```diff
@@ -0,0 +1,88 @@
...
+
+        then:
+        merged.minimumVersion == "2.1.0"
+        merged.maximumVersion == "2.8.x"
+        merged.recommendedVersion == "2.6.0"
+    }
+
+    private RecommendedProductDependency newRecommendation(String min, String max, String recommended) {
+        def rpd = basicDep.clone() as RecommendedProductDependency
+        use InvokerHelper, {
```
iamdanfox, line 83: woah. Is this really how we construct RecommendedProductDependencies elsewhere??

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/MaximumVersion.java
dansanduleac: I don't think this is magic, it's pretty clear what this does coming from a functional programming background. (if the logic were confusing though, that's another issue) 
However it's not a lot more code to write visitors so I can do that.

iamdanfox: This may look perfectly readable to you, but I think we need to try to make it super super obvious for any future maintainers.  Given that all the conjure union types just use FooVisitor, I think the consistency will be helpful to readers.

Aiming to use the simplest, most boring solution that works here!

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/MaximumVersionComparator.java
dansanduleac: yep, good idea

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependencyMerger.java
dansanduleac: This stuff is not passed to a logger, the Safe*Exception classes generate their own message out of the args

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependencyMerger.java
dansanduleac: it sucks that I have to handle the Optional when using `Stream.max`, when i know there will always be a result here

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/RecommendedProductDependencyMergerTest.groovy
dansanduleac: To construct, you can just do `RecommendedProductDependency(an inline map)`.
Here I want to construct taking into account some default values, without having to merge maps manually.

In the gradle plugin, the extension gives you a [configuration point that takes a `Closure`](https://github.com/palantir/sls-packaging/blob/85196ddbf8f10c9fb35f77c985602d3ae715415e/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependenciesExtension.groovy#L25)

dansanduleac: This is just so I can clone things so I don't have to re-type the same fields that never change... I didn't know a better way to do that in groovy

versions.props
dansanduleac: That is used, I just changed one of the versions in there to make them not compatible, such that the merger would kick in

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/MaximumVersionComparator.java
dansanduleac: https://github.com/palantir/sls-version-java/pull/3

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependencyMerger.java
iamdanfox: Let's switch to the guava ones for non-server usage

https://github.com/palantir/safe-logging/blob/develop/preconditions/src/main/java/com/palantir/logsafe/exceptions/SafeIllegalStateException.java

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/RecommendedProductDependencyMergerTest.groovy
```diff
@@ -0,0 +1,88 @@
...
+        given:
+        def dep1 = newRecommendation("2.0.0", "2.6.x", "2.2.0")
+        def dep2 = newRecommendation("2.7.0", "2.x.x", "2.8.0")
+
+        when:
+        def merged = RecommendedProductDependencyMerger.mergeRecommendedProductDependencies(dep1, dep2)
+
+        then:
+        def e = thrown(IllegalArgumentException)
+        e.message.contains("Inferred minimum version does not match inferred maximum version")
```
iamdanfox: Mention non-overlapping ranges

iamdanfox (approved):
Approved pending the exceptions & using the new OSS sls-versions jar

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/MaximumVersion.java
```diff
@@ -0,0 +1,127 @@
...
+ *     <li>an {@link OrderableSlsVersion}, or</li>
+ *     <li>an {@link SlsVersionMatcher}</li>
+ * </ul>
+ * Hence, this class encodes a functional either-type.
+ * <p>
+ * The dichotomy is necessary because {@link SlsVersionMatcher}, though it has some overlap with
+ * {@link OrderableSlsVersion}, doesn't handle all the cases that the latter supports. For instance, it can't represent
+ * non-release versions (see version types enum {@link SlsVersionType}).
+ */
+abstract class MaximumVersion implements Comparable<MaximumVersion> {
```
uschi2000, line 40: does this need to be `Comparable<>`?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/MaximumVersionComparator.java
```diff
@@ -0,0 +1,53 @@
...
+enum MaximumVersionComparator implements Comparator<MaximumVersion> {
+    INSTANCE;
+
+    @Override
+    public int compare(MaximumVersion o1, MaximumVersion o2) {
+        return o1.visit(new MaximumVersionVisitor<Integer>() {
+            @Override
+            public Integer visitVersion(OrderableSlsVersion version) {
+                // o1 is smaller if it satisfies the max constraint of o2
+                return o2.isSatisfiedBy(version) ? -1 : 1;
```
uschi2000, line 32: this seems to violate the `Comparable` interface when o1==o2?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependencyMerger.java
```diff
@@ -0,0 +1,90 @@
...
+import com.palantir.sls.versions.VersionComparator;
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+public final class RecommendedProductDependencyMerger {
+    private RecommendedProductDependencyMerger() { }
+
+    public static RecommendedProductDependency mergeRecommendedProductDependencies(
```
uschi2000, line 29: why not just `merge`?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependencyMerger.java
```diff
@@ -0,0 +1,90 @@
...
+            throw new IllegalArgumentException(String.format("Product names differ: '%s' and '%s'",
+                dep1.getProductName(), dep2.getProductName()));
+        }
+
+        OrderableSlsVersion minimumVersion = Stream
+                .of(
+                        OrderableSlsVersion.valueOf(dep1.getMinimumVersion()),
+                        OrderableSlsVersion.valueOf(dep2.getMinimumVersion()))
+                .max(VersionComparator.INSTANCE)
+                .orElseThrow(() -> new RuntimeException("Impossible"));
```
uschi2000, line 47: "Failed to determine which of dep1/dep2 is larger. This is a bug."

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependencyMerger.java
```diff
@@ -0,0 +1,90 @@
...
+                        && satisfiesMaxVersion(maximumVersion, version))
+                .max(VersionComparator.INSTANCE);
+
+        RecommendedProductDependency result = new RecommendedProductDependency();
+        result.setMinimumVersion(minimumVersion.toString());
+        maximumVersion.map(Objects::toString).ifPresent(result::setMaximumVersion);
+        recommendedVersion.map(Objects::toString).ifPresent(result::setRecommendedVersion);
+        result.setProductGroup(dep1.getProductGroup());
+        result.setProductName(dep1.getProductName());
+        // Verify validity
```
uschi2000, line 79: nit: this comment is redundant since the next line literally says `result.isValid();` . if anything, maybe consider renaming `isValid` to something like `verifyIsValid`. (does it throw?)

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/MaximumVersionComparatorTest.groovy
```diff
@@ -0,0 +1,68 @@
...
+import spock.lang.Specification
+
+class MaximumVersionComparatorTest extends Specification {
+    def comparesMatchers() {
+        given:
+        def m1 = MaximumVersion.valueOf("2.6.x")
+        def m2 = MaximumVersion.valueOf("2.x.x")
+        expect:
+        MaximumVersionComparator.INSTANCE.compare(m1, m2) < 0
+        MaximumVersionComparator.INSTANCE.compare(m2, m1) > 0
```
uschi2000, line 30: I don't think the Comparator interface makes sense here. `2.x.x` is neither larger nor smaller than `2.6.x`: it is simultaneously smaller, equal, and larger.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/RecommendedProductDependencyMergerTest.groovy
```diff
@@ -0,0 +1,88 @@
...
+        def merged = RecommendedProductDependencyMerger.mergeRecommendedProductDependencies(dep1, dep2)
+
+        then:
+        def e = thrown(IllegalArgumentException)
+        e.message.contains("Could not merge recommended product dependencies as their version ranges do not overlap")
+    }
+
+    def "fails if min == max"() {
+        given:
+        def dep1 = newRecommendation("2.5.0", "2.x.x", null)
```
uschi2000, line 56: why is `2.5.0` not a valid merged version here?

PR #294, opened 2018-06-14
AssetDistribution configuration should be mavenBundle

dansanduleac:
Should make the dependency conflict checking more sane first: #293


iamdanfox:
I know you've explained this to me before, but I still can't remember why this is desirable... Where is the code that requires the configuration to be called `mavenBundle` not `assetBundle`? Is it internal?

I think we'd need to push a change to the one internal repo that is using this old `assetBundle`.

Issue #261, opened 2017-12-14
init.sh check hangs


qinfchen:
here is the stacktrace log when running `init.sh check`:
```
Exception in thread "main" java.lang.RuntimeException: Unable to start Jetty server
    at <service>.start(Service.java:411)
    at<service>.startServer(Service.java:37)
    at <service>.main(Service.java:30)
Caused by: java.net.BindException: Address already in use
    at sun.nio.ch.Net.bind0(Native Method)
    at sun.nio.ch.Net.bind(Net.java:433)
    at sun.nio.ch.Net.bind(Net.java:425)
    at sun.nio.ch.ServerSocketChannelImpl.bind(ServerSocketChannelImpl.java:223)
    at sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:74)
    at org.eclipse.jetty.server.ServerConnector.openAcceptChannel(ServerConnector.java:334)
    at org.eclipse.jetty.server.ServerConnector.open(ServerConnector.java:302)
    at org.eclipse.jetty.server.AbstractNetworkConnector.doStart(AbstractNetworkConnector.java:80)
    at org.eclipse.jetty.server.ServerConnector.doStart(ServerConnector.java:238)
    at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:68)
    at org.eclipse.jetty.server.Server.doStart(Server.java:428)
    at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:68)
    at service.start(Service.java:409)
```

iamdanfox:
`init.sh check` was intended to work with a dropwizard healthcheck command.  Now that dropwizard is not the recommended server internally anymore, I think it might be worth deprecating this command. 

Your `Unable to start Jetty server` is probably because your WC server doesn't understand the `check` argument and is just trying to start another server instance.

qinfchen:
Didn't know that `init.sh check` is meant for dropwizard healthcheck command and neither the documentation nor `check.sh` indicates that. Agreed, I think we should deprecate it since it is still a command in all sls-artifacts and it won't work for almost of them

PR #296, opened 2018-06-21
Deprecate non-{start,status,stop} commands

gregakinman:
cc @cbrockington @iamdanfox 

I want to see what the proper route is for deprecating these commands. I'm thinking:

- add these echos for due diligence (though in this day and age, humans should generally not be running these commands, though I don't know what the large internal product is doing in this regard)
- calling this out in changelog for release that includes this deprecation
- email
- search for usages of these and follow up individually with products using them to lessen impact of their removal

(Also, restart is potentially still includable, though I initially left it out from the perspective of ruthless spec-adherence, there's no concept of instantaneous restarts in the spec and status is easily replaceable on the client-side with `stop && start`, which, as a sidenote, is probably how the current command should be anyway, as it currently will still issue a start again even if stop fails, small edge case)

### DIFF
--- a/gradle-sls-packaging/src/main/resources/init.sh
+++ b/gradle-sls-packaging/src/main/resources/init.sh
@@ -52,6 +52,8 @@ STATIC_LAUNCHER_CONFIG="service/bin/launcher-static.yml"
 CUSTOM_LAUNCHER_CONFIG="var/conf/launcher-custom.yml"
 STATIC_LAUNCHER_CHECK_CONFIG="service/bin/launcher-check.yml"
 
+DEPRECATION_MESSAGE="Command is deprecated: the next major release will only support start/status/stop"
+
 case $ACTION in
 start)
     if service/bin/init.sh status &> /dev/null; then
@@ -124,6 +126,7 @@ stop)
     fi
 ;;
 console)
+    echo $DEPRECATION_MESSAGE
     if service/bin/init.sh status &> /dev/null; then
         echo "Process is already running"
         exit 1
@@ -136,10 +139,12 @@ console)
     wait
 ;;
 restart)
+    echo $DEPRECATION_MESSAGE
     service/bin/init.sh stop
     service/bin/init.sh start
 ;;
 check)
+    echo $DEPRECATION_MESSAGE
     printf "%-50s" "Checking health of '$SERVICE'..."
     $LAUNCHER_CMD $STATIC_LAUNCHER_CHECK_CONFIG > var/log/$SERVICE-check.log 2>&1
     RESULT=$?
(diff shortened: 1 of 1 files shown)


cbrockington:
The only other potential step to add is before completely removing the commands, setting the commands to echo the fact they have been removed and exit 1 / some other error code.

I do find this to be overly optimistic

> add these echos for due diligence (though in this day and age, humans should generally not be running these commands, though I don't know what the large internal product is doing in this regard)

while that world would be nice, there will definitely be some who currently rely on the additional functionality (particularly for debugging), so while I don't think everything should be held back by them, I think it's reasonable to expect that this will ruffle some feathers.

gregakinman:
@iamdanfox thoughts?

cbrockington:
Other than `console` I also somewhat still don't understand the need for the deprecation.  Restart can be left in exactly as is, and check would be trivial to add to `go-init` as long as the underlying `start` function had the launcher config location as a param.

I appreciate I'm somewhat harping on about ways to avoid the deprecation, but it's because I still feel the deprecation is very hard to effectively communicate (since it's all bash, and as we mentioned before, there is no compilation / checking against bash APIs), so I consider avoiding changes there if possible to be pretty valuable.

gregakinman:
There's no "need" for deprecation per se, but I think the deprecation is just better/nicer. The mentality here is to support only what the SLS spec requires, since any deployment infrastructure managing SLS services will only be running commands dictated by the SLS spec, since those are the only things that are guaranteed to be present in an SLS service. Therefore, the only things running `console`, `restart`, and `check` are humans or custom scripts. `console` is explicitly for humans so should go away, `restart` seems like it would only be a convenience for a human to not have to do `stop`+`start`, and `check` doesn't even work and has already been marked for deprecation: https://github.com/palantir/sls-packaging/issues/261.

The deprecation is hard to communicate, yes, but as noted above, comes with the expectation that we will do due diligence to seek out uses of these and notify them. The removal of them will be pretty easy to communicate, though, since it will come with a major version bump.

cbrockington:
Since there is no need for it, can we focus on doing other work instead then?

gregakinman:
I think it's better to take this opportunity to remove some not-needed functionality rather than implement it in go-init (however easy it may be) just to support backcompat, especially when we don't deem the functionality necessary or in line with the trajectory of how our software is being managed. Just because something isn't necessary to do doesn't mean we shouldn't do it.

cbrockington:
I doesn't mean it shouldn't ever be done, it does mean we shouldn't do it now.  All the functions we don't want to implement in `go-init` can be left as they are, and we can push forward with other work.

Happy for us to create an issue tracking eventual deprecation of things don't need, but as a P3 or 4, not something we should be spending time on now.

If the tests were already working then I see no problem with merging this since the work is already done now, but given they aren't, I feel it would be more prudent to focus on things which directly and more immediately impact what we are trying to achieve.  (there is an internal burndown list for this).

gregakinman:
Closing in lieu of https://github.com/palantir/sls-packaging/pull/306

PR #299, opened 2018-06-26
Resolve dependencies for publishPlugins

dansanduleac:
#298 wasn't the right thing to do.
Gradle recommends doing something like [this](https://discuss.gradle.org/t/plugin-publish-fails-with-npe-if-dependency-doesnt-have-version/19070/3), this is less extreme than that but suffices to source the un-set versions from nebula.

### DIFF
--- a/gradle-sls-packaging/build.gradle
+++ b/gradle-sls-packaging/build.gradle
@@ -40,10 +40,10 @@ pluginBundle {
 
     withDependencies {
         it.each { dep ->
-         if (!dep.version) {
-             dep.version = dependencyRecommendations.getRecommendedVersion(dep.groupId, dep.artifactId)
-         }
-       }
+            if (!dep.version) {
+                dep.version = dependencyRecommendations.getRecommendedVersion(dep.groupId, dep.artifactId)
+            }
+        }
     }
 }
 
--- a/gradle-sls-packaging/build.gradle
+++ b/gradle-sls-packaging/build.gradle
@@ -3,7 +3,6 @@ apply plugin: 'java-gradle-plugin'
 apply plugin: 'groovy'
 apply plugin: 'org.inferred.processors'
 apply plugin: 'de.undercouch.download'
-apply plugin: 'nebula.maven-resolved-dependencies'
 
 dependencies {
     processor 'org.immutables:value'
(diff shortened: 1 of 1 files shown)


gradle-sls-packaging/build.gradle
```diff
@@ -38,6 +37,18 @@ pluginBundle {
             displayName = 'Creates SLS pod distributions'
         }
     }
+
+    // Fix for publishPlugins not working with dependency recommenders (nebula, in our case).
+    // Gradle suggests the 'withDependencies' approach in the following workaround:
+    // http://plugins.gradle.org/help/plugin/missing-dependency-version
```
qinfchen, line 43: @dansanduleac, do you know why it doesn't fix it. From the readme, the plugin is created exactly for this purpose.

dansanduleac: That plugin fixes the [POMs](https://github.com/nebula-plugins/nebula-publishing-plugin/blob/0a244ec3fba3c279553c437f5064b7bf98764486/src/main/groovy/nebula/plugin/publishing/maven/MavenResolvedDependenciesPlugin.groovy#L38) of maven publications. This is not a maven publication, but a bespoke code path belonging to `com.gradle.plugin-publish` (which publishes plugins to the Plugin Portal - not sure why it needs the deps of the whole plugin bundle, though!).
Other people have been confused as to why that plugin does nothing here: https://discuss.gradle.org/t/pluginpublish-task-not-respecting-nebula-maven-publish-rules-to-publish-a-resolved-pom/22832

gradle-sls-packaging/build.gradle
```diff
@@ -38,6 +37,18 @@ pluginBundle {
...
+    // TODO(dsanduleac): this is bad and we want to remove it
```
iamdanfox, line 44: I don't really want to introduce hacks and TODOs unless there is a clear way that future maintainers can figure out whether the hack can be removed.  For example, if there's an open issue we could link to that (when resolved) would allow us to remove this, then it would feel more OK.

iamdanfox:
Is there any way that we can see pre-merge whether this actually worked? An equivalent of publishToMavenLocal that we could look at the pom?

dansanduleac:
@dfox 2.13.2 was the pre-merge and it [worked](https://circleci.com/gh/palantir/sls-packaging/741) :)

iamdanfox (approved):
OK let's ship it :)

PR #300, opened 2018-06-28
Apply publish-jar.gradle after java-gradle-plugin

dansanduleac:
Turns out, the reason why gradle-baseline's `nebula` publication was correct and ours was not, was that they applied `java-gradle-plugin` before setting up the publication (baseline configures it explicitly, whereas we do it via `nebula.maven-publish`, same thing).
In our case, the wrong order meant that the `java` component (from which the publication was sourcing the dependencies & artifact) was empty when configuring the publishing extension. It gets configured eagerly, and it was being configured too early.


dansanduleac:
So great that with circle 2 you can see the pom it will produce and that it looks fine! 
https://754-41394255-gh.circle-artifacts.com/0/home/circleci/poms/gradle-sls-packaging/build/publications/nebula/pom-default.xml

PR #305, opened 2018-07-04
Timeout at 90 seconds

j-baker:
Our internal infrastructure times out before 240 seconds and this confused at least one dev internally. 90 matches the init script for our internal go services.

PR #309, opened 2018-07-20
kill -9 if fail to stop

markelliot:


gradle-sls-packaging/src/main/resources/init.sh
```diff
@@ -99,7 +99,7 @@ stop)
         PID=$(cat $PIDFILE)
         kill $PID
         COUNTER=0
-        while is_process_service $PID $SERVICE && [ "$COUNTER" -lt "240" ]; do
+        while is_process_service $PID $SERVICE && [ "$COUNTER" -lt "180" ]; do
```
iamdanfox: We've arbitrarily changed the shutdown timeouts in a variety of projects recently, and it's pretty hard to keep track of what the actual source of truth is.

Please could we just pick something a number and put it in the sls-spec, and then make everyone comply with this?  (I think some places might have lower timeouts than 3 minutes)

markelliot: I'm happy to leave this part alone and just add the kill -9 after the timeout. for a long time, the behavior of this project actually defined the number.

there's a discussion on the spec now, but IMO the spec should define the API, and the timeout here seems like an implementation detail to me.

jboreiko:
Magritte would like to see this go in.  Some of the bootvisor's struggles appear to come from failing to stop processes.  It's not really an option to allow the managed process to continue and it forces us to have to access the box to kill them.

The timeout still strikes me as too long given jetty's default graceful shutdown in 30 seconds.

jboreiko:
I would prefer something closer to https://github.com/palantir/sls-packaging/pull/305 for the timeout.

markelliot:
I've removed the timeout change so that can be considered separately

iamdanfox (approved):

iamdanfox:
cc @gregakinman - I guess we're gonna need to also add this functionality in go-init right?

gregakinman:
yeah thanks for the heads up

iamdanfox:
Confirming this `kill -9` functionality was also implemented in go-init: https://github.com/palantir/go-java-launcher/pull/49/files#diff-fdcff91f81dff786b4df5f55c08f423dR97

PR #312, opened 2018-07-31
Add first class gc setting

robert3005:
with sane defaults

cc @schlosna @iamdanfox

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.groovy
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.groovy
@@ -15,12 +15,29 @@
  */
 package com.palantir.gradle.dist.service
 
+import com.google.common.collect.ImmutableMap
 import com.palantir.gradle.dist.BaseDistributionExtension
+import com.palantir.gradle.dist.service.gc.GcProfile
+import com.palantir.gradle.dist.service.gc.Hybrid
+import com.palantir.gradle.dist.service.gc.ResponseTime
+import com.palantir.gradle.dist.service.gc.Throughput
+import javax.annotation.Nullable
+import org.gradle.api.Action
 import org.gradle.api.Project
+import org.gradle.api.model.ObjectFactory
+import org.gradle.util.ConfigureUtil
 
 class JavaServiceDistributionExtension extends BaseDistributionExtension {
 
+    private static final Map<String, Class<? extends GcProfile>> profileNames = ImmutableMap.of(
+            "throughput", Throughput.class,
+            "response-time", ResponseTime.class,
+            "hybrid", Hybrid.class)
+
+    private final ObjectFactory objects
+
     private String mainClass
+    private GcProfile gc
     private List<String> args = []
     private List<String> checkArgs = []
     private List<String> defaultJvmOpts = []
@@ -32,6 +49,8 @@ class JavaServiceDistributionExtension extends BaseDistributionExtension {
     JavaServiceDistributionExtension(Project project) {
         super(project)
         productType("service.v1")
+        objects = project.objects
+        gc = new Throughput()
     }
 
     void mainClass(String mainClass) {
@@ -90,6 +109,10 @@ class JavaServiceDistributionExtension extends BaseDistributionExtension {
         return mainClass
     }
 
+    GcProfile getGc() {
+        return gc
+    }
+
     List<String> getArgs() {
         return args
     }
@@ -117,4 +140,27 @@ class JavaServiceDistributionExtension extends BaseDistributionExtension {
     List<String> getExcludeFromVar() {
         return excludeFromVar
     }
+
+    void gc(String type, @Nullable Closure configuration) {
+        gc = objects.newInstance(profileNames[type])
+        if (configuration != null) {
+            ConfigureUtil.configure(configuration, gc)
+        }
+    }
+
+    void gc(String type) {
+        gc(type, null)
+    }
+
+    def <T extends GcProfile> void gc(Class<T> type, @Nullable Action<T> action) {
+        gc = objects.newInstance(type)
+        if (action != null) {
+            action.execute(gc)
+        }
+    }
+
+    def <T extends GcProfile> void gc(Class<T> type) {
+        gc(type, null)
+    }
+
 }
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.groovy
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.groovy
@@ -59,16 +59,26 @@ class JavaServiceDistributionPlugin implements Plugin<Project> {
 
         CreateStartScripts startScripts = CreateStartScriptsTask.createStartScriptsTask(project, 'createStartScripts')
         project.afterEvaluate {
-            CreateStartScriptsTask.configure(startScripts, distributionExtension.mainClass, distributionExtension.serviceName,
-                    distributionExtension.defaultJvmOpts, distributionExtension.enableManifestClasspath)
+            CreateStartScriptsTask.configure(
+                    startScripts,
+                    distributionExtension.mainClass,
+                    distributionExtension.serviceName,
+                    distributionExtension.defaultJvmOpts,
+                    distributionExtension.enableManifestClasspath)
         }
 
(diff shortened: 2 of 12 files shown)


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.groovy
```diff
@@ -59,16 +59,29 @@ class JavaServiceDistributionPlugin implements Plugin<Project> {
 
         CreateStartScripts startScripts = CreateStartScriptsTask.createStartScriptsTask(project, 'createStartScripts')
         project.afterEvaluate {
-            CreateStartScriptsTask.configure(startScripts, distributionExtension.mainClass, distributionExtension.serviceName,
-                    distributionExtension.defaultJvmOpts, distributionExtension.enableManifestClasspath)
+            def jvmOpts = distributionExtension.defaultJvmOpts + (distributionExtension.heapSize != null
+                    ? ["-Xmx" + distributionExtension.heapSize, "-Xms" + distributionExtension.heapSize] : [])
```
markelliot: We usually want to pick these up from our config management system/would expect these to be parameterized by map keys. Maybe consider pushing further upstream into named parameters in go-java-launcher?

robert3005: Makes sense - will remove from this pr and figure out correct wiring

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
```diff
@@ -31,6 +31,22 @@ class LaunchConfigTask extends DefaultTask {
 
     static final List<String> tmpdirJvmOpts = ['-Djava.io.tmpdir=var/data/tmp']
 
+    static final Map<String, List<String>> gcProfiles = [
+            'throughput'   : [],
```
robert3005: This means we would use jvm default which switched to G1 in 9. Need to put correct settings but at the same time java releases claim that from 10 onwards G1 should supersede parallelold collector

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
```diff
@@ -21,16 +21,35 @@ import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
...
 class LaunchConfigTask extends DefaultTask {
 
     static final List<String> tmpdirJvmOpts = ['-Djava.io.tmpdir=var/data/tmp']
 
+    static final Map<String, List<String>> gcProfiles = [
+            'throughput'   : ['-XX:+UseParallelOldGC'],
+            'response-time': ['-XX:+UseParNewGC',
+                              '-XX:+UseConcMarkSweepGC',
+                              '-XX:+CMSParallelInitialMarkEnabled',
+                              '-XX:+CMSParallelRemarkEnabled',
```
schlosna: We probably don't need to set `-XX:+CMSParallelInitialMarkEnabled -XX:+CMSParallelRemarkEnabled` as those should default to true with Java 8+, see https://github.com/AdoptOpenJDK/openjdk-jdk8u/blame/fcb5661ae078d0e3521a1237a67b3039b1bfdbe3/hotspot/src/share/vm/runtime/globals.hpp#L1792

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
```diff
@@ -21,16 +21,35 @@ import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
...
+                              '-XX:+UseCMSInitiatingOccupancyOnly',
+                              '-XX:CMSInitiatingOccupancyFraction=68',
+                              '-XX:+CMSClassUnloadingEnabled'],
```
schlosna: We probably want to add the following as well for CMS:

```
-XX:+ExplicitGCInvokesConcurrent
-XX:+ClassUnloadingWithConcurrentMark
```

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
```diff
@@ -21,16 +21,35 @@ import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
...
     static final List<String> tmpdirJvmOpts = ['-Djava.io.tmpdir=var/data/tmp']
 
+    static final Map<String, List<String>> gcProfiles = [
+            'throughput'   : ['-XX:+UseParallelOldGC'],
+            'response-time': ['-XX:+UseParNewGC',
+                              '-XX:+UseConcMarkSweepGC',
+                              '-XX:+CMSParallelInitialMarkEnabled',
+                              '-XX:+CMSParallelRemarkEnabled',
+                              '-XX:+UseCMSInitiatingOccupancyOnly',
+                              '-XX:CMSInitiatingOccupancyFraction=68',
```
schlosna: would love a way to make this percentage configurable as different workloads may need adjusted

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.groovy
```diff
@@ -21,6 +21,7 @@ import org.gradle.api.Project
 class JavaServiceDistributionExtension extends BaseDistributionExtension {
 
     private String mainClass
+    private String gcProfile = "throughput"
```
schlosna: do we actually want `throughput` as default?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.groovy
robert3005: so this still can be overridden by anything in launcher-custom. Not sure what's the best api here. I have never seen it set to anything else internally

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.groovy
robert3005: really - I just went with what's the current setup. For java 8 we might want to go with response-time.

robert3005:
Have a nicer way of expressing this - will update the pr tomorrow

robert3005:
updated with new approach - should let us be specific while only exposing the things we think ought to be exposed

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.groovy
```diff
@@ -117,4 +140,19 @@ class JavaServiceDistributionExtension extends BaseDistributionExtension {
     List<String> getExcludeFromVar() {
         return excludeFromVar
     }
+
+    void gc(String type, @Nullable Closure configuration) {
+        gc = objects.newInstance(profileNames[type])
```
robert3005, line 145: It looks like `plugins { ... } ` block doesn't add the plugin classpath to buildscript classpath so we can't reference classes directly, i.e. doing

```
gc (ResponseTime) { ... }
```

fails with MissingPropertyException `ResponseTime`. We could require use of classpath but that seemed a bit suboptimal. Happy to change this if there are better options.

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.groovy
```diff
@@ -117,4 +140,19 @@ class JavaServiceDistributionExtension extends BaseDistributionExtension {
...
+        if (configuration != null) {
```
robert3005, line 146: need to add test that makes sure we can just do `gc 'hybrid'` for instance

robert3005: added

ferozco (approved):
Since the default behaviour for java8 is the same as the previous behaviour, i think thats its good to go

PR #306, opened 2018-07-18
Update init.sh to use go-init

gregakinman:
fysa @cbrockington

PR #313, opened 2018-07-31
Use go-init in init.sh

gregakinman:
This is a rehash of https://github.com/palantir/sls-packaging/pull/306. I had to have the tests also use the repository `maven { url "http://palantir.bintray.com/releases" }` since `go-init` isn't available in JCenter.

According to @bluekeyes this can likely be resolved by contacting JCenter support, if we so desire.

@dansanduleac


ferozco:
Closing in favour of #316

PR #317, opened 2018-08-17
Use go-init 1.5.0

cbrockington:


cbrockington:
@iamdanfox is the new requirement on palantir.bintray.com both expected and acceptable?

cbrockington:
in fact also @dansanduleac since I feel I remember you previously talking to @gregakinman about this

ferozco:
@cbrockington I believe that the bintray requirement is only for external consumption

cbrockington:
Ah ok, since the tests run externally that makes sense.  Is this good to merge and rc?

ferozco (approved):
can you follow up by updating a few services and ensuring everything is in working order

gregakinman:
sweet, thanks @cbrockington!

PR #321, opened 2018-08-24
Add JVM Default GC configuration

raiju:
Enable the usage of the JVM default GC (or set GC using java options).

Follow-up to https://github.com/palantir/sls-packaging/pull/312 as it breaks back-compat for distributions that override GC methods as part of their own configuration. Should this be made the default GC setting?


robert3005:
The whole point of that pr was that per installation overrides should be rare. There's no reason to use jvm defaults and would like to avoid surprises when we switch jvms

raiju:
Closing for now

PR #326, opened 2018-09-06
Add -XX:+CMSScavengeBeforeRemark to response-time

robert3005:


ferozco (approved):

iamdanfox:
@robert3005 clarified in person that @j-baker and @schlosna were already OK with this as there was an internal discussion on `f/<ski-resort>/pull/539`

> it will cause more smaller pauses but potentially decreasing longer pauses and lowering likelihood of those.

PR #329, opened 2018-09-27
Run --scan for all gradle commands on CI

iamdanfox:
For OSS repos, I think this is preferable to `--profile` and possibly worth just adding to the template!


dansanduleac:
@iamdanfox shall we just add it to the template then? FWIW I also think we're not saving the profile to artifacts right now.

iamdanfox:
"do not merge" until we have confirmed the lawyers are actually happy with the https://gradle.com/terms-of-service - have fired off an email to IPninjas to check

PR #335, opened 2018-10-10
Push init.sh --help through to go-init.

cbrockington:
A side effect of this is that deprecated functions will no longer
appear in the help output of the command.

### DIFF
--- a/gradle-sls-packaging/src/main/resources/init.sh
+++ b/gradle-sls-packaging/src/main/resources/init.sh
@@ -42,15 +42,23 @@ STATIC_LAUNCHER_CHECK_CONFIG="service/bin/launcher-check.yml"
 
 DEPRECATION_MESSAGE="Command is deprecated: the next major release of sls-packaging will only support start/status/stop"
 
+function print_help() {
+    echo "$0 forwards commands to go-init with the following usage."
+    echo ""
+}
+
+if [[ "$ACTION" =~ ^(-h|--help)$ && "$2" =~ ^(|start|stop|status)$ ]]; then
+    print_help
+    exec $GO_INIT_CMD --help "$2"
+fi
+
 case $ACTION in
-start)
-    $GO_INIT_CMD start
-;;
-status)
-    $GO_INIT_CMD status
-;;
-stop)
-    $GO_INIT_CMD stop
+start|status|stop)
+    if [[ "$2" =~ ^(-h|--help)$ ]]; then
+        print_help
+        exec $GO_INIT_CMD "$ACTION" $2
+    fi
+    exec $GO_INIT_CMD "$ACTION"
 ;;
 console)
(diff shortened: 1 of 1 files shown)


gradle-sls-packaging/src/main/resources/init.sh
```diff
@@ -42,15 +42,22 @@ STATIC_LAUNCHER_CHECK_CONFIG="service/bin/launcher-check.yml"
...
 DEPRECATION_MESSAGE="Command is deprecated: the next major release of sls-packaging will only support start/status/stop"
 
+function print_help() {
+    echo "$0 forwards commands to go-init with the following usage."
+    echo ""
+    exec $GO_INIT_CMD "$@"
+}
+
+if [[ "$ACTION" =~ ^(-h|--help)$ && "$2" =~ ^(|start|stop|status)$ ]]; then
+    print_help --help "$2"
```
ferozco: We should re-order the arguments here, or the echo within `print_help` will be confusing

gradle-sls-packaging/src/main/resources/init.sh
```diff
@@ -42,15 +42,22 @@ STATIC_LAUNCHER_CHECK_CONFIG="service/bin/launcher-check.yml"
...
-start)
-    $GO_INIT_CMD start
-;;
-status)
-    $GO_INIT_CMD status
-;;
-stop)
-    $GO_INIT_CMD stop
+start|status|stop)
+    if [[ "$2" == "--help" ]]; then
```
ferozco: why do we only handle `--help` here, but `--help|-h` above?

gradle-sls-packaging/src/main/resources/init.sh
```diff
@@ -42,15 +42,22 @@ STATIC_LAUNCHER_CHECK_CONFIG="service/bin/launcher-check.yml"
 
 DEPRECATION_MESSAGE="Command is deprecated: the next major release of sls-packaging will only support start/status/stop"
 
+function print_help() {
```
ferozco, line 45: I don't think the name is correct here. This simply proxies args to `GO_INIT_CMD`

ferozco (changes requested):

gradle-sls-packaging/src/main/resources/init.sh
cbrockington: Can you expand on this, I don't think I understand.

gradle-sls-packaging/src/main/resources/init.sh
cbrockington: Fixed.

gradle-sls-packaging/src/main/resources/init.sh
cbrockington: It now just prints the help header.

ferozco (approved):

PR #343, opened 2018-10-17
Simplify CopyLauncherBinariesTask

dansanduleac:
## Before this PR

CopyLauncherBinariesTask was doing two project.copy invocations, setting up inputs/outputs manually

## After this PR

It's now a Sync task, which sets up inputs/outputs automatically, copies both files in one go, and also deletes any extraneous files from the target dir.


dansanduleac:
Needs some more work

PR #370, opened 2018-12-05
[fix] Ensure `gc: response-time` works on Java11

iamdanfox:
## Before this PR

If you configured a service to use the 'response-time' GC options, the service won't start up on OpenJdk11 as [UseParNewGC was removed in Java 10](https://bugs.openjdk.java.net/browse/JDK-8173421).

```gradle
distribution {
    serviceName 'your-service'
    mainClass 'com.palantir.service.Main'
    gc 'response-time'
}
```

```
Unrecognized VM option 'UseParNewGC'
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

cc @robert3005 

## After this PR

The service should start up.

PR #371, opened 2018-12-10
Add response-time-11 profile using shenandoahgc

robert3005:
@j-baker @iamdanfox turns out this is orthogonal to #370


iamdanfox (approved):

robert3005:
👍

PR #393, opened 2019-01-25
[fix] Restore old setter methods (without set*) and add lazy setters

dansanduleac:
## Before this PR

Rewrite in #385 removed some setter methods e.g. `JavaServiceDistributionExtension#mainClass(String)`.

## After this PR

Re-add old setter methods.


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java
```diff
@@ -89,15 +90,23 @@ public JavaServiceDistributionExtension(Project project) {
         return mainClass;
     }
 
-    public final void setMainClass(String mainClass) {
+    public final void mainClass(String newMainClass) {
+        this.mainClass.set(newMainClass);
+    }
+
+    public final void setMainClass(Provider<String> mainClass) {
```
ferozco: I don't think we should expose such a method. We should follow the pattern that other gradle task/extensions have set and only expose a getter to the underlying property. ex: https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/bundling/AbstractArchiveTask.html

ferozco:
👍

PR #398, opened 2019-01-30
[fix] revert copy task changes

ferozco:
The fix push in #387 caused it to fail in production with the following error:
```
java.lang.IllegalArgumentException: URI is not hierarchical
	at java.io.File.<init>(File.java:418)
	at com.palantir.gradle.dist.service.tasks.CopyYourkitAgentTask.copyYourkitAgent(CopyYourkitAgentTask.java:41)
```


dansanduleac:
Can we still fix the situation that was failing before? I.e. that you can't re-run those tasks

dansanduleac:
:+1:

ferozco:
👍

PR #402, opened 2019-02-05
[break] Detect constraints automatically

dansanduleac:
## Before this PR

Users had multiple ways to declare a product dependency (via a maven coordinate + recommended version, via a closure, or via a few methods declaring the fields explicitly).

When a library they depend on already declared a recommended product dependency, they'd have to explicitly repeat that dependency and use `detectConstraints` to pick up the constraints from the library, otherwise they'd get an error:

```gradle
productDependency {
    productGroup = 'foo'
    productName = 'bar'
    detectConstraints = true
}
```

## After this PR

_The functionality to detect constraints from libraries is now implicit and automatic._

As in, if any library publishes some product dependencies, they will be picked up and published automatically, without the user having to declare them explicitly.
Therefore, the `detectConstraints` property has been removed, since users would never have to use it manually anymore.

Users may declare a product dependency on top of one that comes from a library, though, if they want to further restrict the acceptable version range, and multiple constraints on the same product will be merged.

Also, `minimumVersion` and `maximumVersion` are required when declaring a product dependency via a closure:

```gradle
productDependency {
    productGroup = 'foo'
    productName = 'bar'
    minimumVersion = '1.0.0'
    maximumVersion = '1.x.x'
    recommendedVersion = '1.2.0' // optional
}
```

Alternatively it's still possible to use the short hand method forms:
```gradle
// defaults to maximumVersion: 1.x.x
productDependency('foo', 'bar', '1.0.0')

productDependency('foo', 'bar', '1.0.0', '2.0.x')
```


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.java
```diff
@@ -126,48 +118,7 @@ public final void setProductType(ProductType productType) {
...
-                dependencyName,
-                minVersion,
-                maxVersion == null
-                        ? generateMaxVersion(minVersion)
-                        : maxVersion,
-                recommendedVersion));
-    }
-
-    public final void productDependency(Closure closure) {
+    public final void productDependency(@DelegatesTo(ProductDependency.class) Closure closure) {
```
ferozco, line 175: Not sure we want to remove these methods, just seems like additional churn

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/MaximumVersion.java
```diff
@@ -25,7 +25,7 @@
 import java.util.Optional;
 
 /**
- * A typed version of a maximum version as declared in {@link RecommendedProductDependency#maximumVersion}.
+ * A typed version of a maximum version as declared in {@link ProductDependency#getMaximumVersion}.
```
ferozco: do we need this file? can we replace with methods exposed by sls-versions-java?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependency.groovy
```diff
@@ -61,48 +57,38 @@ class ProductDependency implements Serializable {
...
+            if (it == null || !SlsVersion.check(it) && !SlsVersionMatcher.safeValueOf(it).isPresent()) {
+                throw new IllegalArgumentException(
+                        "maximumVersion must be valid SLS version or version matcher: " + it)
             }
-
-            if (minimumVersion == maximumVersion) {
-                throw new IllegalArgumentException("minimumVersion and maximumVersion must be different "
-                        + "in product dependency on " + this.productName)
+        }
+        [minimumVersion, recommendedVersion].each {
```
ferozco: I think these short hands are undesirable, since they don't really improve readability of the code and make the error messages less precise

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependency.groovy
```diff
@@ -61,48 +57,38 @@ class ProductDependency implements Serializable {
...
+            if (!SlsVersion.check(it)) {
+                throw new IllegalArgumentException(
+                        "minimumVersion and recommendedVersion must be valid SLS versions: " + it)
             }
         }
+        if (minimumVersion == maximumVersion) {
+            throw new IllegalArgumentException("minimumVersion and maximumVersion must be different "
+                    + "in product dependency on " + this.productName)
+        }
```
ferozco, line 102: Need to check the recommended version is in the right range if specified

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
```diff
@@ -142,109 +136,43 @@ public final void setManifestFile(File manifestFile) {
...
-                            existingDep);
-                    dep = RecommendedProductDependencyMerger.merge(recommendedDep, existingDep);
-                }
-
-                mavenCoordsByProductIds.put(productId, coord);
-                recommendedDepsByProductId.put(productId, dep);
-            });
+        if (manifestExtensions.containsKey("product-dependencies")) {
+            throw new IllegalArgumentException("Use productDependencies configuration option instead of setting "
+                    + "'service-dependencies' key in manifestExtensions");
```
ferozco: s/service-dependencies/product-dependencies/

ferozco (changes requested):
We should also update the docs and call out the change in behaviour when users manually specified a full product dependency

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/MaximumVersion.java
dansanduleac: yes we do, it allows comparing two types from sls-versions-java that aren't otherwise comparable.

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependency.groovy
dansanduleac: fair

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependency.groovy
dansanduleac: This gets tricky because it might not be an OrderableSlsVersion in which case you just can't compare.
Also we made it required, so it must be specified.

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.java
dansanduleac: Sure, we can keep them and specify defaults instead of null for recommendedVersion

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/MaximumVersion.java
dansanduleac: as discussed offline, since we don't actually support arbitrary OrderableSlsVesion for maximumVersion, removed this class and simplified the merging logic.

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.java
```diff
@@ -129,11 +130,14 @@ public final void setProductType(ProductType productType) {
     public final void productDependency(String mavenCoordVersionRange) {
         productDependency(mavenCoordVersionRange, null);
     }
+
     public final void productDependency(String mavenCoordVersionRange, String recommendedVersion) {
         Matcher matcher = MAVEN_COORDINATE_PATTERN.matcher(mavenCoordVersionRange);
-        Preconditions.checkArgument(matcher.matches(), "String '%s' is not a valid maven coordinate. "
+        if (!matcher.matches()) {
+            throw new IllegalArgumentException("String '${mavenCoordVersionRange}' is not a valid maven coordinate. "
```
ferozco: not groovy

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/ProductDependencyTest.groovy
```diff
@@ -40,7 +40,7 @@ class ProductDependencyTest extends Specification {
         new ProductDependency("", "", null, "1.2.x", "1.2.3").isValid()
 
         then:
-        thrown(IllegalArgumentException)
+        thrown(NullPointerException)
```
ferozco: remove this test

ferozco (approved):

markelliot (commented):
It would’ve been better for this PR to include a docs update explaining how resolution works now. It seems that as a result of this change:
 - it’s not possible to ignore an automatic dependency
 - the behavior when a conflict is encountered seems less clear

markelliot:
Also, this change seems to require that consumers must now know the min/max bounds, and the original reason for adding this behavior at all in this plugin was to put the ranges in control of producers, who best know/understand these ranges. That’s pretty negative, IMO.

dansanduleac:
Sorry if the description wasn't clear - the functionality to detect constraints is not gone, it's just that it's now the *default*.
As in, you don't need to explicitly specify a product with `detectConstaints = true`. If any product dependencies are found, their constraints are picked up and published. If the user also declares an explicit dependency on the same product, then they are first merged.

dansanduleac:
I'll follow up with a PR to make the readme reflect this.

uschi2000:
@dansanduleac can you please make the link between `Product dependencies are also automatically populated from library dependencies` and `Recommended Product Dependencies Plugin` more explicit? Reading top-down it's hard to grok how exactly SLS dependencies are inferred from libraries.

dansanduleac:
> @dansanduleac can you please make the link between `Product dependencies are also automatically populated from library dependencies` and `Recommended Product Dependencies Plugin` more explicit? Reading top-down it's hard to grok how exactly SLS dependencies are inferred from libraries.

https://github.com/palantir/sls-packaging/pull/405

PR #404, opened 2019-02-06
[fix] Don't include self dependencies

dansanduleac:
## Before this PR

A regression introduced in #402 meant that products which have an API that recommends _that product_ will see a product dependency on _that same product_ written to their manifest.

## After this PR

Dependencies on the same product are no longer included in the generated `manifest.yml`

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
@@ -146,6 +146,10 @@ public class CreateManifestTask extends DefaultTask {
         Map<ProductId, ProductDependency> allProductDependencies = Maps.newHashMap();
         getProductDependencies().get().forEach(declaredDep -> {
             ProductId productId = new ProductId(declaredDep.getProductGroup(), declaredDep.getProductName());
+            Preconditions.checkArgument(!serviceGroup.get().equals(productId.getProductGroup())
+                    || !serviceName.get().equals(productId.getProductName()),
+                    "Invalid for product to declare an explicit dependency on itself, please remove: %s",
+                    declaredDep);
             if (getIgnoredProductIds().get().contains(productId)) {
                 throw new IllegalArgumentException(String.format(
                         "Encountered product dependency declaration that was also ignored for '%s', either remove the "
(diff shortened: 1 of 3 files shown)


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
```diff
@@ -215,8 +215,8 @@ final void createManifest() throws Exception {
             }
 
             try {
-                RecommendedProductDependencies recommendedDeps = jsonMapper.readValue(pdeps.get(),
-                        RecommendedProductDependencies.class);
+                RecommendedProductDependencies recommendedDeps =
```
ferozco, line 222: This technically allows users to declare a dependency on themselves... I would move the logic and remove the dependency right before we serialise the manifest

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
```diff
@@ -223,9 +223,7 @@ class CreateManifestTaskIntegrationSpec extends GradleIntegrationSpec {
         runTasks(':testCreateManifest')
 
         then:
-        def manifest = CreateManifestTask.jsonMapper.readValue(
-                file('build/deployment/manifest.yml', projectDir).text, Map)
-        manifest.get("extensions").get("product-dependencies").size() == 1
```
dansanduleac, line 228: this is bad style and short-circuits the better introspection that the check on the following line gives you, which is to actually show a diff

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
dansanduleac: Not sure if moving is the right answer, since then users could still declare a dependency on themselves, even if it's then removed.
I'd instead also check the explicitly declared dependencies and throw if any of them is a self dependency.

ferozco: ok, we should that additional verification then

ferozco (approved):

PR #406, opened 2019-02-06
[fix] createManifest is usable on a dirty repo

iamdanfox:
## Before this PR

With 3.0.0-rc2, it's impossible to run `./gradlew createManifest` on a dirty repo.

## After this PR


iamdanfox:
New idea: ProductDependency should just be strings, with no built-in validation. Then, in createManifest and when we embed in jars, we perform super strict validation, but only if the $CI environment variable is set.

This ensures people can still develop and run tests with .dirty versions, but should never be able to publish them.

ferozco:
👍

PR #412, opened 2019-02-11
[break] Delete `reponse-time-11` GcProfile

iamdanfox:
## Before this PR

We [added](https://github.com/palantir/sls-packaging/pull/371) a somewhat experimental `response-time-11` GcProfile which relied on Shenandoah-enabled OpenJDK11 JREs.  Unfortunately, these JREs don't run on Centos6 boxes, so while some on-prem deployments still need to work on Centos6, we likely won't be able to recommend using this widely.

## After this PR

The option is gone, so hopefully people won't shoot themselves in the foot.


ferozco:
👍

PR #414, opened 2019-02-11
[fix] Mask minimum dependency from lockfile on service pubished in repo if same as project version

dansanduleac:
## Before this PR

If there are multiple services in a single repo and there are interdependencies, then the project's version (which changes with every commit) gets written to some product-dependencies.lock files, forcing them to change with every commit.

This is undesirable as it will cause every PR to conflict.

## After this PR

Mask the minimum version of a product dependency, if it refers to a service published in the same repo, and the version is the same as that project's version.

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyLockFile.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyLockFile.java
@@ -17,21 +17,30 @@
 package com.palantir.gradle.dist;
 
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 public final class ProductDependencyLockFile {
 
     private static final String HEADER = "# Run ./gradlew --write-locks to regenerate this file\n";
 
-    public static String asString(List<ProductDependency> deps) {
-        return deps.stream()
-                    .map(dep -> String.format("%s:%s (%s, %s)",
-                            dep.getProductGroup(),
-                            dep.getProductName(),
-                            dep.getMinimumVersion(),
-                            dep.getMaximumVersion()))
-                    .sorted()
-                    .collect(Collectors.joining("\n", HEADER, "\n"));
+    public static String asString(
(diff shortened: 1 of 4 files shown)


gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
```diff
@@ -332,10 +334,110 @@ class CreateManifestTaskIntegrationSpec extends GradleIntegrationSpec {
...
+                mainClass 'com.palantir.foo.bar.MyServiceMainClass'
+                args 'server', 'var/conf/my-service.yml'
+            }
+        """.stripIndent())
+
+        when:
+        runTasks('--write-locks')
+
+        then:
+        file('bar-server/product-dependencies.lock').readLines().contains 'com.palantir.group:foo-service (project-version, 1.x.x)'
```
iamdanfox: Could we use `$project-version` or `$version` instead of just `project-version` to emphasise this value is somehow special?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyLockFile.java
```diff
@@ -17,21 +17,31 @@
...
+            List<ProductDependency> deps, Set<ProductId> servicesDeclaredInProject, String projectVersion) {
+        return deps.stream().map(dep -> String.format(
+                "%s:%s (%s, %s)",
+                dep.getProductGroup(),
+                dep.getProductName(),
+                renderDepMinimumVersion(servicesDeclaredInProject, projectVersion, dep),
+                dep.getMaximumVersion())).sorted().collect(Collectors.joining("\n", HEADER, "\n"));
+    }
+
+    private static String renderDepMinimumVersion(
```
iamdanfox, line 43: Surely this needs a little javadoc explaining why we're doing this weird masking?

dansanduleac: sure

iamdanfox:
:+1:

PR #417, opened 2019-02-12
Fix ProductDependencyLockFile header

diogoholanda:

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyLockFile.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyLockFile.java
@@ -22,7 +22,7 @@ import java.util.stream.Collectors;
 
 public final class ProductDependencyLockFile {
 
-    private static final String HEADER = "# Run ./gradlew --write-locks to regenerate this file\n";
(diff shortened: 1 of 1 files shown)


ferozco (approved):

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyLockFile.java
```diff
@@ -22,7 +22,7 @@
 
 public final class ProductDependencyLockFile {
 
-    private static final String HEADER = "# Run ./gradlew --write-locks to regenerate this file\n";
+    private static final String HEADER = "# Run ./gradlew createManifest --write-locks to regenerate this file\n";
```
iamdanfox, line 25: mmm not sure about this because both work

iamdanfox (changes requested):
Not sure

diogoholanda:
I think this was a red herring. I think the actual issue is that the lock file is not used as input to the task?
What I am seeing is that, if the manifest is already created, even if I remove the lock file or change it, or try to regenerate it, nothing actually gets done

iamdanfox:
Ah dang I think that missing 'input' declaration thing sounds exactly right, because the checking and re-generation is done in the body of the task and if gradle decides the task is already up to date then it won't actually run that body!

@dansanduleac any chance you can pop a fix in for this?

PR #419, opened 2019-02-12
[fix] Re-run CreateManifestTask if lock file changed

dansanduleac:
## Before this PR

Modifying the `product-dependencies.lock` file after it has been written would not cause `createManifest` to re-run (it would be considered UP_TO_DATE), which means it wouldn't verify if the locks are still correct.

## After this PR

Mark `product-dependencies.lock` as an _input_ of CreateManifestTask, so the task is re-run if the file changed.

### DIFF
--- a/gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
+++ b/gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
@@ -52,14 +52,23 @@ class CreateManifestTaskIntegrationSpec extends GradleIntegrationSpec {
                 productDependenciesConfig = configurations.runtime
             }
         """.stripIndent()
-        file('product-dependencies.lock').text = """\
-        # Run ./gradlew --write-locks to regenerate this file
-        group:name (1.0.0, 1.x.x)
-        group:name2 (2.0.0, 2.x.x)
-        """.stripIndent()
     }
 
     def 'fails if lockfile is not up to date'() {
+        buildFile << """
+            dependencies {
+                runtime 'b:b:1.0'
+            }
+        """.stripIndent()
+
+        file('product-dependencies.lock').text = """\
+            # Run ./gradlew --write-locks to regenerate this file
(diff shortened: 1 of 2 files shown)


iamdanfox:
just one teeny unit test?

dansanduleac:
three unit tests! and they proved that the `@OutputFile` thing didn't work...

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
```diff
@@ -52,14 +53,23 @@ class CreateManifestTaskIntegrationSpec extends GradleIntegrationSpec {
                 productDependenciesConfig = configurations.runtime
             }
         """.stripIndent()
-        file('product-dependencies.lock').text = """\
-        # Run ./gradlew --write-locks to regenerate this file
-        group:name (1.0.0, 1.x.x)
-        group:name2 (2.0.0, 2.x.x)
-        """.stripIndent()
```
dansanduleac, line 59: I didn't like that we set up an inconsistent state by default (no lock file should be the correct original state)

iamdanfox (approved):
Thanks @dansanduleac!  (cc @diogoholanda I think this should fix the local dev issue you encountered earlier)

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
```diff
@@ -76,6 +86,61 @@ class CreateManifestTaskIntegrationSpec extends GradleIntegrationSpec {
...
+        runTasks('createManifest').task(':createManifest').outcome == TaskOutcome.UP_TO_DATE
+
+        when:
+        file('product-dependencies.lock') << '\nthis should not be here'
+
+        then:
+        runTasksAndFail('createManifest').task(':createManifest').outcome == TaskOutcome.FAILED
+    }
+
+    def 'fails if lockfile has suddenly disappeared since last invocation'() {
```
ferozco: fails if unexpected lockfile exists

PR #420, opened 2019-02-12
[fix] createManifest is always run as part of check

ferozco:
## Before this PR
There was no explicit link between `createMaifest` and `check` allowing for a build to pass without the product-dependency lock file having been verified.

## After this PR
We explicitly add a dependency between `check` and `createManifest`

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
@@ -386,6 +386,9 @@ public class CreateManifestTask extends DefaultTask {
                 });
         project.afterEvaluate(p ->
                 createManifest.configure(task -> task.setManifestExtensions(ext.getManifestExtensions())));
+        project.getPluginManager().withPlugin("base", p -> {
+            project.getTasks().getByName("check").dependsOn(createManifest);
+        });
 
         // We want `./gradlew --write-locks` to magically fix up the product-dependencies.lock file
         // We can't do this at configuration time because it would mess up gradle-consistent-versions.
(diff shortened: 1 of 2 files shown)


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
```diff
@@ -386,6 +386,9 @@ private void validateProjectVersion() {
                 });
         project.afterEvaluate(p ->
                 createManifest.configure(task -> task.setManifestExtensions(ext.getManifestExtensions())));
+        project.getPluginManager().withPlugin("base", p -> {
+            project.getTasks().getByName("check").dependsOn(createManifest);
```
dansanduleac: ```.named(LifecycleBasePlugin.CHECK_TASK_NAME).configure```

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
```diff
@@ -442,6 +442,16 @@ class CreateManifestTaskIntegrationSpec extends GradleIntegrationSpec {
         file('bar-server/product-dependencies.lock').readLines().contains 'com.palantir.group:foo-service (0.0.0, 1.x.x)'
     }
 
+    def "check depends on createManifest"() {
+        file("product-dependencies.lock").delete()
```
dansanduleac: why?

ferozco: so the task doesn't fail

dansanduleac: yep setup() was broken before - fixed as part of https://github.com/palantir/sls-packaging/pull/419#discussion_r256128260 so it shouldn't be necessary anymore

dansanduleac:
👍

PR #424, opened 2019-02-13
[improvement] When enabled, hprof heap dumps are written to var/log

### LINKED ISSUE #423: Specify hprof directory
## What happened?

The current setting for `HeapDumpPath` is currently implicitly `-XX:HeapDumpPath=./java_pid<pid>.hprof`.

In the same spirit as https://github.com/palantir/sls-packaging/pull/314, specify `-XX:HeapDumpPath=var/log`

raiju:
Fixes #423

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
@@ -58,6 +58,7 @@ public class LaunchConfigTask extends DefaultTask {
             "-XX:+CrashOnOutOfMemoryError",
             "-Djava.io.tmpdir=var/data/tmp",
             "-XX:ErrorFile=var/log/hs_err_pid%p.log",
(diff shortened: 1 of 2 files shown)


ferozco:
@raiju We set `CrashOnOutOfMemoryError` so no heap dump should be produced

carterkozak:
@ferozco The two parameters are not mutually exclusive, the crash occurs after a heap dump is written to disk. While we do not enable `-XX:+HeapDumpOnOutOfMemoryError` by default, it is reasonable to enable while monitoring/debugging a failing service.

ferozco:
Ok, I thought that they were. I'm hesitant to encourage enabling heap dumps though.  Enabling heap dumps can be a foot gun since a bad release of a single service could crash multiple times taking down the box with it because the disk fills up.

raiju:
@ferozco This isn't enabling heap dumps. It's only specifying where heap dumps are written if enabled so we can hook it into our metrics (alert on heap dump creation), and is in a place where there is some file rotation policy in place.

ferozco:
👍

robert3005:
@raiju can you point me to the services which are producing heap dumps that you care about?

PR #427, opened 2019-02-14
[improvement] Lazy product dependency closure

dansanduleac:
## Before this PR

`productDependency` blocks are evaluated right away, at configuration time.

If you need to look up a dynamic version of a dependency in a `productDependency` closure, this would force a configuration resolution at configuration-time, which is [discouraged](https://guides.gradle.org/performance/#don_t_resolve_dependencies_at_configuration_time).

## After this PR

`productDependency` taking a closure is configured lazily.
This is the case for both the `BaseDistributionExtension` (`distribution`) and `RecommendedProductDependenciesExtension` (`recommendedProductDependencies`).

The closure is not run until the values are needed, which is typically during the task execution phase.


iamdanfox:
:+1:

PR #430, opened 2019-02-20
[fix] Don't inject services that require gradle 5.2

dansanduleac:
## Before this PR

#427 introduced a regression such the recommended-product-dependencies plugin would now require gradle 5.2 (because we used [this feature](https://docs.gradle.org/5.2/release-notes.html#service-injection-into-plugins-and-project-extensions))

## After this PR

Implement the extension in a gradle 4.10-compatible way

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependenciesExtension.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependenciesExtension.java
@@ -20,7 +20,7 @@ import groovy.lang.Closure;
 import groovy.lang.DelegatesTo;
 import java.util.Set;
 import javax.inject.Inject;
-import org.gradle.api.model.ObjectFactory;
+import org.gradle.api.Project;
 import org.gradle.api.provider.ProviderFactory;
 import org.gradle.api.provider.SetProperty;
 import org.gradle.util.ConfigureUtil;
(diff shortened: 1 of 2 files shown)


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependenciesExtension.java
```diff
@@ -30,9 +30,9 @@
     private final ProviderFactory providerFactory;
 
     @Inject
-    public RecommendedProductDependenciesExtension(ObjectFactory objectFactory, ProviderFactory providerFactory) {
-        this.recommendedProductDependencies = objectFactory.setProperty(ProductDependency.class).empty();
-        this.providerFactory = providerFactory;
+    public RecommendedProductDependenciesExtension(Project project) {
```
ferozco, line 33: Do we not have to pass in the project when we create the extension? ex: https://github.com/palantir/sls-packaging/blob/develop/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/asset/AssetDistributionPlugin.java#L50

dansanduleac: we might!

ferozco: why didn't this fail originally? does gradle auto inject some things?

ferozco:
👍

PR #432, opened 2019-02-21
[improvement] log what artifacts are making what recommendations

tpetracca:
## Before this PR
I was surprised to see that my product dependency lock files had stricter version constraints on some of my dependencies than I expected, but didn't have a quick/easy way of figuring out why the versions specified were actually there.

## After this PR
Adds a debug level log that prints out for product dependency recommendation, what the recommending artifact is and what the recommendation itself is.

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
@@ -342,7 +342,13 @@ public class CreateManifestTask extends DefaultTask {
                                 recommendedDep.getProductName(),
                                 recommendedDep.getMinimumVersion(),
                                 recommendedDep.getMaximumVersion(),
-                                recommendedDep.getRecommendedVersion()));
+                                recommendedDep.getRecommendedVersion()))
+                        .peek(productDependency -> log.debug(
+                                "Product dependency recommendation made by artifact '{}', file '{}', dependency "
+                                + "recommendation '{}'",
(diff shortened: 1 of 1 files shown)


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
```diff
@@ -342,7 +342,13 @@ private void ensureLockfileIsUpToDate(List<ProductDependency> productDeps) {
                                 recommendedDep.getProductName(),
                                 recommendedDep.getMinimumVersion(),
                                 recommendedDep.getMaximumVersion(),
-                                recommendedDep.getRecommendedVersion()));
+                                recommendedDep.getRecommendedVersion()))
+                        .peek(productDependency -> log.debug(
```
iamdanfox: debug is actually crazily noisy - shall we take this up to info?

```suggestion
                        .peek(productDependency -> log.info(
```

ferozco:
Could we provide functionality similar to gradle-consistent-version's `why` task which allows you to ask about the provenance of a specific dependency instead of asking users to read through the logs

dansanduleac:
@ferozco that would be awesome, but not sure we need to do it in this PR, unless you'd rather we _only_ do that

ferozco:
I'm suggesting implementing the provenance command instead of thisw

iamdanfox (approved):
I think this is a strict improvement over the current world, and doesn't preclude us from implementing a more first-class `why` task!

dansanduleac:
👍

PR #433, opened 2019-02-21
[fix] go-java-launcher 1.6.0 -> 1.6.1 to allow digits in process name

iamdanfox:
## Before this PR

@hsaraogi and @hturki found that upgrading from 2.X -> 3.X stopped allowing them to use a service name with numbers in them!

## After this PR

This regression is now fixed.


dansanduleac:
👍

PR #435, opened 2019-02-21
[fix] Allow merging non-orderable minimum versions if they are equal

dansanduleac:
## Before this PR

Minimum version was only enforced to be an `SlsVersion` but not an `OrderableSlsVersion`. However this means that merging code could then fail with a bad error if all minimum versions were not parseable.
This could happen if the repo was dirty as that would cause the version to look like `1.0.0-g3872819.dirty`.

## After this PR

* Allow merging non-orderable minimum versions in the ProdutDependencyMerger as long as they are equal. Added test to verify that this fails predictably when they are not equal.
* Improve the error messages to all include the product group/name when a problem is encountered.


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.java
```diff
@@ -181,8 +186,18 @@ public final void productDependency(
     public final void productDependency(@DelegatesTo(ProductDependency.class) Closure closure) {
         productDependencies.add(providerFactory.provider(() -> {
             ProductDependency dep = new ProductDependency();
-            ConfigureUtil.configureUsing(closure).execute(dep);
-            dep.isValid();
+            try {
+                ConfigureUtil.configureUsing(closure).execute(dep);
+                if (dep.getMinimumVersion() != null && dep.getMaximumVersion() == null) {
```
ferozco: This is pretty unexpected. Can we call this out in the changelog/PR description and add some docs about the new behaviour

ferozco (commented):
Core code change looks good. Maybe we can do the change to how you define product dependencies in another PR?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.java
dansanduleac: yep fair I can move this bit out

ferozco:
👍

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependency.groovy
```diff
@@ -79,72 +80,86 @@ class ProductDependency implements Serializable {
...
         }
 
         if (recommended.isPresent()) {
             Preconditions.checkArgument(
                     maximum.compare(recommended.get()) >= 0,
-                    "Recommended version (%s) is greater than maximum version (%s)",
-                    recommendedVersion, maximumVersion)
+                    "Recommended version is greater than maximum version",
+                    SafeArg.of("recommendedVersion", recommendedVersion),
+                    SafeArg.of("maximumVersion", maximumVersion))
```
iamdanfox, line 110: I'm not convinced all these SafeArgs are worth it here... this is a gradle plugin so we don't ever ship structured logs anywhere???

PR #436, opened 2019-02-21
Upgrade to consistent-versions

dansanduleac:
## Before this PR

Not using consistent versions for dependency versions.

## After this PR

Using [consistent-versions](https://github.com/palantir/gradle-consistent-versions)


iamdanfox:
:+1:

PR #437, opened 2019-02-22
[improvement] Derive max version in closure if missing

dansanduleac:
## Before this PR

Users could specify dependencies like:
```gradle
productDependency('group', 'name', '1.0.0')
```
and a maximum version of `1.x.x` would be auto-generated from their supplied minimum version.

The same wasn't true of configuring through a closure though:

```gradle
productDependency {
  productGroup 'group'
  productName 'name'
  minimumVersion '1.0.0'
}
```

Would fail validation as maximum version is null.

However, a lot of projects use this form of configuring dependencies, and their logic for the maximum version is almost always to manually parse minimumVersion in order to derive the equivalent semantic version maximum.

## After this PR

This code is now valid, and generates a maximumVersion of `1.x.x`:

```gradle
productDependency {
  productGroup 'group'
  productName 'name'
  minimumVersion '1.0.0'
}
```


ferozco (commented):
Looks good, can we test the new behaviour?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/BaseDistributionExtension.java
```diff
@@ -181,8 +183,16 @@ public final void productDependency(
     public final void productDependency(@DelegatesTo(ProductDependency.class) Closure closure) {
         productDependencies.add(providerFactory.provider(() -> {
             ProductDependency dep = new ProductDependency();
-            ConfigureUtil.configureUsing(closure).execute(dep);
-            dep.isValid();
+            try {
+                ConfigureUtil.configureUsing(closure).execute(dep);
+                dep.isValid();
+            } catch (Exception e) {
```
iamdanfox, line 195: how does this change have the effect described in the PR description?? Is this PR unfinished?

iamdanfox:
Waiting on @dansanduleac to push some lines??

ferozco (approved):

PR #438, opened 2019-02-22
Fix publishing

dansanduleac:
Regression introduced in #436 meant we can't publish


iamdanfox:
Gross, but to unblock publishing... :+1:

PR #439, opened 2019-02-25
[improvement] go-java-launcher 1.5.1 -> 1.6.1, to allow $JAVA_11_HOME

iamdanfox:
## Before this PR

We're trying to excavate out Java11 everywhere internally, but the PR has ended up with a slightly unnecessary dependency on the sls-packaging 2.X -> 3.X upgrade.

## After this PR

This is just a little backport so that Java11 excavator PR doesn't require people to do the 2.X -> 3.X upgrade... 

If we want to go with this approach, then I'd tag a `2.23.0` on the `release/2.x` branch.


iamdanfox:
Whoops looks like this merged a bit early.  @ferozco let me know what you think in  the morning and either run with this or delete the branch entirely.

PR #444, opened 2019-02-26
[fix] Declare dependency on safe-logging to pick up correct minimum version

### LINKED ISSUE #443: Applying sls-asset-distribution leads to NoClassDefFoundError com/palantir/logsafe/exceptions/SafeRuntimeException
## What happened?

Applying just sls-packaging leads to NoClassDefFoundError

 ```
> Failed to apply plugin [id 'com.palantir.sls-asset-distribution']
   > Could not create an instance of type com.palantir.gradle.dist.asset.AssetDistributionExtension.
      > Could not generate a decorated class for class com.palantir.gradle.dist.asset.AssetDistributionExtension.
         > com/palantir/logsafe/exceptions/SafeRuntimeException
```

no error should have been thrown

robert3005:
## Before this PR
#443 

## After this PR
fixes #443 

This is happening because buildscript classpath probably doesn't have improved pom support


ferozco:
👍

PR #448, opened 2019-02-27
[improvement] Upgrade YourKit 2017.02-b75 -> 2019.1-b112

yixunx:
## Before this PR

YourKit sometimes crashes when using a client on Mac OS X to connect to an sls service running on a remote server.

## After this PR

I confirmed that the crash no longer happens after upgrading to the latest YourKit build, possibly due to the fix in build 110 (https://www.yourkit.com/download/yjp_2019_1_builds.jsp). Upgrading the library in case other people are hitting the same problem. 

Warning: this will require users to upgrade their clients which might require getting another license. Feel free to close the PR if we want to avoid such breaks.

### DIFF
--- a/gradle-sls-packaging/build.gradle
+++ b/gradle-sls-packaging/build.gradle
@@ -45,7 +45,7 @@ pluginBundle {
     }
 }
 
-def yourkitVersion = "2017.02-b75"
+def yourkitVersion = "2019.1-b112"
 def yourkitFilename = "YourKit-JavaProfiler-${yourkitVersion}"
 
 task downloadYourkitDist(type: Download) {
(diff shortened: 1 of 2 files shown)


iamdanfox:
Thanks for the PR @yixunx, always in favour of staying up to date with latest bugfixes & improvements.

Out of curiosity, what kind of error message would you get if you had an old YourKit client and tried to connect to a running server with the new libyjpagent.so?

ellisjoe:
You get something of the form:
```
Cannot connect to the profiled application. The application has been started with incompatiable profile agent version:

Client version: YourKit Java Profiler 2014 build 14090
Agent version: YourKit Java Profiler 2013 build 13088
```

I've regularly seen this indicating the Client and Agent versions are tightly coupled.

yixunx:
^ Yes this is the error I saw as well. Mine looked like this:

image

iamdanfox:
Cool, in that case I think I'd be OK framing this as just an improvement rather than a break - I don't think it really necessary to jump straight to 4.X for this change, as I think it's reasonable to ask people to keep their local copy of yourkit up to date.

iamdanfox:
:+1:

PR #462, opened 2019-03-14
[fix] Fix gradle 4.10

dansanduleac:
## Before this PR

Was breaking on gradle 4.10:

```
> Invalid receiver type interface org.gradle.api.artifacts.result.ResolvedComponentResult; not a subtype of implementation type interface org.gradle.api.artifacts.result.ComponentResult
```

## After this PR

Works in gradle 4.10 (probably).

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
@@ -58,7 +58,6 @@ import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.component.ComponentIdentifier;
 import org.gradle.api.artifacts.component.ProjectComponentIdentifier;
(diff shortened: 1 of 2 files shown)


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
```diff
@@ -135,7 +134,8 @@ final void setManifestExtensions(Map<String, Object> manifestExtensions) {
     final Set<String> getProductDependenciesConfig() {
         // HACKHACK serializable way of representing all dependencies
         return productDependenciesConfig.get().getIncoming().getResolutionResult().getAllComponents().stream()
-                .map(ComponentResult::getId)
+                // intentionally using a lambda as otherwise we break gradle 4.10 support
+                .map(result -> result.getId())
```
iamdanfox, line 138: Have filed https://github.com/iamdanfox/gradle-clippy/issues/3 so we could eventually build a productized solution

iamdanfox:
:+1:

PR #451, opened 2019-03-06
[improvement] createManifest no longer requires compiling project source code

ferozco:
## Before this PR
To discover in repo product dependencies, `createManifest` would cause all project dependencies to run `Jar` causing them to fully compile.

## After this PR
Project dependencies are no longer compiled

PR #488, opened 2019-04-05
[fix] com.palantir.recommended-product-dependencies is now usable with gradle-consistent-versions

dansanduleac:
## Before this PR

`com.palantir.recommended-product-dependencies` eagerly reads the product dependency versions in afterEvaluate. This breaks with gradle-consistent-versions which no longer allows you to do this sort of thing.

## After this PR
==COMMIT_MSG==
`com.palantir.recommended-product-dependencies` loads product dependencies from other subprojects more lazily, while still not requiring that every other subproject be compiled. This keeps the benefit of #451 while allowing the actual configuration to happen at task execution time.
==COMMIT_MSG==


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/RecommendedProductDependenciesExtension.java
```diff
@@ -47,6 +49,17 @@ public final void productDependency(@DelegatesTo(ProductDependency.class) Closur
         }));
     }
 
+    final Provider<Set<ProductDependency>> getFinalizedRecommendedProductDependencies() {
+        if (GradleVersion.current().compareTo(GradleVersion.version("5.0")) >= 0) {
+            return recommendedProductDependencies.map(rpd -> {
+                recommendedProductDependencies.finalizeValue();
```
dansanduleac: this stuff is actually not necessary anymore since we now guarantee that we don't read this until execution time.

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/ConfigureProductDependenciesTask.java
```diff
@@ -0,0 +1,81 @@
...
+import org.gradle.api.Project;
+import org.gradle.api.java.archives.Manifest;
+import org.gradle.api.plugins.JavaPlugin;
+import org.gradle.api.plugins.JavaPluginConvention;
+import org.gradle.api.provider.Provider;
+import org.gradle.api.provider.SetProperty;
+import org.gradle.api.tasks.TaskAction;
+import org.gradle.api.tasks.bundling.Jar;
+
+public class ConfigureProductDependenciesTask extends DefaultTask {
```
iamdanfox, line 40: Can we have a bit of javadoc that explains why this hack exists - we wouldn't have to do this if there was a way of setting a Jar task's Manifest using a Provider? (At some point in the future this API might exists and I'd like to make sure it's clear when we can clean things up!)

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
```diff
@@ -601,8 +601,9 @@ class CreateManifestTaskIntegrationSpec extends GradleIntegrationSpec {
 
         then:
         result.task(":foo-server:createManifest").outcome == TaskOutcome.SUCCESS
+        result.task(":foo-api:configureProductDependencies").outcome == TaskOutcome.SUCCESS
         result.task(':foo-api:jar') == null
-        result.tasks.size() == 1
+        result.tasks.size() == 2
```
iamdanfox, line 606: Given that we're pushing for all internal projects to use gradle-consistent-versions, what do you think about having one integration tests that actually uses GCV?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
```diff
@@ -96,6 +99,36 @@
...
+
+                // Depend on the ConfigureProductDependenciesTask, if it exists, which will wire up the jar manifest
+                // with recommended product dependencies.
+                if (id instanceof ProjectComponentIdentifier) {
+                    Project dependencyProject =
+                            getProject().getRootProject().project(((ProjectComponentIdentifier) id).getProjectPath());
+                    return Stream.of(dependencyProject.getTasks().withType(ConfigureProductDependenciesTask.class));
+                }
+                return Stream.empty();
+            }).forEach(emptyFileCollection::builtBy);
```
iamdanfox, line 127: this is a pretty neat trick

dansanduleac: yeah, I wish this was less obscure, but it's the only way I know how to do it

iamdanfox (approved):
Just CR comments pls, then let's merge!

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
dansanduleac: wrote an integration test ensure that recommended-product-dependencies can get a version out of consistent-versions' getVersion

PR #494, opened 2019-04-12
[wip] [improvement] Create sls dist publication that captures product dependencies in the POM

dansanduleac:
## Before this PR

This plugin was not controlling the way we publish dists via the three plugins (java-service, asset, pod).
We only created a configuration called `sls` and declared an outgoing artifact on it, then relied on an internal plugin to wrap that into a maven publication and figure out where to publish it.

## After this PR
==COMMIT_MSG==
If running on gradle 5.3 or higher, a maven publication called `slsDist` will be created from any of the three dist plugins, that wires in the dist artifact (from the `sls` configuation) and additionally adds the discovered product dependencies as maven runtime dependencies.

The motivation here is to automate figuring out the full set of services (and their versions) you need to start up in your integration tests, given a set of products you know you need.
Once these dependencies are written to the POM, we can leverage gradle's dependency resolution in order to find a full set of product dependencies that are transitively necessary to actually run the requested products.

If running on gradle < 5.3, no changes in behaviour.
==COMMIT_MSG==

Next steps:
our internal publishing plugin needs to start picking up this `slsDist` publication if it exists, otherwise fall back to creating the publication as it currently does.

Remaining work:
- [ ] tests
- [ ] `com.palantir.sls-java-service-distribution` has no integration tests???
- [ ] (optionally) figure out running all tests against older gradle

## Potential downsides

A our circle templates / other gradle plugins might expect the publication to be called `dist` (by directly referencing a publish task derived from that name). Need to investigate if this coupling exists.

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyIntrospectionPlugin.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyIntrospectionPlugin.java
@@ -35,7 +35,7 @@ import org.gradle.util.GFileUtils;
 
 public final class ProductDependencyIntrospectionPlugin implements Plugin<Project> {
     private static final Logger log = Logging.getLogger(ProductDependencyIntrospectionPlugin.class);
-    private static final String PRODUCT_DEPENDENCIES_CONFIGURATION = "productDependencies";
+    static final String PRODUCT_DEPENDENCIES_CONFIGURATION = "productDependencies";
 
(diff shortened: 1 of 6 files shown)


iamdanfox:
So this is yet another way of capturing product dependencies?

dansanduleac:
This is a way of capturing _transitive_ product dependencies, so that you don't have to guess how the minimum version of a transitive changes as your services minimum versions get bumped. It's specifically targetting easy resolution of the complete set of products, for integration tests.

PR #500, opened 2019-04-24
[fix] Use pom extension for product dependencies

pkoenig10:
## Before this PR
Resolving configurations would download potentially large distribution artifacts for product dependencies.

## After this PR
Resolving configurations only downloads POMs for for product dependencies.

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyIntrospectionPlugin.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyIntrospectionPlugin.java
@@ -112,7 +112,8 @@ public final class ProductDependencyIntrospectionPlugin implements Plugin<Projec
                 .map(dependency -> project.getDependencies().create(ImmutableMap.of(
                         "group", dependency.getProductGroup(),
                         "name", dependency.getProductName(),
-                        "version", dependency.getMinimumVersion())))
+                        "version", dependency.getMinimumVersion(),
+                        "ext", "pom")))
                 .collect(Collectors.toList());
     }
 }


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyIntrospectionPlugin.java
```diff
@@ -112,7 +112,8 @@ private static String getMinimumProductVersion(Project project, String group, St
                 .map(dependency -> project.getDependencies().create(ImmutableMap.of(
                         "group", dependency.getProductGroup(),
                         "name", dependency.getProductName(),
-                        "version", dependency.getMinimumVersion())))
+                        "version", dependency.getMinimumVersion(),
+                        "ext", "pom")))
```
pkoenig10, line 116: These keys are defined in https://github.com/gradle/gradle/blob/v5.4.0/subprojects/dependency-management/src/main/java/org/gradle/api/internal/notations/DependencyMapNotationConverter.java#L45

pkoenig10:
This may cause problems with gofigure, which automatically extends the docker configuration and will now have the wrong artifact. I would prefer to merge this change and upgrade gofigure.

iamdanfox (approved):
Thanks @pkoenig10!

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyIntrospectionPlugin.java
ferozco: I don't think this works in all cases and doesn't leave room for an escape hatch. For example, most published assets do not produce a POM and therefore can not be resolved in this manner

dansanduleac:
This change breaks inter-project dependencies..
@pkoenig10 the workflow you raised is not a problem, because gofigure just extracts the version out of the resolved components. It doesn't rely on the `gofigure` configuration to download tgzs.

PR #501, opened 2019-04-25
Fix YourKit download URL

pkoenig10:
Recent CI builds have been failing because this URL now returns a 404. This is updates the version to use the download link currently posted on the YourKit website.


ferozco:
👍

PR #505, opened 2019-05-06
[fix] Explicitly set newRatio=2

ferozco:
## Before this PR
Our "Response" GC profile used CMS and did not explicitly set `NewRatio`, assuming that the default value of 2 would be respected. It turns out that CMS "ergonomically" selects a smaller new gen size to keep GC https://bugs.openjdk.java.net/browse/JDK-8153578

## After this PR
We explicitly set `NewRatio` to 2, increasing the size of the newGen and hopefully improving time spent in GC

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
@@ -42,6 +42,7 @@ public interface GcProfile extends Serializable {
     }
 
     class ResponseTime implements GcProfile {
+        private int initialNewRatio = 2;
         private int initiatingOccupancyFraction = 68;
 
         @Override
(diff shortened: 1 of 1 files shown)


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
```diff
@@ -48,6 +48,12 @@
         public final List<String> gcJvmOpts() {
             return ImmutableList.of("-XX:+UseParNewGC",
                     "-XX:+UseConcMarkSweepGC",
+                    /*
+                     * When setting UseConcMarkSweepGC the default value of NewRatio (2) is completely ignored.
+                     *
+                     * https://bugs.openjdk.java.net/browse/JDK-8153578
+                     */
+                    "-XX:NewRatio=2",
```
robert3005: Wanna add a setter the same way as `initiatingOccupancyFraction` so it's possible to modify this

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
```diff
@@ -67,6 +68,10 @@
         public final void initiatingOccupancyFraction(int occupancyFraction) {
             this.initiatingOccupancyFraction = occupancyFraction;
         }
+
+        public final void initialNewRatio(int newRatio) {
```
schlosna: nit `newRatio` as this doesn’t change

schlosna (approved):

dansanduleac:
👍

PR #511, opened 2019-05-09
Re-enable tests

iamdanfox:
Reverts palantir/sls-packaging#508


iamdanfox:
Slightly better idea would be to disable this only on CI, so that local runs should keep it working?

dansanduleac:
already done in #849

PR #513, opened 2019-05-10
Revert "[fix] Use pom extension for product dependencies (#500)"

pkoenig10:
Product dependencies may include other subprojects and explicitly depending on the `pom` artifact won't resolve to the project.


dansanduleac:
👍

PR #519, opened 2019-05-21
[improvement] Add `-XX:-OmitStackTraceInFastThrow` to default jvm opts

bkrieger:
## Before this PR
The JVM will have its default behavior, where if an exception is thrown many times, eventually the stacktrace will be omitted. This can make debugging difficult at times when there is a bug that causes an exception to be thrown over and over, but it isn't noticed until the old logs with the stacktrace have been rolled.

See http://jawspeak.com/2010/05/26/hotspot-caused-exceptions-to-lose-their-stack-traces-in-production-and-the-fix/
## After this PR
Stacktraces will always be present. While this disables a performance optimization, I can't think of a situation where we want these stacktraces to be optimized away. In general we should fix our code so that we don't throw the same exception over and over, rendering the optimization unnecessary.


schlosna:
This was discussed previously in #264 . I don’t think we want this enabled by default, as it  indicates you’ve got an expensive exception on a hot path. There should be full stacktrace exceptions from before the JIT kicks in and optimizes them away, so I’d suggest searching log archives for the original exceptions.

bkrieger:
That's fair, reading #264 I see the negative consequences.

PR #531, opened 2019-06-26
[improvement] Support resolving the sls publication via variant-aware resolution

dansanduleac:
## Before this PR

On a project that defines an sls dist (by applying `com.palantir.sls-java-service-distribution` etc), the presence of variants (consumable configurations with attributes), which are added by plugins such as gradle consistent versions, breaks dependencies from other projects.

## After this PR

Make the `sls` configuration a variant (by enriching it with a Usage attribute with the bespoke value of `sls-dist`).
Configure it such that consumer configurations that don't declare any required Usage attribute (e.g. gradle-docker's `docker` configuration) still manage to resolve the sls dist when depending on the sls project normally (e.g. `docker project(':my-dist')`).

==COMMIT_MSG==
Support resolving the sls publication from other projects via variant-aware resolution.
==COMMIT_MSG==

## Potential downsides

* Makes projects go through a separate code path when resolving dependencies onto sls projects. However, GCV already caused this to happen on all projects where it is applied, and the new code path is what gradle recommends using anyway (as opposed to the legacy one that was being used before)


CRogers (commented):
The tests look good to me, I'm not sure I know enough about the gradle variants etc to give good feedback on that.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/asset/AssetDistributionPluginIntegrationSpec.groovy
```diff
@@ -151,6 +151,52 @@ class AssetDistributionPluginIntegrationSpec extends GradleIntegrationSpec {
...
+            version '0.0.1'
+            distribution {
+                serviceName "my-asset"
+            }
+        ''')
+
+        runTasks('--write-locks')
+
+        expect:
+        runTasks(':verify')
```
ferozco, line 197: runTasksSuccessfully?

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/ServiceDistributionPluginTests.groovy
```diff
@@ -627,6 +627,103 @@ class ServiceDistributionPluginTests extends GradleIntegrationSpec {
         new File(childProject,'build/exploded/my-service-0.0.1/deployment/manifest.yml').exists()
     }
 
+    def 'exposes an artifact via dependency with sls-dist usage'() {
```
ferozco, line 641: restructure into `producer` and `consumer` subprojects?

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/asset/AssetDistributionPluginIntegrationSpec.groovy
dansanduleac: that's the default behaviour

ferozco:
👍

PR #543, opened 2019-07-11
Correctly merge product dependecies

ferozco:
## Before this PR
If multiple Jar's contributed a constraint on a given coordinate we would not merge them and instead use the first constraint we encountered

## After this PR
==COMMIT_MSG==
Correctly merge discovered product dependencies
==COMMIT_MSG==

## Possible downsides?
N/A


dansanduleac:
Went from faux merge to fo/merge!
👍

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/CreateManifestTask.java
```diff
@@ -427,7 +427,7 @@ private void ensureLockfileIsUpToDate(List<ProductDependency> productDeps) {
                 .forEach(productDependency -> discoveredProductDependencies.merge(
                         new ProductId(productDependency.getProductGroup(), productDependency.getProductName()),
                         productDependency,
-                        (key, oldValue) -> ProductDependencyMerger.merge(oldValue, productDependency)));
```
iamdanfox, line 430: oof that's subtle.  I wonder if error-prone should have told us `key` was unused and complained?

PR #580, opened 2019-09-02
Support gradle 6.0

ferozco:
## Before this PR
We would fail to build a Java service using gradle 6.0 because of a couple deprecations.

## After this PR
==COMMIT_MSG==
Raise minimum compatible version of Gradle to 5.1 to become compatible with  Gradle 6.0
==COMMIT_MSG==

## Possible downsides?
We will only be compatible with Gradle 5.1+ which gives a ~9 month compatibility window

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/ConfigTarTask.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/ConfigTarTask.java
@@ -37,8 +37,8 @@ public final class ConfigTarTask {
 
             task.from(new File(project.getProjectDir(), "deployment"));
             task.from(new File(project.getBuildDir(), "deployment"));
-            task.getDestinationDirectory().set(new File(project.getBuildDir(), "distributions"));
-            task.getArchiveBaseName().set(ext.getDistributionServiceName().get());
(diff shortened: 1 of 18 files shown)


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/tasks/ConfigTarTask.java
```diff
@@ -37,23 +37,21 @@ private ConfigTarTask() {}
...
                     return productTypeString.substring(1, productTypeString.lastIndexOf('.')).concat(".config.tgz");
                 } catch (IOException e) {
                     throw new RuntimeException(e);
                 }
-            }).get());
+            }));
+        });
 
-            // TODO(forozco): make this lazy since into does not support providers, but does support callable
+        // TODO(forozco): make this lazy since into does not support providers, but does support callable
```
robert3005, line 53: maybe we could fix this as well?

iamdanfox:
Out of curiosity, what were the deprecations?

ferozco:
@iamdanfox we mainly had to start using the provider based API for the tar task. There was also a runtime failure when attempting to create a property of `Map` instead of using a Map property

iamdanfox (approved):

PR #644, opened 2019-10-03
Automatically detect mainClass

ferozco:
## Before this PR
We required users to provide the main className in the service extension.

## After this PR
==COMMIT_MSG==
Services no longer need to specify the mainClassName, instead it will be inferred from the main source set
==COMMIT_MSG==

## Possible downsides?
N/A


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LazyCreateStartScriptTask.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Nullable
+    @Override
+    public final String getMainClassName() {
+        return getLazyMainClassName().get();
+    }
+
+    @Override
+    public final void setMainClassName(@Nullable String mainClassName) {
+        throw new UnsupportedOperationException("Use lazy property instead");
```
iamdanfox, line 38: surely this should just delegate to that `mainClassName.set` ? That way if people had multiple `public static void main(` for some reason, they still have an avenue to override one?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/util/MainClassResolver.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    private static Optional<String> findMainClass(SourceSet sourceSet) {
+        return sourceSet.getAllJava().getFiles().stream()
+                .flatMap(file -> {
+                    String contents = GFileUtils.readFile(file);
+                    if (contents.contains("public static void main(")) {
+                        Matcher matcher = PACKAGE_PATTERN.matcher(contents);
+                        if (matcher.find()) {
+                            return Stream.of(matcher.group(1) + "." + Files.getNameWithoutExtension(file.getName()));
```
iamdanfox: Why are we regexing here? Doesn't the `file` variable that we're flatMapping contain all the path info we need?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/util/MainClassResolver.java
```diff
@@ -0,0 +1,57 @@
...
+
+public final class MainClassResolver {
+    private static final Pattern PACKAGE_PATTERN = Pattern.compile("^package ([^\n;]+)");
+
+    public static String resolveMainClass(Project project) {
+        SourceSet main = project.getConvention().getPlugin(JavaPluginConvention.class)
+                .getSourceSets()
+                .getByName("main");
+        return findMainClass(main).orElseThrow(() -> new GradleException(
+                "Failed to infer main class, please ensure a main class exists"));
```
iamdanfox: `please ensure you have a method with the signature 'public static void main(String[] args) {}'`

iamdanfox:
I really like this idea!! One more line of gradle bites the dust.  I think @CRogers wanted to do a similar thing in gradle-atlas!

ferozco:
@CRogers was the one with the original idea and it has already been merged and released in gradle-atlas

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LazyCreateStartScriptTask.java
ferozco: if they wanted to override it, they should do so using the extension not by manipulating the task

iamdanfox: Ok fair enough

iamdanfox (approved):

iamdanfox:
While we're on the topic, could we make the default service name just be the project name? So many repos start with:

```gradle

distribution {
    serviceName project.name // <--- always the same!!!
    mainClass 'com.palantir.preview.PreviewServer'
```

ferozco:
Looks like we already do that actually haha

iamdanfox:
🤦‍♂

markelliot:
could you add a test to cover what happens if there are multiple main classes?

ferozco:
@markelliot I've added a test here: https://github.com/palantir/sls-packaging/pull/652

PR #649, opened 2019-10-08
Actually configure run task correctly

ferozco:
## Before this PR
#647 correctly made run task configuration lazy, but it sourced the mainClass directly from the extension instead of falling back to the discovered main class

## After this PR
==COMMIT_MSG==
Run task correctly uses the discovered mainClass
==COMMIT_MSG==

## Possible downsides?
No, just a bug fix


qinfchen (approved):

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
```diff
@@ -207,7 +207,7 @@ public void apply(Project project) {
             task.doFirst(new Action<Task>() {
                 @Override
                 public void execute(Task _task) {
-                    task.setMain(distributionExtension.getMainClass().get());
+                    task.setMain(mainClassName.get());
```
qinfchen, line 210: maybe adding a test?

CRogers (approved):

PR #652, opened 2019-10-12
Demonstrate main class inference failure

ferozco:
## Before this PR
We did not demonstrate what would happen if multiple main classes were discovered

## After this PR
Added a test showing that we fail we a precise error message when multiple main classes are discovered

## Possible downsides?
N/A


markelliot:
Thanks! Should we allow for specifying which main class when there’s more than one present? I imagine some servers may still provide CLI main classes, and we probably shouldn’t outright prevent those from building.

ferozco:
Users can actually still specify the main class like used to. The main class inference just attempts to provide a default if non is specified. I've also updated the README to make that more clear

markelliot:
Makes sense. Could we include test coverage for the multiple-classes but user specifies which one case, too, please?

ferozco:
@markelliot I've added another test. Is there anything else you'd like?

markelliot (approved):

PR #664, opened 2019-10-30
Flip response-time-11 to use ZGC

j-baker:

### DIFF
--- a/changelog/@unreleased/pr-664.v2.yml
+++ b/changelog/@unreleased/pr-664.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Flip response-time-11 to use ZGC
+  links:
+  - https://github.com/palantir/sls-packaging/pull/664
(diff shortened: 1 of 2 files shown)


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
```diff
@@ -87,8 +88,9 @@ public final void newRatio(int newerRatio) {
         @Override
         public final List<String> gcJvmOpts() {
             return ImmutableList.of(
-                    // https://wiki.openjdk.java.net/display/shenandoah/Main
-                    "-XX:+UseShenandoahGC",
+                    // https://wiki.openjdk.java.net/display/zgc/Main
+                    "-XX:+UnlockExperimentalVMOptions",
+                    "-XX:+UseZGC",
```
CRogers, line 93: Why is this an improvement?

carterkozak:
I don't believe zgc supports compressed oops, so heaps lower than 31.5g will effectively shrink.
@schlosna might have feedback on this change.

changelog/@unreleased/pr-664.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Flip response-time-11 to use ZGC
```
carterkozak, line 3: If this goes through, we should note that this adds a new `response-time-11` gc profile.

robert3005: We used to have this profile but removed it as it was relying on redhats jdk builds with shenandoahgc. The only confusing thing about ZGC is that it's linux only

carterkozak: Can we put that in the changelog?

PR #678, opened 2019-11-11
Run task inherits GC options

ferozco:
## Before this PR
The GC profile was ignored when running locally

## After this PR
==COMMIT_MSG==
Service Distributions "run" task correctly respects the GC profile
==COMMIT_MSG==

## Possible downsides?
N/A


carterkozak:
:+1:

Issue #533, opened 2019-06-26
Use 'hybrid' as default with java 11 and shenandoah with java 12


robert3005:
## What happened?

I got defaulted to parallelold collector even though I was using java 11

## What did you want to happen?

Default would be G1GC for 11 and shenandoah for 12

PR #679, opened 2019-11-11
Expose shenandoah as response-time-13

ferozco:
## Before this PR
There was a `ResponseTime11` GC Profile in code that was not easily exposed to users that would ostensibly configure the JVM (with Java >= 11) to use Shenandoah. 

Unfortunately, "hacking" to use this profile with Java 11 would cause a process to crash with `Unrecognized VM option 'UseShenandoahGC'`. 

When run with Java 13 you'd the JVM would fail to start with the following error `Error: VM option 'UseShenandoahGC' is experimental and must be enabled via -XX:+UnlockExperimentalVMOptions.`

## After this PR
==COMMIT_MSG==
Expose `response-time-13` GC profile 
==COMMIT_MSG==

## Possible downsides?
Technically this is break since users _could_ have been using the GC profile and then made it work by passing the missing JVM flag separately. I don't think thats the case though, so I feel pretty comfortable marking this as an improvement

### DIFF
--- a/changelog/@unreleased/pr-679.v2.yml
+++ b/changelog/@unreleased/pr-679.v2.yml
@@ -1,0 +1,5 @@
+type: feature
+feature:
+  description: Expose `response-time-13` GC profile
+  links:
+  - https://github.com/palantir/sls-packaging/pull/679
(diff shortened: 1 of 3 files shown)


robert3005:
What do you think about taking into account the target jvm version and changing behaviour of profiles? Something along the lines of https://github.com/palantir/sls-packaging/issues/533 ?

I have not implemented it only because it requires plumbing this additional version through but that would let us customize things in the future more easily

ferozco:
I'd be supportive of adding a target Java version to the `JavaServiceDistributionExtension` so that we can make better decision on how to configure a JVM for users

ferozco:
Closing in favour of https://github.com/palantir/sls-packaging/pull/680

PR #680, opened 2019-11-11
Expose javaVersion property in ServiceDistributionExtension

### LINKED ISSUE #533: Use 'hybrid' as default with java 11 and shenandoah with java 12

ferozco:
Closes #533 

## Before this PR
Users had no way of telling us what version of Java they expect to run with, the best they could do was implicitly specify the version using `JAVA_HOME`.

This made tuning GC and other JVM options hard since we had no way of telling what we options we could safely use.

## After this PR
==COMMIT_MSG==
Add `javaVersion` property to JavaServiceDistributionExtension which specifies the version of Java the service intends to run with. javaVersion defaults to the source compatibility of the project
==COMMIT_MSG==

## Possible downsides?
It could be a little confusing that users have to specify a different javaVersion in addition to JAVA_HOME to change which version of Java they use. We _could_ specify a convention where the default java home is `JAVA_${javaVersion}_HOME` so that users (at least internally) don't have to think about it.

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
@@ -22,6 +22,7 @@ import com.google.common.collect.ImmutableMap;
 import java.io.Serializable;
 import java.util.List;
 import java.util.Map;
+import org.gradle.api.JavaVersion;
 
 public interface GcProfile extends Serializable {
     long serialVersionUID = 1L;
@@ -32,11 +33,11 @@ public interface GcProfile extends Serializable {
             "response-time", GcProfile.ResponseTime.class,
             "hybrid", GcProfile.Hybrid.class);
 
-    List<String> gcJvmOpts();
+    List<String> gcJvmOpts(JavaVersion javaVersion);
 
     class Throughput implements GcProfile {
         @Override
-        public final List<String> gcJvmOpts() {
+        public final List<String> gcJvmOpts(JavaVersion javaVersion) {
             return ImmutableList.of("-XX:+UseParallelOldGC");
         }
     }
@@ -46,7 +47,17 @@ public interface GcProfile extends Serializable {
         private int initiatingOccupancyFraction = 68;
 
         @Override
-        public final List<String> gcJvmOpts() {
+        public final List<String> gcJvmOpts(JavaVersion javaVersion) {
+            // HACKHACK constant for Java13 only exists in Gradle 6.0
+            if (javaVersion == JavaVersion.VERSION_HIGHER) {
+                return ImmutableList.of(
+                        "-XX:+UnlockExperimentalVMOptions",
+                        // https://wiki.openjdk.java.net/display/shenandoah/Main
+                        "-XX:+UseShenandoahGC",
+                        // "forces concurrent cycle instead of Full GC on System.gc()"
+                        "-XX:+ExplicitGCInvokesConcurrent",
+                        "-XX:+ClassUnloadingWithConcurrentMark");
+            }
             return ImmutableList.of("-XX:+UseParNewGC",
                     "-XX:+UseConcMarkSweepGC",
                     /*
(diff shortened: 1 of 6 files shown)


iamdanfox:
Ok so presently an internal service that is following our recommended best practises has a snippet like the following and we're gonna ask them to add one more javaVersion:

```diff
 distribution {
     serviceName project.name
     manifestExtensions 'upgrade-strategy': 'rolling'
     manifestExtensions 'public-proxy-endpoints': ['service-endpoint']
 
     javaHome '$JAVA_11_HOME'
+    javaVersion JavaVersion.VERSION_11
 }
 
 slsDocker {
     port 7825
     javaVersion JavaVersion.VERSION_11
 }
```

I guess we could potentially make slsDocker detect this automatically if we're ok with introducing another plugin dependency?

Perhaps `javaVersion` should actually be `runtimeJavaVersion` or `runtimeCompatibility` or something to minimize possible confusion with gradle's baked-in sourceCompatibility and targetCompatibility?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
```diff
@@ -46,7 +47,18 @@
...
+            if (javaVersion == JavaVersion.VERSION_HIGHER) {
+                return ImmutableList.of(
+                        "-XX:+UnlockExperimentalVMOptions",
+                        // https://wiki.openjdk.java.net/display/shenandoah/Main
+                        "-XX:+UseShenandoahGC",
+                        // "forces concurrent cycle instead of Full GC on System.gc()"
+                        "-XX:+ExplicitGCInvokesConcurrent",
+                        "-XX:+ClassUnloadingWithConcurrentMark",
+                        "-XX+UseNUMA");
+            }
```
iamdanfox, line 61: Might be worth calling out in the changelog that opting into `gc 'response-time'` now gets you shenandoah if you're intending to run on Java13+?

ferozco:
Because `javaVersion` defaults to the projects `sourceCompatability` most repos would not have to take any action. We could go a step further and derive `javaHome` using `javaVersion` so that the final impact to end users would be:


```diff
 // Somewhere else in build
 sourceCompatability 11

 distribution {
     serviceName project.name
     manifestExtensions 'upgrade-strategy': 'rolling'
     manifestExtensions 'public-proxy-endpoints': ['service-endpoint']
 
-     javaHome '$JAVA_11_HOME'
 }
```

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
```diff
@@ -46,7 +47,18 @@
...
+            // TODO(forozco): Update to use proper constant once we can support gradle 6.0
+            if (javaVersion == JavaVersion.VERSION_HIGHER) {
+                return ImmutableList.of(
+                        "-XX:+UnlockExperimentalVMOptions",
+                        // https://wiki.openjdk.java.net/display/shenandoah/Main
+                        "-XX:+UseShenandoahGC",
+                        // "forces concurrent cycle instead of Full GC on System.gc()"
+                        "-XX:+ExplicitGCInvokesConcurrent",
+                        "-XX:+ClassUnloadingWithConcurrentMark",
+                        "-XX+UseNUMA");
```
robert3005, line 60: this is on by default https://github.com/openjdk/jdk/blob/master/src/hotspot/share/gc/shenandoah/shenandoahArguments.cpp#L62

robert3005:
I think this lets us get rid of `JAVA_11_HOME` hack we had to expose everywhere and instead tie the runtime version back to your java settings. We should teach gradle-docker about this setting so that we don't have to duplicate it

robert3005:
However, I would leave all suggestions that remove code duplication as a follow up

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
ferozco: I'd prefer to explicitly set it given that the [shenandoah](https://wiki.openjdk.java.net/display/shenandoah/Main#Main-Basicconfiguration) docs encourage setting the flag and not having the flag set results in a performance loss

carterkozak:
For the follow-up, it would be nice to have a configuration point for the jvm flavor (zulu/openjdk/etc) separately from the version.

iamdanfox (approved):

PR #696, opened 2019-11-25
$JAVA_11_HOME for java 11

CRogers:
## Before this PR
Each repo internally has had to manually specify they're using `$JAVA_11_HOME` when using Java 11. `JAVA_11_HOME` is an environment given by our internal deployment infrastructure tool.

## After this PR
==COMMIT_MSG==
For Java 11 and above, the `javaHome` property on the `distribution` extension will be set to `$JAVA_11_HOME` etc by default.
==COMMIT_MSG==

## Possible downsides?
Is setting `$JAVA_11_HOME` tying this too much to the behaviour of our internal deployment infrastructure? Without this though every gradle java gradle project needs to manually specify this (and it is unlikely to ever change).

If someone was depending on the `javaHome` to be unset when using java 11, they now may be broken. Unknown if people are actually depending on this, as SL uses `JAVA_HOME` has java 8.

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java
@@ -61,11 +61,12 @@ public class JavaServiceDistributionExtension extends BaseDistributionExtension 
         mainClass = objectFactory.property(String.class);
 
         javaHome = objectFactory.property(String.class).value(javaVersion.map(javaVersionValue -> {
-            if (javaVersionValue.compareTo(JavaVersion.VERSION_11) < 0) {
+            boolean javaVersionLessThan11 = javaVersionValue.compareTo(JavaVersion.VERSION_11) < 0;
+            if (javaVersionLessThan11) {
                 return "";
             }
 
-            return "JAVA_" + javaVersionValue.getMajorVersion() + "_HOME";
+            return "$JAVA_" + javaVersionValue.getMajorVersion() + "_HOME";
         }));
 
         addJava8GcLogging = objectFactory.property(Boolean.class).value(false);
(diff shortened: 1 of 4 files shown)


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java
```diff
@@ -59,7 +59,15 @@ public JavaServiceDistributionExtension(Project project) {
         javaVersion = objectFactory.property(JavaVersion.class).value(project.provider(() ->
                 project.getConvention().getPlugin(JavaPluginConvention.class).getSourceCompatibility()));
         mainClass = objectFactory.property(String.class);
-        javaHome = objectFactory.property(String.class);
+
+        javaHome = objectFactory.property(String.class).value(javaVersion.map(javaVersionValue -> {
+            boolean javaVersionLessThan11 = javaVersionValue.compareTo(JavaVersion.VERSION_11) < 0;
+            if (javaVersionLessThan11) {
+                return "";
```
CRogers, line 66: Returning empty string here because that was what was happening in:
https://github.com/palantir/sls-packaging/blob/5266eb5bca9beb40c13988ff2ec74090f6b79aed/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java#L162

Ideally I'd make this provider somehow indicate it's empty, but I can't see how to do that and keep this lazy in Gradle.

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java
```diff
@@ -59,7 +59,15 @@ public JavaServiceDistributionExtension(Project project) {
         javaVersion = objectFactory.property(JavaVersion.class).value(project.provider(() ->
                 project.getConvention().getPlugin(JavaPluginConvention.class).getSourceCompatibility()));
```
iamdanfox: should this also be targetCompatibility just in case?

iamdanfox (approved):
Would kinda like to have a little README note about this contract/expectation of sls-packaging

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java
iamdanfox: our internal deployment daemon has the implicit assumption of Java 8 anywhere - rather than a < Java11 check, shall we do <= Java8?

PR #727, opened 2020-01-06
Fix innacurate error message

### LINKED ISSUE #695: Remove product dependency warning if higher version
## What happened?

Manually bumped a product dependency to pick up a semantic API change. The warning suggests the dependency is unnecessary but that isn't true:

```
Please remove your declared product dependency on '{product}' because it is already provided by a jar dependency:

        Provided:     {product}(min: 3.46.1, recommended: 3.46.1, max: 3.x.x)
        You declared: {product}(min: 3.109.0, recommended: null, max: 3.x.x)
Updated {product}/product-dependencies.lock
```

Warn only if manual dependency is _lower_ than recommended

ferozco:
Closes #695 

## Before this PR
If a user specified a product dependency that was also discovered they would always see an error message stating that the declared dependency was unnecessary.

## After this PR
==COMMIT_MSG==
Only warn users about unnecessary product dependencies when they are actually unnecessary
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/changelog/@unreleased/pr-727.v2.yml
+++ b/changelog/@unreleased/pr-727.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Only warn users about unnecessary product dependencies when they are
+    actually unnecessary
+  links:
+  - https://github.com/palantir/sls-packaging/pull/727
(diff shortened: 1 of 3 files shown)


dansanduleac (approved):
nice!

PR #749, opened 2020-01-27
yourkit b127 -> b137

iamdanfox:
## Before this PR

I can't run `publishToMavenLocal` anymore because the yourkit dist has disappeared.

## After this PR
==COMMIT_MSG==
sls-packaging now embeds libyjpagent.so from YourKit 2019.8 b137 in server distributions
==COMMIT_MSG==


carterkozak:
:+1:

PR #750, opened 2020-01-27
Allow opting-in to graal

iamdanfox:
## Before this PR

https://www.graalvm.org/docs/why-graal/

> The compiler of GraalVM provides performance advantages for highly abstracted programs due to its ability to remove costly object allocations in many scenarios. You can find details in this research paper and try an example. Better inlining and more aggressive speculative optimizations can lead to additional benefits for complex long-running applications, see a Stream API example.

As per https://github.com/oracle/graal/blob/master/compiler/README.md:

> JVMCI version is built into the JDK as of JDK 11.

## After this PR
==COMMIT_MSG==
distributions with `gc-profile: 'hybrid'` (aka G1) can now opt-in to try the Graal C2 compiler using `graal true` (on Java11+)
==COMMIT_MSG==

We already have the reasonably experimental shenandoah one in there, I think it's ok to have lightly used feature flags in here rather than collecting duplicate flags in gradle files.

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
@@ -35,6 +35,9 @@ public interface GcProfile extends Serializable {
 
     List<String> gcJvmOpts(JavaVersion javaVersion);
 
+    ImmutableList<String> GRAAL = ImmutableList.of(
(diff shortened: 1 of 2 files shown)


ferozco (commented):
could we split out compiler config from GC settings since you can set graal as your optimizing compiler regardless of GC

robert3005:
Can we split this out as something separate to gc settings? It’s just conditioned on java 11 being available and we have done the work of specifying jvm runtime version already

iamdanfox:
Closing out as gonna focus on getting java13 excavated out first before we tackle graal.

PR #753, opened 2020-01-28
Add support for arm64 architectures

markelliot:
## Before this PR
sls-packaging supports Linux x86_64 and MacOS x86_64.

## After this PR
==COMMIT_MSG==
Add support for Linux arm64
==COMMIT_MSG==

## Possible downsides?
Increased size of output tgz files due to additional native architecture support.


iamdanfox (approved):
LGTM, I think a webhook from CircleCI -> github got lost, so your tests are still showing as pending.  Pushing either an empty commit (or just power merging should do the trick!)

PR #800, opened 2020-02-21
Factor out `com.palantir.recommended-product-dependencies`

ferozco:
## Before this PR
It was difficult for other plugins to integrate with recommended product dependencies. In particular, gradle-conjure did some gross things to configure product dependencies.

## After this PR
==COMMIT_MSG==
Refactor `com.palantir.recommended-product-dependencies` into separate JAR
==COMMIT_MSG==

## Possible downsides?
N/A


gradle-recommended-product-dependencies/build.gradle
```diff
@@ -0,0 +1,18 @@
+apply plugin: 'groovy'
+apply plugin: 'java-gradle-plugin'
+apply plugin: 'org.inferred.processors'
+apply from: "${rootDir}/gradle/publish-jar.gradle"
```
iamdanfox, line 4: If we're expecting to compile against this jar, should we also include revapi to catch regressions?

iamdanfox (approved):

PR #844, opened 2020-03-12
Handle multiple in repo recommended pdeps sources

ferozco:
## Before this PR
Having many projects producing recommended product dependencies on the same coordinate could result in an inconsistent lockfile when one of the recommendations was the current project version and the project was currently dirty. We would do the wrong thing and ignore the dirty version in favour of the other recommended version which would then break at CI time when the repo was clean and we returned to respecting the project version.

## After this PR
==COMMIT_MSG==
Correctly handle multiple in repo recommended product dependencies when the constraint is the project version and the project version is not orderable
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyMerger.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyMerger.java
@@ -21,7 +21,6 @@ import com.google.common.collect.Streams;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import com.palantir.sls.versions.OrderableSlsVersion;
-import com.palantir.sls.versions.SlsVersion;
 import com.palantir.sls.versions.SlsVersionMatcher;
 import com.palantir.sls.versions.VersionComparator;
 import java.util.Objects;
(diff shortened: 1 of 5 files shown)


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyMerger.java
```diff
@@ -92,6 +81,32 @@ public static ProductDependency merge(ProductDependency dep1, ProductDependency
...
+    private static String getMinimumVersion(
+            ProductDependency dep1, ProductDependency dep2, Optional<String> defaultMinimumVersion) {
+        if (defaultMinimumVersion.isPresent()
+                && (!OrderableSlsVersion.check(dep1.getMinimumVersion())
+                        || !OrderableSlsVersion.check(dep2.getMinimumVersion()))) {
+            return defaultMinimumVersion.get();
+        }
+
+        // This could be empty if both of the versions are dirty
+        Optional<OrderableSlsVersion> minimumVersionOrderable = Stream.of(dep1.parseMinimum(), dep2.parseMinimum())
```
fawind: We can already return this version as the if-else checks are no longer necessary. If one of the versions is not parseable, we already return early in line `89`. Based on that, `minimumVersionOrderable` will never be empty.

fawind (commented):
Looks good!

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyMerger.java
ferozco: thats only true if default minimum is present, right?

dansanduleac (approved):

PR #849, opened 2020-03-24
Time to deal with these deprecations

dansanduleac:
## Before this PR

Ignoring deprecations causing other projects to have to ignore them too when testing against sls-packaging on gradle 6.

## After this PR
==COMMIT_MSG==
Stop ignoring deprecations so we know we're clean on gradle 6.

Also reverts #508 as I haven't seen the reported failure anymore.
==COMMIT_MSG==

PR #852, opened 2020-03-26
Fix regression: no longer require gradle 6

dansanduleac:
## Before this PR

#849 accidentally introduced code that only works in Gradle 6.

```
Caused by: org.gradle.internal.typeconversion.UnsupportedNotationException: Cannot convert the provided notation to a String: provider(?).
The following types/formats are supported:
  - String or CharSequence instances, for example 'some/path'.
  - Boolean values, for example true, Boolean.TRUE.
  - Number values, for example 42, 3.14.
  - A File instance
  - A Closure that returns any supported value.
  - A Callable that returns any supported value.
	at org.gradle.internal.typeconversion.ErrorHandlingNotationParser.parseNotation(ErrorHandlingNotationParser.java:57)
	at org.gradle.api.internal.file.copy.DefaultCopySpec$DefaultCopySpecResolver.getDestPath(DefaultCopySpec.java:554)
```

## After this PR
==COMMIT_MSG==
Minimum gradle is now 5.6, which is checked in the base plugin, and integration tests run against it.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-852.v2.yml
+++ b/changelog/@unreleased/pr-852.v2.yml
@@ -1,0 +1,7 @@
+type: fix
+fix:
+  description: Fix regression that required gradle 6, bringing minimum gradle down
+    to 5.6, which is checked in the base plugin, and integration tests run against
+    it.
+  links:
+  - https://github.com/palantir/sls-packaging/pull/852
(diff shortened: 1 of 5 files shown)


dansanduleac:
I don't understand how this failed but looks like a flake
```
Condition not satisfied:

execWithOutput('dist/service-name-0.0.1/service/bin/init.sh', 'check') ==~ /.*\n*Checking health of 'service-name'\.\.\.\s+Healthy.*\n/
|                                                                      |
|                                                                      false
Command is deprecated: the next major release of sls-packaging will only support start/status/stop
Checking health of 'service-name'...              Healthy

	at com.palantir.gradle.dist.service.JavaServiceDistributionPluginTests.produce distribution bundle and check start, stop, restart, check behavior(JavaServiceDistributionPluginTests.groovy:78)
```

iamdanfox (approved):

PR #868, opened 2020-04-09
Make DistTarTask public for broader usage

chrismale:
## Before this PR
Custom gradle plugins which build on top of the DistTarTask cannot access its configure method currently. This was the case some time ago when the class was written in groovy but now
the package private visibility is enforced by Java.

## After this PR
==COMMIT_MSG==
Make DistTarTask public for broader usage
==COMMIT_MSG==


ferozco:
We only hold buildscript visible extensions as API so I don't think we want to be exposing ways for people to depend sls-packaging as a libary. Why do you need access to the configuration of distTar?

chrismale:
We have a custom gradle plugin that wraps much of the configuration related to distTar and sets the values inside the plugin itself, rather than having those values explicitly set in the gradle files.

carterkozak:
Have y'all met? It might be helpful to jump into a call and chat for a moment.

PR #886, opened 2020-05-12
Support gradle 6.4

ferozco:
## Before this PR
`run` task would fail at configuration time with gradle >= 6.4 because of an implementation change in `JavaExec` which broke the way we lazily configured the main class. 

## After this PR
==COMMIT_MSG==
`run` works with gradle 6.4 and up
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
@@ -47,6 +47,7 @@ import org.gradle.api.tasks.bundling.Compression;
 import org.gradle.api.tasks.bundling.Jar;
 import org.gradle.api.tasks.bundling.Tar;
 import org.gradle.util.GFileUtils;
+import org.gradle.util.GradleVersion;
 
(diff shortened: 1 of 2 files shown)


dansanduleac (approved):

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
```diff
@@ -217,12 +218,16 @@ public void apply(Project project) {
...
-            });
+            if (GradleVersion.current().compareTo(GradleVersion.version("6.4")) < 0) {
+                task.doFirst(new Action<Task>() {
+                    @Override
+                    public void execute(Task _task) {
+                        task.setMain(mainClassName.get());
+                    }
+                });
+            } else {
+                task.getMainClass().set(mainClassName);
```
iamdanfox, line 229: would this just throw a NoSuchMethodError if we ran it on gradle 6.3?

ferozco: yup

iamdanfox:
Do we have any tests that run with various different `gradleVersion`s?

PR #899, opened 2020-05-26
response-time uses shenandoah from 14 onwards

iamdanfox:
## Before this PR

We have a ton of services running on Java11, and jumping them all the way to Java14 is going to require migrating them from CMS -> Shenandoah. This is reasonably scary, partly because Shenandoah takes up more heap, but also partly because CMS has been very carefully tuned and we've found that just flipping the switch on catalog was not a magic bullet.

## After this PR
As such, I'd like to get 11 -> 13 rolling as this seems pretty non-scary, while we investigate shenandoah on 14 in parallel.

==COMMIT_MSG==
For users of the `response-time` GC Profile, options to enable the CMS garbage will still be set on Java 13, as the GC is just deprecated but not full removed. From Java 14 onwards, CMS does not exist anymore so we auto-switch to Shenandoah.
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
@@ -52,7 +52,6 @@ public interface GcProfile extends Serializable {
             // use it up until this release.
             if (javaVersion.compareTo(JavaVersion.toVersion("14")) >= 0) {
                 return ImmutableList.of(
-                        "-XX:+UnlockExperimentalVMOptions",
                         // https://wiki.openjdk.java.net/display/shenandoah/Main
                         "-XX:+UseShenandoahGC",
                         // "forces concurrent cycle instead of Full GC on System.gc()"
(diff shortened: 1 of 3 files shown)


carterkozak (approved):

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
```diff
@@ -48,7 +48,9 @@
 
         @Override
         public final List<String> gcJvmOpts(JavaVersion javaVersion) {
-            if (javaVersion.compareTo(JavaVersion.toVersion("13")) >= 0) {
+            // The CMS garbage collector was removed in Java 14: https://openjdk.java.net/jeps/363. Users are free to
+            // use it up until this release.
+            if (javaVersion.compareTo(JavaVersion.toVersion("14")) >= 0) {
                 return ImmutableList.of(
                         "-XX:+UnlockExperimentalVMOptions",
```
ferozco: I don't think you need this flag anymore

robert3005:
@iamdanfox java 13 is eol

iamdanfox:
Do you think we can't actually use it as a stepping stone?

PR #907, opened 2020-05-29
Improve npe on jdk 14

ferozco:
## Before this PR
[Jep 358](https://openjdk.java.net/jeps/358) described an improved NPE message that would help devs debug the problem. https://bugs.openjdk.java.net/browse/JDK-8227717 added the implementation behind the flag `-XX:ShowCodeDetailsInExceptionMessages`.

## After this PR
==COMMIT_MSG==
Enable Improved NullPointerException messages on modern JDKs
==COMMIT_MSG==

We can separately decide whether we feel comfortable safe-logging the NPE message so that devs can easily take advantage of the information

## Possible downsides?
N/A

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
@@ -119,7 +119,7 @@ public class LaunchConfigTask extends DefaultTask {
     }
 
     @Input
-    public Property<JavaVersion> getJavaVersion() {
+    public final Property<JavaVersion> getJavaVersion() {
         return javaVersion;
     }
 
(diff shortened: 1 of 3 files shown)


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -165,6 +173,10 @@ public final void createConfig() throws IOException {
                         .classpath(relativizeToServiceLibDirectory(classpath))
                         .addAllJvmOpts(alwaysOnJvmOptions)
                         .addAllJvmOpts(addJava8GcLogging.get() ? java8gcLoggingOptions : ImmutableList.of())
+                        .addAllJvmOpts(
+                                javaVersion.get().compareTo(JavaVersion.toVersion("14")) >= 0
```
robert3005, line 177: this is fine since we only have one use case so far but long term this should turn into a lookup table

robert3005 (approved):

PR #912, opened 2020-06-04
Fix: Upgrade yourkit to b139 to fix broken develop

robert3005:
## Before this PR
You can't run build on develop

## After this PR
==COMMIT_MSG==
Upgrade yourkit agent version to avoid broken develop
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/gradle-sls-packaging/build.gradle
+++ b/gradle-sls-packaging/build.gradle
@@ -58,7 +58,7 @@ task downloadYourkitDist(type: Download) {
 task verifyYourkitDist(type: Verify, dependsOn: downloadYourkitDist) {
     src file("${buildDir}/${yourkitFilename}.zip")
     algorithm 'SHA-256'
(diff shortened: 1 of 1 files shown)


gradle-sls-packaging/build.gradle
```diff
@@ -45,7 +45,7 @@ pluginBundle {
     }
 }
 
-def yourkitVersion = "2019.8-b137"
+def yourkitVersion = "2019.8-b139"
```
ferozco, line 48: you have to update the sha :(

ferozco:
We've been periodically making these manual PRs for a while now. It would be really great if we could automate this

gradle-sls-packaging/build.gradle
robert3005: just did

robert3005:
tbh I think we should rely on jfr instead of yourkit agent. There's a lot more infrastructure around that than yourkit

ferozco:
👍

PR #916, opened 2020-06-08
Expose product dependency list property

ferozco:
## Before this PR
We only exposed a Provider for reading a distributions product dependencies

## After this PR
==COMMIT_MSG==
Expose ListProperty for accessing and configuring a distribution's product dependencies
==COMMIT_MSG==


iamdanfox (approved):
Cool, lets us do much nicer laziness in other plugins

PR #921, opened 2020-06-10
Fix: CreateStartScripts works with gradle 6.4+

ferozco:
## Before this PR

@fmance had his start script broken on windows

## After this PR
==COMMIT_MSG==
Correctly configure start scripts in Gradle 6.4+
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/LazyCreateStartScriptTaskIntegrationSpec.groovy
+++ b/gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/LazyCreateStartScriptTaskIntegrationSpec.groovy
@@ -1,0 +1,38 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/LazyCreateStartScriptTaskIntegrationSpec.groovy
```diff
@@ -0,0 +1,38 @@
...
+            applicationName = 'test-service'
+            lazyMainClassName = 'myMainClass'
+            outputDir = file('build/scripts')
+        }
+        """.stripIndent()
+        def successfully = runTasksSuccessfully('createStartScripts')
+
+        then:
+        file('build/scripts/test-service').text.contains 'myMainClass'
+        file('build/scripts/test-service.bat').text.contains 'myMainClass'
```
iamdanfox, line 36: rather than checking a single line, could we just check in the expected `test-service.bat` and verify it's the same? (with the whole `-Drecreate=true` thing)?

ferozco: I don't really want to test the entire output of createStartScripts since its a gradle task that we configure/extend. I only want to test the logic which we added around lazily configuring the main class

iamdanfox: I dunno, it actually seems kinda interesting if we got a gradle upgrade and this changed significantly?

ferozco: well the issue is that we were hacking around the fact that the gradle task wasn't lazily configurable so when they did make the task lazily configurable our hack broke

robert3005:
Please add a ==COMMIT_MSG== or remove it's hidden html contents

dansanduleac (approved):

PR #924, opened 2020-06-11
Improvement: Use G1GC as default for java 11 in throughput mode

robert3005:
## Before this PR
We would force new jvms to ParallelOldGC even though java 9 makes G1 the default

## After this PR
==COMMIT_MSG==
Set G1GC as default on java 11+ in throughput mode
==COMMIT_MSG==

## Possible downsides?
This will change people's gc's just because they're running new version but we find it acceptable for response-time already

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
@@ -38,6 +38,9 @@ public interface GcProfile extends Serializable {
     class Throughput implements GcProfile {
         @Override
         public final List<String> gcJvmOpts(JavaVersion javaVersion) {
+            if (javaVersion.compareTo(JavaVersion.toVersion("11")) >= 0) {
+                return ImmutableList.of("-XX:+UseG1GC", "-XX:+UseStringDeduplication");
+            }
             return ImmutableList.of("-XX:+UseParallelOldGC");
         }
     }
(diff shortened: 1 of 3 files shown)


carterkozak (changes requested):
I don't understand the rationale. When the throughput garbage collection profile is selected, shouldn't we provide collector that's optimized for throughput?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
```diff
@@ -38,6 +38,9 @@
     class Throughput implements GcProfile {
         @Override
         public final List<String> gcJvmOpts(JavaVersion javaVersion) {
+            if (javaVersion.compareTo(JavaVersion.toVersion("11")) >= 0) {
+                return ImmutableList.of("-XX:+UseG1GC", "-XX:+UseStringDeduplication");
+            }
```
schlosna, line 43: For throughput on JDK11+ I think we still want `-XX:+UseParallelOldGC`.

carterkozak:
Perhaps the default when no value is defined should be `hybrid` to support a wider variety of usage patterns out of the box?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
schlosna: I could see potentially using G1GC & enabling string dedupe (`UseStringDeduplication` disabled by default on JDK11) on JDK11+ for `ResponseTime` but I believe some latency sensitive services saw perf regressions with G1 in certain conditions on JDK11 specifically and switched back to CMS, so should be cautious with general fleet rollout.

robert3005:
Agree this is a poor way to achieve the goal I intended. The right thing would be to change default profile but after looking at some benchmarks again don’t think it’s worth it.

PR #992, opened 2020-08-27
go-java-launcher 1.9.1 -> 1.10.0

GrahamDennis:
## Before this PR
go-java-launcher was built against go 1.10.3, which is old.

## After this PR

==COMMIT_MSG==
Upgrade go-java-launcher 1.9.1 -> 1.10.0
==COMMIT_MSG==

## Possible downsides?
Go in these binaries is upgraded to 1.15. Could there be any downstream effects due to this?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ buildscript {
     dependencies {
         classpath 'com.palantir.javaformat:gradle-palantir-java-format:1.0.1'
         classpath 'com.netflix.nebula:gradle-info-plugin:9.1.1'
-        classpath 'com.palantir.gradle.consistentversions:gradle-consistent-versions:1.26.0'
+        classpath 'com.palantir.gradle.consistentversions:gradle-consistent-versions:1.25.0'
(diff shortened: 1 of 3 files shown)


carterkozak (approved):

GrahamDennis:
@carterkozak: I can't add labels to this PR. Would you mind adding autorelease / merge when ready?

GrahamDennis:
This appears to be failing because gradle-revapi tries to resolve configurations during its own dependency resolution, and this breaks when used with gradle-consistent-versions.

GrahamDennis:
I created a test for gradle-revapi that demonstrates the issue when used with gradle-consistent-versions here: https://github.com/palantir/gradle-revapi/pull/307. I'm not sure the correct way to fix gradle-revapi to unblock this.

ferozco:
@GrahamDennis thanks for opening this up. To avoid blocking on GCV can you downgrade to 1.25.0 on your fork?

GrahamDennis:
@ferozco: This is now ready to be merged, and can you autorelease this too please?

PR #1005, opened 2020-09-28
Disable string deduplication

j-baker:
Internally we've seen that string deduplication can cause GC overheads to go way up (from 1.5% STW to ~15% STW). This may contribute to the number of people who've tried G1 and needed to turn it off. Given the all-or-nothing nature of this flag, this makes the default garbage collector be rather unsuitable for general use - it has overly sharp edges. We should probably make the hybrid profile not enable it by default, and instead enable it in specific places where it makes sense, if at all.

### DIFF
--- a/changelog/@unreleased/pr-1005.v2.yml
+++ b/changelog/@unreleased/pr-1005.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Disable string deduplication for the G1 collector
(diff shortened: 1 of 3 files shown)


robert3005 (approved):

carterkozak (commented):
:+1: this would be disabled moving to any other garbage collector, in favor of allowing us to compare gcs more directly

PR #1007, opened 2020-10-06
UseNUMA with G1GC.

neiln31l:
## Before this PR
Where systems are running with a single NUMA node, this would have no effect where JDK >= 11. Refer to https://hg.openjdk.java.net/jdk8u/jdk8u/hotspot/file/2198ef7e1c17/src/os/linux/vm/os_linux.cpp#l4931.

## After this PR
==COMMIT_MSG==
Where systems are running with multiple NUMA nodes, this would enable G1GC to make better decisions.
==COMMIT_MSG==

## Possible downsides?
This should be a no-op for systems with single NUMA nodes while benefiting multiple NUMA node systems.

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
@@ -91,7 +91,7 @@ public interface GcProfile extends Serializable {
     class Hybrid implements GcProfile {
         @Override
         public final List<String> gcJvmOpts(JavaVersion javaVersion) {
-            return ImmutableList.of("-XX:+UseG1GC", "-XX:+UseNUMA", "-XX:+PrintFlagsFinal");
+            return ImmutableList.of("-XX:+UseG1GC", "-XX:+UseNUMA");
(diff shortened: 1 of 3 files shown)


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
```diff
@@ -91,7 +91,7 @@ public final void newRatio(int newerRatio) {
     class Hybrid implements GcProfile {
         @Override
         public final List<String> gcJvmOpts(JavaVersion javaVersion) {
-            return ImmutableList.of("-XX:+UseG1GC");
+            return ImmutableList.of("-XX:+UseG1GC", "-XX:+UseNUMA", "-XX:+PrintFlagsFinal");
```
carterkozak: I’m not sure we should make PrintFlagsFinal dependent on the GC profile

neiln31l: Fair enough -- I've removed it. 👍

carterkozak (approved):
Thanks!

PR #1012, opened 2020-10-14
Lock asset configuration with GCV

dansanduleac:
## Before this PR

The assetBundle configuration is not locked.

Due to what I think is a bug in GCV, [published constraints](https://github.com/palantir/gradle-consistent-versions/blob/6372d7ffd97753d65d6447119472eb2dc349928f/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java#L1003-L1017) (i.e. an outgoing java configuration like `runtimeElements` and `apiElements` exposing minimum version constraints for its transitive dependencies) somehow don't get exposed correctly when running with `--write-locks`.

This leads to the resolution of the `assetBundle` configuration to differ if you run with `--write-locks` compared with the normal case.

If so, then product dependencies sometimes change too, causing `createManifest` to produce one output when running with `--write-locks`, but expect another in the normal case.

## After this PR

==COMMIT_MSG==
If gradle-consistent-versions is detected, lock the `assetBundle` configuration, to ensure that it always gets the locked dependency versions.
==COMMIT_MSG==


gradle-sls-packaging/build.gradle
```diff
@@ -23,6 +26,16 @@ dependencies {
...
 }
 
+
+/**
+ * Per default, Gradle TestKit excludes compileOnly dependencies resulting in ClassNotFoundExceptions when sls-packaging
+ * tries to access 'gradle-consistent-versions' classes in integ tests e.g. in 'AssetDistributionPluginIntegrationSpec'.
+ * https://discuss.gradle.org/t/testkit-somehow-excludes-compileonly-dependencies
+ */
+tasks.withType(PluginUnderTestMetadata).configureEach {
+    pluginClasspath.from(configurations.testRuntime)
```
fawind, line 36: This is pretty hacky but I could not get it to work otherwise.

The problem is that TestKit does not put compileOnly dependencies on the same classpath as the plugin under test which results in a `NoClassDefFoundError: com/palantir/gradle/versions/VersionsLockExtension` when running tests ([circle](https://app.circleci.com/pipelines/github/palantir/sls-packaging/743/workflows/b23dece6-1c41-426c-bc4f-021501baa28a/jobs/10049)).

See https://github.com/gradle/gradle/issues/11338#issuecomment-568228102 and https://discuss.gradle.org/t/testkit-somehow-excludes-compileonly-dependencies for more info.

I tried adding the GCV plugin via buildscript dependency or just as a `testCompile/testRuntime` dependency but got the same error. Using `pluginClasspath.from(configurations.compileOnly)` also fails at configuration time with `Could not resolve all files for configuration ':gradle-sls-packaging:compileOnly'. > Could not find com.palantir.sls-packaging:sls-packaging:4.16.2-20-g3611d3b.dirty`.

Note that this is only an issue in tests. When ran on a real project, gcv classes are on the same classpath when `project.getRootProject().getPlugins().withId("com.palantir.consistent-versions", plugin -> { //... })` gets evaluated.

iamdanfox (approved):
I like it!

fawind:
Tested it on the two internal repos where we encountered the problem and its looking good!

Power-merging and overwriting the `cla/palantir` check as Dan opened the PR while still working here which ensures compliance.

PR #1013, opened 2020-10-14
Fix: Downgrade jackson to 2.11.1

robert3005:
## After this PR
==COMMIT_MSG==
Downgrade jackson to 2.11.1 to avoid jackson bugs
==COMMIT_MSG==


iamdanfox:
Why make this change to gradle plugins?

robert3005:
I have a repo that has a service and a gradle plugin and it gets higher version from this plugin. I am happy to force downgrade in repo but I thought more people are going to have the same problem

carterkozak:
+1 for uniformity.

ferozco:
👍

iamdanfox:
🦜

Issue #998, opened 2020-09-07
Cannot easily override GC settings via configuration


j-baker:
sls-packaging encourages the use of gc defaults, e.g. 'response-time', 'hybrid', 'throughput'. These use cms/shenandoah, g1, throughput collectors in usually sensible ways.

Unfortunately, they are specifiable only once, at service authorship time, and never at configuration time. This makes things tricky. If one e.g. wants to switch from response-time to hybrid via config, one must write things like `-XX:-UseConcMarkSweepGC` to disable the previously added flags. But not all of the flags are actually unsettable, e.g. new gen ratio. This means that we poorly support a few use cases:

1. Trying out a different garbage collector. Ideally we'd be able to test for an extended period on a single installation, but right now that means tagging an RC release with the new profile and holding back all other upgrades for that deployment.
2. Setups where a single installation has a very different workload for all others. For example, we have a setup where most deployments have a 16GB heap, but one has a 64GB heap and another has a 256GB heap. In such a case, many of the default settings (e.g. new size ratio) become incorrect. This is clearly an antipattern, but we've multiple times found that fixing the issues that cause you to need a 256GB heap is a process that takes months if not years. We don't end up with a 256GB heap via laziness, we end up with that because it's the local maximum.

I think the intent of this was to remove as much from configuration as possible, but in practice it adds friction to testing out GC changes and instead has led in at least one case to service binary monkey patching (e.g. manually modifying launcher-static.yml (see PDS-126032 internally)) (which seems worse than the alternative). It would be great if we could specify one of the gc profiles via launcher custom.

I think there are some impediments in the implementation; right now I think the gc profiles are implemented purely at the gradle level. If the go-java-launcher were to have a little more 'sympathy' for how we define configuration internally then we'd have less issue. As one brief proposal, you could imagine launcher static containing maps of options vs simply a list

```yaml
profile-definitions:
  # populated automatically by gradle plugin
  gc-hybrid:
    - "-XX:+UseG1GC"
  gc-response-time:
   - "-XX:+UseConcMarkSweepGC"
  # added by user
  experimental-disable-biased-locking:
   - "-XX:-UseBiasedLocking"
profiles:
  gc-hybrid: true
```

and then launcher custom could contain something like:

```yaml
configType: java
configVersion: 1 # this is fully backwards compatible
profiles:
  gc-hybrid: false
  gc-response-time: true # hybrid didn't work very well
  experimental-disable-biased-locking: true
jvmOpts:
[...]

PR #1015, opened 2020-10-15
Add empty GC profile

fawind:
## After this PR

Introduce an empty GC profile called `dangerous-no-profile`, that does not apply any GC flags. This is intended to allow easier overriding of GC settings for performing GC experiments or other "expert" usage and is not intended to be used by most services.

More context in https://github.com/palantir/sls-packaging/issues/998

==COMMIT_MSG==
Add empty GC profile
==COMMIT_MSG==

## Possible downsides?
Brings in the danger that users experiment with GC flags and then forget to revert the profile afterwards. Maybe we should leave it undocumented or put in some highly visible logging when selecting this empty profile?

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
@@ -33,7 +33,7 @@ public interface GcProfile extends Serializable {
             "throughput", GcProfile.Throughput.class,
             "response-time", GcProfile.ResponseTime.class,
             "hybrid", GcProfile.Hybrid.class,
-            "no-profile", GcProfile.NoProfile.class);
+            "dangerous-no-profile", GcProfile.NoProfile.class);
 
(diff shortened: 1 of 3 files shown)


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
```diff
@@ -31,7 +32,8 @@
     Map<String, Class<? extends GcProfile>> PROFILE_NAMES = ImmutableMap.of(
             "throughput", GcProfile.Throughput.class,
             "response-time", GcProfile.ResponseTime.class,
-            "hybrid", GcProfile.Hybrid.class);
+            "hybrid", GcProfile.Hybrid.class,
+            "no-profile", GcProfile.NoProfile.class);
```
ferozco: can we name this `dangerous-no-profile`

fawind: 👍 Updated.

ferozco (approved):

PR #1016, opened 2020-10-15
Use G1 as default GC for Java 15 and above

fawind:
## After this PR

The default GC profile for Java 15 and above is now `hybrid` which uses G1. For previous Java versions, we still use `throughput` which uses `ParallelOldGc`.

==COMMIT_MSG==
Use hybrid as default GC profile for Java 15 and above
==COMMIT_MSG==


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
```diff
@@ -38,6 +38,10 @@
     class Throughput implements GcProfile {
         @Override
         public final List<String> gcJvmOpts(JavaVersion javaVersion) {
+            // For Java 15 and above, use G1 as the default garbage collector
+            if (javaVersion.compareTo(JavaVersion.toVersion("14")) > 0) {
```
ferozco: I think we want to change the default GCProfile that is used rather than changing the behaviour of `throughput` when on Java 15+

ferozco (approved):

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java
fawind: Makes sense. Updated the PR to use `hybrid` for java >= 15 and `throughput` for java < 15.

PR #1026, opened 2020-10-30
Java services record heap dumps on OOM

carterkozak:
## Before this PR
Previously OOMs did not result in heap dumps, only crash files.

## After this PR
Now a heap dump is recorded to `var/log/heapdump.hprof` in the
event of an OOM. If an hprof already exists in this location, it
will not be overwritten, and no hprof will be recorded to prevent
services that quickly OOM on startup from filling the disk.
Any time a service OOMs there's action required to figure out
precisely what went wrong. Without recording heap, there may
not be any bread crumbs to understand the failure scenario.

==COMMIT_MSG==
Java services record heap dumps to `var/log/heapdump.hprof` on OOM. If the file already exists from a previous failure, no new heap dump is recorded.
==COMMIT_MSG==

## Possible downsides?
Heap dumps are large, even a single heap dump may dominate disk
utilization compared to all other logging combined. In the worst
case this may fill the disk and prevent a service from restarting,
however this is a price we pay for the ability to understand how
products have failed.
Services which already set `-XX:+HeapDumpOnOutOfMemoryError` and
expect multiple heap dumps to be recorded will need to add the
`-XX:HeapDumpPath=var/log/` option to override defaults if they wish
to avoid changing behavior.

### DIFF
--- a/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
+++ b/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
@@ -44,7 +44,7 @@ import org.immutables.value.Value;
 
 public class LaunchConfigTask extends DefaultTask {
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper(new YAMLFactory());
-    private static final List<String> java8gcLoggingOptions = ImmutableList.of(
+    private static final ImmutableList<String> java8gcLoggingOptions = ImmutableList.of(
             "-XX:+PrintGCDateStamps",
             "-XX:+PrintGCDetails",
             "-XX:-TraceClassUnloading",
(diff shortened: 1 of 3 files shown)


carterkozak:
`do not merge` while collecting feedback

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -53,13 +53,18 @@
...
+    private static final ImmutableList<String> alwaysOnJvmOptions = ImmutableList.of(
             "-XX:+CrashOnOutOfMemoryError",
+            "-XX:+HeapDumpOnOutOfMemoryError",
             "-Djava.io.tmpdir=var/data/tmp",
             "-XX:ErrorFile=var/log/hs_err_pid%p.log",
-            "-XX:HeapDumpPath=var/log",
+            // Provide a full filename to avoid filling disks with heap dumps.
+            // If a heap dump already exists in this location, subsequent OOMs
+            // will not dump heap.
+            "-XX:HeapDumpPath=var/log/heapdump.hprof",
```
iamdanfox, line 67: presumably this also means that if we trigger a hprof via jcmd it'll go to this path by default?

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -53,13 +53,18 @@
...
             "-Xloggc:var/log/gc-%t-%p.log",
             "-verbose:gc");
-    private static List<String> java14Options = ImmutableList.of("-XX:+ShowCodeDetailsInExceptionMessages");
+    private static final ImmutableList<String> java14Options =
+            ImmutableList.of("-XX:+ShowCodeDetailsInExceptionMessages");
 
-    private static final List<String> alwaysOnJvmOptions = ImmutableList.of(
+    private static final ImmutableList<String> alwaysOnJvmOptions = ImmutableList.of(
             "-XX:+CrashOnOutOfMemoryError",
+            "-XX:+HeapDumpOnOutOfMemoryError",
```
iamdanfox, line 61: When running in k8s, writing out heap dumps to var/log is gonna be pretty pointless i think, but maybe not actually bad?

iamdanfox:
So historically we've always shied away from doing this out of fear that runaway disk usage will wreck us in exciting ways.  I think this approach solves this quite nicely, because if you've already OOMed and written a heap dump _once_, there's clearly something broken with your service and someone should probably look into it.  If that single heap dump _was_ enough to fill up disk, then that person will need to move/delete the heapdump before they can proceed, but at least there's someone looking at it.

If a service _isn't_ hobbled by a single OOM (i.e. it comes back to life without anyone noticing) and the OOM doesn't fill up disk, then it can just stay that way until someone notices and comes round to investigate, because only a single hprof will ever exist on disk.

Overall, I think it's a :+1: from me.

markelliot:
I think that the safety of this, then, assumes:
 1. Services are guaranteed to affect only their own disk
 2. Service owners have (a) a means retrieve, (b) a right to read, and (c) a means to delete the heap dump

I'm not sure either of these are close to present reality. Additionally, waiting for disks to fill in order to generate an actionable alert seems like a bad idea. Minimally, if this change goes in, our infrastructure should _proactively_ alert on the existence of a heap dump.

We used to explicitly capture heap dumps and removed this flag (I'm, sure the PR is in this repo somewhere) because (a) there was no way to retrieve the file, (b) the people who needed to read the file didn't have the right to and (c) the system owners rarely if ever deleted the files, leading to disks filling and impacting unrelated services on the same hosts.

PR #1033, opened 2020-11-17
Automatically wire up the --enable-preview option if necessary

iamdanfox:
## Before this PR

In baseline 3.52.0 we added this new plugin as a one-liner to add the --enable-preview flag wherever necessary (https://github.com/palantir/gradle-baseline/pull/1549). We're using the extraProperties thing to avoid taking a compile dependency on baseline.

## After this PR
==COMMIT_MSG==
sls-packaging will automatically include the `--enable-preview` jvm arg in launcher-static.yml if `com.palantir.baseline-enable-preview-flag` is applied.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1033.v2.yml
+++ b/changelog/@unreleased/pr-1033.v2.yml
@@ -1,0 +1,6 @@
+type: feature
+feature:
+  description: sls-packaging will automatically include the `--enable-preview` jvm
+    arg in launcher-static.yml if `com.palantir.baseline-enable-preview-flag` is applied.
+  links:
+  - https://github.com/palantir/sls-packaging/pull/1033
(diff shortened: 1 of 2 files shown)


ferozco (commented):
What do you think about keeping all of this inside of baseline? I think it might be a bit cleaner since consolidates the preview logic in one place and stays true to the nature of baseline configuring other plugins correctly

iamdanfox:
Right, that's the other options for where to put this stuff.  I wanted to preserve the situation we have now where there is no compilation dependency from sls-packaging to baseline or vice versa.   It seems that if we put the logic in baseline and make it 'reach out' to configure sls-packaging then we'll need to have a compile dependency in order to get hold of the `JavaServiceDistributionExtension`.

I figured doing it this way round (with the `Provider<Boolean>`) avoid this kinda nicely? Also I figured that _in theory_ there might be lots of different plugins that could all need to read this flag (e.g. maybe some gradle-graal things too??) so exposing a nice readable Provider from baseline seemed sort-of future-proof.

iamdanfox:
~Somewhat related thought - what do you think about baking something into our manifest or something so that we can do a papaya analysis and keep track of all the releases that enable this flag?? Feels like at some point we're going to want to keep an eye on who's using this!~

EDIT nvm there's a dataset in papaya of launcher-static files already, so we can just detect the `--enable-preview` flag in that: ri.foundry.main.dataset.e3c4b54c-99eb-48cf-9f6d-df35d58e77bf

ferozco (approved):

PR #1034, opened 2020-11-18
[wip] Populate the 'diagnostics' manifest extension, discovering info from runtimeClasspath (and local files)

iamdanfox:
## Before this PR

@bmoylan is working on the Debug Service, spec'd out internally here: https://github.p.b/deployability/sls-spec/pull/521).

Essentially, servers might be capable of responding to a few different types of request (e.g. _jstack.v1_ or _heap.histogram.v1_), and we want a user interface to be able to display clickable buttons to trigger these.  To make this declarative, the sls-spec PR above requires information about which types are supported to be embedded in the manifest that is present in all sls.tgzs.

## After this PR
==COMMIT_MSG==
For java services, `createManifest` task will now populate the `diagnostics` extension of the sls manifest.
==COMMIT_MSG==

For example, 
```diff
 {
   "manifest-version" : "1.0",
   "product-type" : "service.v1",
   "product-group" : "com.palantir.log-receiver",
   "product-name" : "log-receiver",
   "product-version" : "1.125.0-1-g594cdd6.dirty",
   "extensions" : {
     "upgrade-strategy" : "rolling",
     "public-proxy-endpoints" : [ "service-endpoint" ],
     "product-dependencies" : [ {
       "product-group" : "com.palantir.s",
       "product-name" : "a",
       "minimum-version" : "1.51.0",
       "maximum-version" : "2.x.x"
     }, {
       "product-group" : "com.palantir.mp",
       "product-name" : "mp",
       "minimum-version" : "3.297.0",
       "maximum-version" : "3.x.x"
     } ],
+    "diagnostics" : [ {
+      "type" : "bar.v1"
+    }, {
+      "type" : "foo.v1"
+    } ]
   }
 }
```

Needs:
- [ ] docs
- [ ] tests
- [ ] proper up-to-dateness stuff
- [ ] any more logging / observability??

PR #1035, opened 2020-11-19
[v2] Populate the 'diagnostics' manifest extension, discovering info from runtimeClasspath (and local files)

iamdanfox:
## Before this PR

bmoylan is working on the Debug Service, spec'd out internally here: https://github.p.b/deployability/sls-spec/pull/521).

Essentially, servers might be capable of responding to a few different types of request (e.g. jstack.v1 or heap.histogram.v1), and we want a user interface to be able to display clickable buttons to trigger these. To make this declarative, the sls-spec PR above requires information about which types are supported to be embedded in the manifest that is present in all sls.tgzs.

_I took a stab at implementing this earlier (https://github.com/palantir/sls-packaging/pull/1034), but it felt like we were writing yet another hacky implementation of the idea of extracting a single file from jars. We've found caching these things are a bit subtle, so I wondered could we build it up using first-class gradle features?_

## After this PR
==COMMIT_MSG==
For java services, a new `diagnostics` extension of the sls manifest will be populated based on the `sls-manifest/diagnostics.json` file found locally or in any jar in runtimeClasspath. Contents should look like `{["type": "foo.v1", "docs": "..."]}`, as defined in the sls-spec.
==COMMIT_MSG==

Inspired by @ferozco's usage in preflight, this PR uses gradle's concept of 'attributes' and 'artifact transforms' to do the actual extracting. https://docs.gradle.org/current/userguide/artifact_transforms.html

By expressing the work in terms of gradle's first-class building blocks, we get much nicer parallelization, caching and debuggability support.

## Possible downsides?
- Using more gradle features means we're more exposed to possible gradle bugs.  In this specific case I don't think it's actually _too_ scary, because they seem to use attributes _very_ heavily internally (i.e. it's what allows consumers to ask for either raw classes or a zipped up jar).


gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/DiagnosticsManifestPlugin.java
```diff
@@ -0,0 +1,190 @@
...
+    @Override
+    public void apply(Project project) {
+        String fileToExtract = "sls-manifest/diagnostics.json";
+        String attribute = "extracted-" + fileToExtract;
+
+        project.getDependencies().registerTransform(ExtractSingleFile.class, details -> {
+            details.getParameters().getPathToExtract().set(fileToExtract);
+
+            // this USAGE_ATTRIBUTE is already present on everything, so gradle can figure out how to transform to our
+            // attribute value
```
iamdanfox, line 77: Usage is defined here https://github.com/gradle/gradle/blob/master/subprojects/core-api/src/main/java/org/gradle/api/attributes/Usage.java

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/DiagnosticsManifestPlugin.java
```diff
@@ -0,0 +1,190 @@
...
+        // configuration and add a 'self' dependency.
+        Configuration consumable = project.getConfigurations().create("runtimeClasspath2", conf -> {
+            conf.extendsFrom(project.getConfigurations().getByName("runtimeClasspath"));
+            conf.setDescription("DiagnosticsManifestPlugin uses this configuration to extract single file");
+            conf.setCanBeConsumed(true);
+            conf.setCanBeResolved(true);
+        });
+        project.getDependencies().add(consumable.getName(), project);
+
+        ArtifactView myView = consumable.getIncoming().artifactView(v -> {
```
ferozco: haven't played with artifactViews before but I'm pretty sure you can accomplish something similar by setting attributes on the configuration itself.

Either way, I think that this approach has a bit of a problem since it requires all project dependencies to be compiled, packaged into a JAR and then unpackaged to resolve the supported diagnostics. We've tried pretty hard to avoid doing so since `createManifest` is executed on `--write-locks` and compiling all dependant projects can be quiet expensive

iamdanfox: Ok have added a direct java-resources -> my attribute transform, so we now pass those tests that check the createManifest task doesn't compile stuff!!

As to your question about artifactView, I did actually try setting the attributes on the configuration directly, but couldn't seem to get it work work... have written a little essay in the comments to hopefully explain its existence. Maybe we can come back with greater understanding and delete it!!

iamdanfox:
I've done a publishToMavenLocal and ran this on devx/template-witchcraft with the new version of WC. Here's the internal build scan: https://gradle.p.b/s/bqg27jkbkglk6

gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/DiagnosticsManifestPlugin.java
```diff
@@ -0,0 +1,260 @@
...
+     * - "attributes" (https://docs.gradle.org/current/userguide/variant_attributes.html)
+     * - "artifact transforms" (https://docs.gradle.org/current/userguide/artifact_transforms.html)
+     *
+     * These seem to be the new idiomatic way of doing things (and Gradle uses them _heavily_ internally, e.g. to
+     * delineate the 'api' vs 'implementation' stuff in the {@link org.gradle.api.plugins.JavaPlugin}).
+     * They let us declaratively express things in a _really_ nice way,
+     * with beautiful granular caching etc. See https://docs.gradle.org/current/userguide/variant_model.html for some
+     * helpful diagrams. Also try running `./gradlew outgoingVariants` on a project to visualize what's going on.
+     *
+     * We define a couple of {@link TransformAction}s so that when we define we want just than extracted file, and
```
fawind:
```suggestion
     * We define a couple of {@link TransformAction}s so that when we define that we want just an extracted file, and
```

ferozco:
Pushed up some thoughts on a cleanup/refactor: https://github.com/palantir/sls-packaging/compare/fo/refactor?expand=1

ferozco (approved):

ferozco:
👍

iamdanfox:
I also approve felipe's commits :+1:

PR #1039, opened 2020-11-24
[test-only] Fix JavaServiceDistributionPluginTest's non-existent external.jar

iamdanfox:
## Before this PR

I've got a failing test in one of my other PRs because it tries to analyze a non-existent jar. I think this test is actually not representative of the real world, so I'm just landing an independent commit so that I can make sure I don't break anything else independently!

## After this PR
==COMMIT_MSG==
Fix JavaServiceDistributionPluginTest's non-existent external.jar
==COMMIT_MSG==


iamdanfox:
power merging test-only change

PR #1040, opened 2020-11-25
[stacked] CR

ferozco:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


iamdanfox (approved):
Chatted over zoom, think I'm fine with these tweaks :)

PR #1048, opened 2020-11-30
Disable TLS session tickets by default

carterkozak:
## Before this PR
We've run into some incompatibilities that haven't been debugged
yet, and prefer to defer session ticket rollout until we've
upgraded JVMs.

https://www.oracle.com/java/technologies/javase/14-relnote-issues.html#JDK-8228396

## After this PR
==COMMIT_MSG==
Disable TLS session tickets by default to decouple JVM upgrades from significant TLS changes.
==COMMIT_MSG==

## Possible downsides?
Straying from defaults always introduces risk. Most java shops have not upgraded to jre14/15 yet, so it may be safer to revert to battle-tested non-session-ticket code.

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
@@ -62,7 +62,12 @@ public class LaunchConfigTask extends DefaultTask {
             "-XX:HeapDumpPath=var/log",
             // Set DNS cache TTL to 20s to account for systems such as RDS and other
             // AWS-managed systems that modify DNS records on failover.
-            "-Dsun.net.inetaddr.ttl=20");
+            "-Dsun.net.inetaddr.ttl=20",
+            // Disable RFC 5077 TLS session ticket support due to incompatibility with common implementations.
+            // This allows us to decouple JVM upgrades from enabling session ticket support.
+            // See https://www.oracle.com/java/technologies/javase/14-relnote-issues.html#JDK-8228396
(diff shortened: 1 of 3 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -62,7 +62,12 @@
             "-XX:HeapDumpPath=var/log",
             // Set DNS cache TTL to 20s to account for systems such as RDS and other
             // AWS-managed systems that modify DNS records on failover.
-            "-Dsun.net.inetaddr.ttl=20");
+            "-Dsun.net.inetaddr.ttl=20",
+            // Disable RFC 5077 TLS session ticket support due to incompatibility with common implementations.
```
iamdanfox: rather than common implementations, can we explicitly call out the problem of each server node just making up it's own session key (and then nginx breaking), with a specific mentionof PDS-123520?

carterkozak: I don't think the servers using different keys is the problem we're hitting with nginx, that would be the case if we had dns resolution across several nodes or transparent load balancing. However nginx knows each upstream is a unique entity with its own key. We can name PDS-123520 and nginx though

carterkozak:
Adding do-not-merge for now, I'm doing some research today, and will plan to merge by EoD if I'm unable to discover anything new.

carterkozak:
Feel free to approve if you're happy with the the PR though.

iamdanfox (approved):

PR #1053, opened 2020-12-01
Add recommended product dependencies to all jars

CRogers:
## Before this PR
Add `recommendProductDependencies` to all jars, not just to the `jar` task. This then should handle the case where we have a `shadowJar` from the shadow plugin.

## After this PR

==COMMIT_MSG==
Recommended product dependencies are added to the manifests of all jar tasks in the projects where `com.palantir.recommended-product-dependencies` is applied. This means `shadowJar` from the gradle shadow plugin now picks up recommended product dependencies.
==COMMIT_MSG==

## Possible downsides?
Maybe we add the pdeps to too many jars? Doesn't seem like too bad a downside.

### DIFF
--- a/gradle-recommended-product-dependencies/src/test/groovy/com/palantir/gradle/dist/RecommendedProductDependenciesPluginIntegrationSpec.groovy
+++ b/gradle-recommended-product-dependencies/src/test/groovy/com/palantir/gradle/dist/RecommendedProductDependenciesPluginIntegrationSpec.groovy
@@ -95,14 +95,69 @@ class RecommendedProductDependenciesPluginIntegrationSpec extends IntegrationSpe
         runTasksSuccessfully('-DignoreDeprecations=true', '--write-locks', ':jar')
 
         then:
-        fileExists("build/libs/root-project.jar")
+        verifyCorrectRecommendedProductDeps('build/libs/root-project.jar')
+    }
+
+    def 'when gradle-shadow-jar is applied, the pdeps are put into the thin and shadow jar manifest'() {
+        settingsFile  << '''
+            rootProject.name = "root-project"
+        '''.stripIndent()
+
+        buildFile << '''
+            buildscript {
+                repositories {
+                    jcenter()
+                    maven { url 'https://dl.bintray.com/palantir/releases' }
+                }
+            
+                dependencies {
+                    classpath 'com.palantir.gradle.consistentversions:gradle-consistent-versions:1.27.0'
+                    classpath 'com.palantir.gradle.shadow-jar:gradle-shadow-jar:1.2.0'
+                }
+            }
+            
+            apply plugin: 'java-library'
+            apply plugin: 'com.palantir.recommended-product-dependencies'
+            apply plugin: 'com.palantir.consistent-versions'
+            apply plugin: 'com.palantir.shadow-jar'
+            
+            repositories {
+                jcenter()
+            }
+            
+            dependencies {
+                implementation 'one.util:streamex:0.7.3'
+            }
+
+            recommendedProductDependencies {
+                productDependency {
+                    productGroup = 'group'
+                    productName = 'name'
+                    minimumVersion = '1.0.0'
+                    maximumVersion = '1.x.x'
+                }
+            }
+        '''.stripIndent()
+
+        when:
+        runTasksSuccessfully('--write-locks', 'jar', 'shadowJar')
+
+        then:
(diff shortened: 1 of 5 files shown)


ferozco:
@CRogers I had tried this out with shadow-jar and it didn't seem to work properly

CRogers:
@ferozco well, I got the test working in the repo for gradle-shadow-jar for now, the issue seems to be in a-c-j that the `shadowJar` task never produces a manifest, even if you just give it some manifest attributes in gradle:

```gradle
shadowJar {
    manifest {
        attributes('Foo': 'Bar')
    }
}
```

It just emits nothing. So this change with pTML does not work. I assume it's some kind of repo specific issue tripping up the jrenglemen shadow plugin.

CRogers:
Looks like the combination of this and https://github.com/palantir/gradle-shadow-jar/pull/67 fixes the issue in a-c-j! This ensures we actually get the data into the manifest at the right time in the build process, https://github.com/palantir/gradle-shadow-jar/pull/67 means we don't erroneously relocate the manifest as it's in the same package as stuff that's being relocated.

gradle-sls-packaging/build.gradle
```diff
@@ -58,7 +58,7 @@ pluginBundle {
     }
 }
 
-def yourkitVersion = "2020.9-b411"
+def yourkitVersion = "2020.9-b412"
```
iamdanfox, line 61: lol

gradle-sls-packaging/build.gradle
```diff
@@ -58,7 +58,7 @@ pluginBundle {
     }
 }
 
-def yourkitVersion = "2020.9-b411"
```
CRogers, line 61: yourkit deleted this. Should probably mirror this internally and avoid using the download task so we don't suffer from random remote deletions again (also Download task sucks).

gradle-recommended-product-dependencies/src/test/groovy/com/palantir/gradle/dist/RecommendedProductDependenciesPluginIntegrationSpec.groovy
```diff
@@ -95,14 +95,60 @@ class RecommendedProductDependenciesPluginIntegrationSpec extends IntegrationSpe
         runTasksSuccessfully('-DignoreDeprecations=true', '--write-locks', ':jar')
 
         then:
-        fileExists("build/libs/root-project.jar")
+        verifyCorrectRecommendedProductDeps('build/libs/root-project.jar')
+    }
+
+    def 'when gradle-shadow-jar is applied, the pdeps are put into the shadow jar manifest'() {
```
iamdanfox, line 101: is it worth adding a multi-project test here too to more closely mirror the a-c-j use-case?

gradle-sls-packaging/build.gradle
CRogers: Scratch that, can't mirror this internally when this project is open source.

iamdanfox (approved):
Looks great to me, no strong preference about whether we do the extra test or just run with the fix as is!

ferozco:
I think we actually just want to add a dependency between `shadowJar` and `jar` since it already will inherit the manifest

iamdanfox:
@ferozco why is that better? It seems plausible that we might also come up with some kind of `publish-minimized-jar` plugin in the future too, and there's still one central place to control these things??

iamdanfox:
Also I think callum's current implementation seems kinda efficient because if you just want the ./gradlew shadowJar then it won't need to even zip up classes into a regular jar??

ferozco:
Well I think adding the dependency is a sure fire way to ensure that _all_ additions to the normal Jars manifest are replicated in the shadowJar. Otherwise each thing that contributes to the manifest will have to be aware of shadowJar.

You can avoid bundling the original jar by either applying all of the dependencies of jar to shadowJar or by marking jar to be skipped

carterkozak:
I'm not sure we need to add product deps to the sources or javadoc jars either, not that there's any harm in it.

CRogers:
The `shadowJar.depends jar` approach does seem to now work. I'm worried it might rely on pure luck, as the manifest is "inherited" at task configuration time, but I think the `Manfiest` object is mutable so it gets the updates even if the `shadowJar` was configured (but not run) before the `configureProductDependencies` task.

CRogers:
Closing this out to just add the dependsOn in `gradle-shadow-jar`

CRogers:
PR to do that here: https://github.com/palantir/gradle-shadow-jar/pull/74

PR #1054, opened 2020-12-04
tolerate extra json keys in diagnostics manifest

iamdanfox:
## Before this PR
@carterkozak has a PR on WC internally to add this `safe : true` field.  We realised it's actually pretty lame to have a coupling between WC and sls-packaging, so this PR removes the validation that would block this stuff.

## After this PR
==COMMIT_MSG==
When parsing the `sls-manifest/diagnostics.json` file out of jars, sls-packaging no longer performs strict validation on its structure (as this can prevent the rollout of jars which include extra fields). Instead it just passes everything on as long as it's a JSON object.
==COMMIT_MSG==

## Possible downsides?
- in theory one jar could possibly include a non-sensical diagnostics.json which would be merged into the single output file, potentially 'poisoning' the ability of downstream stuff to deserialize the whole thing.

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/Diagnostics.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/Diagnostics.java
@@ -16,110 +16,40 @@
 
 package com.palantir.gradle.dist.service;
 
-import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonValue;
 import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.palantir.gradle.dist.tasks.CreateManifestTask;
-import com.palantir.logsafe.Preconditions;
-import com.palantir.logsafe.SafeArg;
-import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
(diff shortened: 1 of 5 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/Diagnostics.java
```diff
@@ -16,112 +16,42 @@
...
-
-    public static List<SupportedDiagnostic> parse(Project proj, File file) {
+    public static List<ObjectNode> parse(Project proj, File file) {
         Path relativePath = proj.getRootDir().toPath().relativize(file.toPath());
         String string = null;
         try {
             string = new String(Files.readAllBytes(file.toPath()), StandardCharsets.UTF_8).trim();
-            List<SupportedDiagnostic> value =
-                    CreateManifestTask.jsonMapper.readValue(string, new TypeReference<List<SupportedDiagnostic>>() {});
+            List<ObjectNode> value =
```
ferozco, line 43: `Map<String, Object>`?

carterkozak (approved):

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/Diagnostics.java
iamdanfox: eh I actually prefer the ol ObjectNode types, it feels a bit more 'pure data' than having that `Object` value type, as technically that could be one of those jackson vulns like a FileOutputStream (which is technically an object)

PR #1060, opened 2020-12-17
Improvement: Expose pdeps as a property

ferozco:
## After this PR

==COMMIT_MSG==
Expose pdeps as a property
==COMMIT_MSG==


robert3005:
👍

PR #1065, opened 2021-01-14
Add workaround for JDK-8253566 on Java 15

fawind:
## Before this PR
There is a regression in j15 where `clazz.isAssignableFrom` will return false for interface implementors. See: https://bugs.openjdk.java.net/browse/JDK-8253566

As a workaround, you can add `-XX:+ExpandSubTypeCheckAtParseTime` to your jvm args.

A fix got backported to 15.0.2 but the Azul release is still pending.

## After this PR
==COMMIT_MSG==
Add workaround for JDK-8253566 on Java 15
==COMMIT_MSG==

This was backported to Java 15 in 15.0.2.
https://bugs.openjdk.java.net/browse/JDK-8255390

We can remove the workaround once:

* Azul releases a version based on 15.0.2
* Our internal and external CI image is upgraded to use this version
* 15.0.2 is used in production

Might also need to also add this to peoples integ test build configs, but this is probably better done through excavator.

cc @pkoenig10 who discovered this

More internal context on https://g.p.b/f/mp/issues/10275

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
@@ -44,7 +44,7 @@ import org.immutables.value.Value;
 
 public class LaunchConfigTask extends DefaultTask {
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper(new YAMLFactory());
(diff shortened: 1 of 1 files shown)


fawind:
As per @carterkozak suggestion, we will hold off rolling out java 15 to the fleet for now and wait for the Azul 15.0.2 release as migrating off of this workaround can be quite tricky once its rolled out widely.

PR #1074, opened 2021-02-11
Add support for optional dependencies

nemines:
## After this PR

==COMMIT_MSG==
Support `optional` field in the `ProductDependency` class.
It also allows discovered dependencies to be set optional through the `optionalProductDependency` declaration in the manifest.
==COMMIT_MSG==

### DIFF
--- a/.palantir/revapi.yml
+++ b/.palantir/revapi.yml
@@ -7,3 +7,9 @@ acceptedBreaks:
       new: "method org.gradle.api.provider.SetProperty<com.palantir.gradle.dist.ProductDependency>\
         \ com.palantir.gradle.dist.RecommendedProductDependenciesExtension::getRecommendedProductDependenciesProvider()"
       justification: "not really a break"
+  "4.26.0":
+    com.palantir.sls-packaging:gradle-recommended-product-dependencies:
+    - code: "java.class.defaultSerializationChanged"
+      old: "class com.palantir.gradle.dist.ProductDependency"
+      new: "class com.palantir.gradle.dist.ProductDependency"
+      justification: "Old class is semantically equivalent to optional=false"
--- a/gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/ProductDependency.java
+++ b/gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/ProductDependency.java
@@ -45,6 +45,9 @@ public final class ProductDependency implements Serializable {
     @JsonProperty("maximum-version")
     private String maximumVersion;
 
+    @JsonProperty("optional")
+    private boolean optional = false;
+
     public ProductDependency() {}
 
     public ProductDependency(
@@ -192,6 +195,14 @@ public final class ProductDependency implements Serializable {
 
     public void setMaximumVersion(String maximumVersion) {
         this.maximumVersion = maximumVersion;
+    }
+
+    public boolean getOptional() {
+        return optional;
+    }
+
+    public void setOptional(boolean optional) {
+        this.optional = optional;
     }
 
     @Override
--- a/gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/ProductDependency.java
+++ b/gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/ProductDependency.java
@@ -164,8 +164,13 @@ public final class ProductDependency implements Serializable {
     @Override
     public String toString() {
         return String.format(
-                "%s:%s(min: %s, recommended: %s, max: %s)",
-                productGroup, productName, minimumVersion, recommendedVersion, maximumVersion);
+                "%s:%s(min: %s, recommended: %s, max: %s)%s",
+                productGroup,
+                productName,
+                minimumVersion,
+                recommendedVersion,
+                maximumVersion,
+                optional ? " optional" : "");
     }
 
     public String getProductGroup() {
@@ -229,11 +234,12 @@ public final class ProductDependency implements Serializable {
                 && productName.equals(that.productName)
                 && minimumVersion.equals(that.minimumVersion)
                 && Objects.equals(recommendedVersion, that.recommendedVersion)
-                && maximumVersion.equals(that.maximumVersion);
+                && maximumVersion.equals(that.maximumVersion)
+                && optional == that.optional;
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(productGroup, productName, minimumVersion, recommendedVersion, maximumVersion);
+        return Objects.hash(productGroup, productName, minimumVersion, recommendedVersion, maximumVersion, optional);
     }
 }
--- a/gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/ProductDependency.java
+++ b/gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/ProductDependency.java
@@ -55,13 +55,24 @@ public final class ProductDependency implements Serializable {
             String productName,
             String minimumVersion,
             String maximumVersion,
-            String recommendedVersion) {
+            String recommendedVersion,
+            boolean optional) {
         this.productGroup = productGroup;
         this.productName = productName;
         this.minimumVersion = minimumVersion;
         this.maximumVersion = maximumVersion;
         this.recommendedVersion = recommendedVersion;
+        this.optional = optional;
         isValid();
+    }
+
+    public ProductDependency(
+            String productGroup,
+            String productName,
+            String minimumVersion,
+            String maximumVersion,
+            String recommendedVersion) {
+        this(productGroup, productName, minimumVersion, maximumVersion, recommendedVersion, false);
     }
 
     /**
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/BaseDistributionExtension.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/BaseDistributionExtension.java
@@ -53,6 +53,7 @@ public class BaseDistributionExtension {
     private final Property<String> podName;
     private final Property<ProductType> productType;
     private final ListProperty<ProductDependency> productDependencies;
+    private final SetProperty<ProductId> optionalProductDependencies;
     private final SetProperty<ProductId> ignoredProductDependencies;
     private final ProviderFactory providerFactory;
     private final MapProperty<String, Object> manifestExtensions;
@@ -67,6 +68,7 @@ public class BaseDistributionExtension {
         podName = project.getObjects().property(String.class);
         productType = project.getObjects().property(ProductType.class);
         productDependencies = project.getObjects().listProperty(ProductDependency.class);
+        optionalProductDependencies = project.getObjects().setProperty(ProductId.class);
         ignoredProductDependencies = project.getObjects().setProperty(ProductId.class);
 
         serviceGroup.set(project.provider(() -> project.getGroup().toString()));
@@ -206,6 +208,18 @@ public class BaseDistributionExtension {
             }
             return dep;
         }));
+    }
+
+    public final Provider<Set<ProductId>> getOptionalProductDependencies() {
+        return optionalProductDependencies;
+    }
+
+    public final void optionalProductDependency(String productGroup, String productName) {
+        this.optionalProductDependencies.add(new ProductId(productGroup, productName));
+    }
+
+    public final void optionalProductDependency(String ignoredProductId) {
+        this.optionalProductDependencies.add(new ProductId(ignoredProductId));
     }
 
     public final Provider<Set<ProductId>> getIgnoredProductDependencies() {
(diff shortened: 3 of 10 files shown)


gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/ProductDependency.java
```diff
@@ -45,22 +45,36 @@
...
         this.productGroup = productGroup;
         this.productName = productName;
         this.minimumVersion = minimumVersion;
         this.maximumVersion = maximumVersion;
         this.recommendedVersion = recommendedVersion;
+        this.optional = optional;
         isValid();
     }
 
+    public ProductDependency(
```
nemines, line 69: I added a constructor instead of editing the existing one because it flagged further API breaks, but I'd be happy to remove it.

.palantir/revapi.yml
```diff
@@ -7,3 +7,9 @@ acceptedBreaks:
       new: "method org.gradle.api.provider.SetProperty<com.palantir.gradle.dist.ProductDependency>\
         \ com.palantir.gradle.dist.RecommendedProductDependenciesExtension::getRecommendedProductDependenciesProvider()"
       justification: "not really a break"
+  "4.26.0":
+    com.palantir.sls-packaging:gradle-recommended-product-dependencies:
+    - code: "java.class.defaultSerializationChanged"
```
CRogers, line 12: This should be fine as the java serialization is only used by Gradle, so all it will do is invalidate caches.

gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/ProductDependency.java
CRogers: This is the right approach, I believe `gradle-conjure` uses the above constructor.

gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/ProductDependency.java
```diff
@@ -194,6 +208,14 @@ public void setMaximumVersion(String maximumVersion) {
...
+        return optional;
+    }
+
+    public void setOptional(boolean optional) {
+        this.optional = optional;
+    }
+
     @Override
     public boolean equals(Object other) {
         if (this == other) {
```
CRogers, line 226: Need to update equals and hashCode

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/CreateManifestTask.java
```diff
@@ -98,6 +98,8 @@
 
     private final ListProperty<ProductDependency> productDependencies =
             getProject().getObjects().listProperty(ProductDependency.class);
+    private final SetProperty<ProductId> optionalProductIds =
```
CRogers, line 101: Should this be `ProductDependency` instead of `ProductId`? Optional product deps can have version info?

gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/ProductDependency.java
nemines: Done.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/CreateManifestTask.java
CRogers: Or maybe there are two cases:

1. A product dependency is discovered that you want to make optional
2. You want to manually add an optional product dependency

?

CRogers: Ah, silly me, you could just add the manual optional product dependency like so:

```gradle
productDependencies {
	// ...
	optional = true
}
```

CRogers: So am I right in thinking this is really this is just for discovered optional pdeps?

nemines: They can if they are defined as explicit product dependencies with an optional field.
They way I see the `optionalProductDependency` here is just as a flag to mark a discovered dependency as optional.

nemines: I was answering to the first comment above. Yes to the last question, that bit is just for discovered dependencies.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
```diff
@@ -285,6 +310,52 @@ class CreateManifestTaskIntegrationSpec extends GradleIntegrationSpec {
...
+        setup:
+        buildFile << """
+            dependencies {
+                runtime 'a:a:1.0'
+            }
+
+            tasks.createManifest {
+                optionalProductIds = [
+                    new com.palantir.gradle.dist.ProductId("group:name"), 
+                    new com.palantir.gradle.dist.ProductId("group:name2")
```
CRogers, line 323: Seems pretty grim that you need to do `new com.palantir.gradle.dist.ProductId(` - perhaps it should just be `String` instead so it's a bit easier to use? I realise the `ignoredProductIds` is like that, but a search on our internal code shows no one using `ignoredProductIds` except to set it to `[]`, so I'm not too worried about diverging with it.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/ProductDependencyLockFileTest.groovy
```diff
@@ -25,13 +25,13 @@ class ProductDependencyLockFileTest extends Specification {
...
                 new ProductDependency("com.palantir.product", "foo", "1.20.0", "1.x.x", null),
-                new ProductDependency("com.palantir.other", "bar", "0.2.0", "0.x.x", null)
+                new ProductDependency("com.palantir.other", "bar", "0.2.0", "0.x.x", null, true)
         ]
 
         then:
         ProductDependencyLockFile.asString(sample, [] as Set<ProductId>, '0.0.0') == """\
         # Run ./gradlew --write-locks to regenerate this file
-        com.palantir.other:bar (0.2.0, 0.x.x)
+        com.palantir.other:bar (0.2.0, 0.x.x) optional
```
CRogers, line 34: @ferozco Made a good point to me that this will break our pdeps-bot, which needs to be updated to handle optional deps. The file should be the same until someone adds an optional dep, then pdeps-bot will error out for that PR, so I don't think that should block this PR, but we need to fix that too.

gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/ProductDependency.java
```diff
@@ -45,22 +45,36 @@
     @JsonProperty("maximum-version")
     private String maximumVersion;
 
+    @JsonProperty("optional")
+    private boolean optional = false;
```
CRogers, line 49: I think we need a test that you can't set an `optional` pdep on a jar using the `recommended-product-dependencies` plugin (as we don't want optional pdeps on jars, at least not yet). I think we can add a check [after this line](https://github.com/palantir/sls-packaging/blob/138001775a7f3a16a539e4d8050efb32f88c46b9/gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/RecommendedProductDependenciesExtension.java#L43) that throws if the `optional` field is set.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
CRogers: (take a string but still use ProductId internally, that is)

nemines: I am not sure I get the wanted change. This complicates the logic of either the Extension or the CreateManifestTask for the conversion and storing of the properties, while this is just a test. In practice it's the `optionalProductDependency` that is used.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/ProductDependencyLockFileTest.groovy
nemines: I am not familiar with them, but I can look into fixing them if you can point me to them after this PR

gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/ProductDependency.java
nemines: Done

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
CRogers: Ah, missed the difference between `optionalProductIds` and `optionalProductDependency` - in that case please can we use `optionalProductDependency` in the test, just so that we're testing the API we expect people to use?

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
```diff
@@ -180,6 +180,26 @@ class CreateManifestTaskIntegrationSpec extends GradleIntegrationSpec {
         buildResult.output.contains('Encountered product dependency declaration that was also ignored')
     }
 
+    def 'throws if declared dependency is also optional'() {
+        setup:
+        buildFile << """
+            createManifest {
+                productDependencies = [
```
CRogers, line 187: ditto about using the expected API here (ie `productDependency {`), even if the other tests don't do this.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
nemines: It does not work if I use it in the test, I tried. I think it's because it is looking for it in the "createManifest" task while those functions are created in the extension

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
nemines: Ditto on above, it does not work directly. I believe this is the reason tests are written this way (I just copied and edited given existing templates)

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
CRogers: Ah, I see this is testing the generic `createManifest` task rather than the service/asset/recommended-pdeps plugins.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/BaseDistributionExtension.java
```diff
@@ -208,6 +210,18 @@ public final void productDependency(@DelegatesTo(ProductDependency.class) Closur
...
 
+    public final Provider<Set<ProductId>> getOptionalProductDependencies() {
+        return optionalProductDependencies;
+    }
+
+    public final void optionalProductDependency(String productGroup, String productName) {
+        this.optionalProductDependencies.add(new ProductId(productGroup, productName));
+    }
+
+    public final void optionalProductDependency(String ignoredProductId) {
```
CRogers: `ignoredProductId` -> `optionalProductId`

nemines: Done

CRogers:
:+1:

fawind (approved):

iamdanfox:
Also thoughts on including a little copy & pastable gradle example in the README? I've seen a bunch of people referring to the https://github.com/palantir/sls-packaging#product-dependencies section in #help-apollo channels

nemines:
@iamdanfox Good point, I filed a PR here: https://github.com/palantir/sls-packaging/pull/1076

PR #1076, opened 2021-02-15
Added optional dependencies to README

nemines:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -80,6 +80,7 @@ distribution {
         minimumVersion = "1.0.0"
         maximumVersion = "1.x.x"
         recommendedVersion = "1.2.1"
+        optional = false
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -90,27 +91,31 @@ sls-packaging also maintains a lockfile, `product-dependencies.lock`, which shou
...
 com.palantir.auth:auth-service (1.2.0, 1.6.x)
 com.palantir.storage:storage-service (3.56.0, 3.x.x)
-com.palantir.email:email-service (1.200.3, 2.x.x)
+com.palantir.email:email-service (1.200.3, 2.x.x) optional
 com.palantir.foo:foo-service ($projectVersion, 1.x.x)
 ```
 
 _The `$projectVersion` string is a placeholder that will appear if your repo publishes multiple services, and one of them depends on another.  The actual manifest will contain a concrete version._
 
+The suffix `optional` will be added for `optional = true` in the `productDependency` declaration. All dependencies are required by default.
```
nemines, line 100: I don't think it is this project job to describe what an optional dependency is, but I can add more details if needed.

fawind: Yes, I think it's alright to leave it like this. The term "optional dependency" is also decently self-explanatory.

fawind (approved):
Thanks for adding this!

PR #1078, opened 2021-02-19
correctly merge optional and non-optional dependencies

lwwmanning:
## Before this PR
If a dependency is discovered (non-optional) but declared explicitly as optional, the resulting merged dependency will always be non-optional, since the merge logic ignored the value of the `optional` field (and the default is `false`).

## After this PR

==COMMIT_MSG==
If a dependency is discovered (non-optional) but created explicitly as optional, it will correctly be categorized as `optional`.
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/ProductDependencyMerger.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/ProductDependencyMerger.java
@@ -82,12 +82,18 @@ public final class ProductDependencyMerger {
                 .filter(version -> satisfiesMaxVersion(maximumVersion, version))
                 .max(VersionComparator.INSTANCE);
 
+        // Since recommended--and thus, discovered--product dependencies cannot be optional, if we are merging
+        // optional and non-optional dependencies, then the one with `optional = true` must be an explicit product
+        // dependency and should always take precedence.
+        boolean optional = dep1.getOptional() || dep2.getOptional();
+
         ProductDependency result = new ProductDependency();
         result.setMinimumVersion(minimumVersion.toString());
         result.setMaximumVersion(maximumVersion.toString());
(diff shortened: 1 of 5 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/ProductDependencyMerger.java
```diff
@@ -82,12 +82,18 @@ public static ProductDependency merge(ProductDependency dep1, ProductDependency
                 .filter(version -> satisfiesMaxVersion(maximumVersion, version))
                 .max(VersionComparator.INSTANCE);
 
+        // Since recommended--and thus, discovered--product dependencies cannot be optional, if we are merging
+        // optional and non-optional dependencies, then the one with `optional = true` must be an explicit product
+        // dependency and should always take precedence.
```
CRogers: This seems odd to me - it kind of makes sense for the current world, but if we ever allow pdeps in jars to be optional then I would expect the opposite behaviour, where merging an optional and non-optional dep results in a non-optional dep.

Instead, to me it makes to do all the merging then make the deps that have been marked as optional actually optional. I think this would look like moving this logic:

https://github.com/palantir/sls-packaging/blob/56695cbb51765dd79e91d43aff3dba9a38cddc4e/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/CreateManifestTask.java#L254-L257

below the merging, in fact after that `discoverProductDependencies().forEach` block entirely.

lwwmanning: Cool, I buy that. In which case this logic should probably be `boolean optional = dep1.getOptional() && dep2.getOptional();` to be conservative (because that makes more sense than always `optional = false` even if both the inputs are optional).

CRogers:
:+1:

CRogers (commented):
:+1:

PR #1080, opened 2021-02-23
JavaServiceDistributionPlugin supports configuring java agents

carterkozak:
Early draft, pushing for early feedback before I invest too much time working in the wrong direction. Needs test coverage.

## Before this PR
Difficult to add java agents to distributions

## After this PR
==COMMIT_MSG==
JavaServiceDistributionPlugin supports configuring java agents
==COMMIT_MSG==

## Possible downsides?
Agents are scary

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
@@ -90,7 +90,7 @@ public final class JavaServiceDistributionPlugin implements Plugin<Project> {
         });
 
         Configuration runtimeClasspath = project.getConfigurations().getByName("runtimeClasspath");
-        Configuration javaAgentConfiguration = project.getConfigurations().maybeCreate("javaAgent");
+        Configuration javaAgentConfiguration = project.getConfigurations().create("javaAgent");
         javaAgentConfiguration.setTransitive(false);
 
         // Set default configuration to look for product dependencies to be runtimeClasspath
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
@@ -257,6 +257,7 @@ public final class JavaServiceDistributionPlugin implements Plugin<Project> {
             task.setGroup(JavaServiceDistributionPlugin.GROUP_NAME);
             task.setDescription("Runs the specified project using configured mainClass and with default args.");
             task.dependsOn("jar");
+            task.dependsOn(javaAgentConfiguration);
             if (GradleVersion.current().compareTo(GradleVersion.version("6.4")) < 0) {
                 task.doFirst(new Action<Task>() {
                     @Override
@@ -276,6 +277,9 @@ public final class JavaServiceDistributionPlugin implements Plugin<Project> {
                     jarTask.get().getArchiveFile().get(), p.getConfigurations().getByName("runtimeClasspath")));
             task.setArgs(distributionExtension.getArgs().get());
             task.setJvmArgs(ImmutableList.builder()
+                    .addAll(javaAgentConfiguration.getFiles().stream()
+                            .map(file -> "-javaagent:" + file.getAbsolutePath())
+                            .collect(Collectors.toList()))
                     .addAll(distributionExtension.getDefaultJvmOpts().get())
                     .addAll(distributionExtension.getGcJvmOptions().get())
                     .build());
@@ -287,9 +291,17 @@ public final class JavaServiceDistributionPlugin implements Plugin<Project> {
             // Set compression in constructor so that task output has the right name from the start.
             task.setCompression(Compression.GZIP);
             task.getArchiveExtension().set("sls.tgz");
-
-            task.dependsOn(startScripts, initScript, checkScript, yourkitAgent, yourkitLicense);
-            task.dependsOn(copyLauncherBinaries, launchConfigTask, manifest, manifestClassPathTask);
+            task.dependsOn(
+                    startScripts,
+                    initScript,
+                    checkScript,
+                    yourkitAgent,
+                    yourkitLicense,
+                    copyLauncherBinaries,
+                    launchConfigTask,
+                    manifest,
+                    manifestClassPathTask,
+                    javaAgentConfiguration);
         });
 
         project.afterEvaluate(_p -> launchConfigTask.configure(task -> {
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
@@ -89,9 +89,12 @@ public final class JavaServiceDistributionPlugin implements Plugin<Project> {
             }));
         });
 
+        Configuration runtimeClasspath = project.getConfigurations().getByName("runtimeClasspath");
+        Configuration javaAgentConfiguration = project.getConfigurations().maybeCreate("javaAgent");
+        javaAgentConfiguration.setTransitive(false);
+
         // Set default configuration to look for product dependencies to be runtimeClasspath
-        distributionExtension.setProductDependenciesConfig(
-                project.getConfigurations().getByName("runtimeClasspath"));
+        distributionExtension.setProductDependenciesConfig(runtimeClasspath);
 
         Provider<String> mainClassName = distributionExtension
                 .getMainClass()
@@ -290,6 +293,7 @@ public final class JavaServiceDistributionPlugin implements Plugin<Project> {
         });
 
         project.afterEvaluate(_p -> launchConfigTask.configure(task -> {
+            task.setJavaAgents(javaAgentConfiguration.getAsFileTree());
             if (distributionExtension.getEnableManifestClasspath().get()) {
                 task.setClasspath(manifestClassPathTask.get().getOutputs().getFiles());
             } else {
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
@@ -312,18 +312,14 @@ public final class JavaServiceDistributionPlugin implements Plugin<Project> {
 
         project.afterEvaluate(_p -> launchConfigTask.configure(task -> {
             task.getJavaAgents().setFrom(javaAgentConfiguration);
-            if (distributionExtension.getEnableManifestClasspath().get()) {
-                task.getClasspath().builtBy(manifestClassPathTask.get());
-                task.getClasspath()
-                        .from(manifestClassPathTask.get().getOutputs().getFiles());
-            } else {
-                task.getClasspath().builtBy(jarTask.get());
-                task.getClasspath()
-                        .from(jarTask.get()
-                                .getOutputs()
-                                .getFiles()
-                                .plus(distributionExtension.getProductDependenciesConfig()));
-            }
+            task.getClasspath()
+                    .from(
+                            distributionExtension.getEnableManifestClasspath().get()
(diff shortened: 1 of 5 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -148,10 +149,19 @@ public final FileCollection getClasspath() {
         return classpath;
     }
 
+    @InputFiles
+    public final FileCollection getJavaAgents() {
```
ferozco: I would use a `ConfigurableFileCollection` from Gradle. It is more idiomatic, lets you configure it directly (instead of in an afterEvaluate) and allows you to avoid the setter. You could even use Gradle's magic property injection instead of defining a field:

```
@InputFiles
abstract ConfigurableFileCollection getJavaAgents();
```

carterkozak: Should we update the classpath to do the same?

ferozco: ya we could honestly clean up all of the fields here to get rid of a bunch of boilerplate. I didn't want to slam you with unrelated changes though

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
```diff
@@ -273,6 +277,9 @@ public void execute(Task _task) {
                     jarTask.get().getArchiveFile().get(), p.getConfigurations().getByName("runtimeClasspath")));
             task.setArgs(distributionExtension.getArgs().get());
             task.setJvmArgs(ImmutableList.builder()
+                    .addAll(javaAgentConfiguration.getFiles().stream()
```
ferozco: this will resolve the configuration at configuration time, which will make GCV sad. We will need to set this slightly differently. I believe the task exposes a way for you to register an args provider which is only evaluated at execution time

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
```diff
@@ -89,9 +89,12 @@ public void apply(Project project) {
             }));
         });
 
+        Configuration runtimeClasspath = project.getConfigurations().getByName("runtimeClasspath");
+        Configuration javaAgentConfiguration = project.getConfigurations().maybeCreate("javaAgent");
```
ferozco: be bold! create or fail

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
carterkozak: Just discovered this testing with log-receiver -- I've added a JavaExec doFirst action which appears to do what we want. lmk if there's a smarter way to go about it though.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
carterkozak: Fair enough!

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
ferozco: let me open this up in Intellij and maybe I can find something!

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
carterkozak: No worries, I implemented this one by pattern matching against the classpath config.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
```diff
@@ -264,6 +268,21 @@ public void execute(Task _task) {
             } else {
                 task.getMainClass().set(mainClassName);
             }
+            task.doFirst(new Action<Task>() {
```
ferozco: I think we can avoid the `doFirst` by using https://docs.gradle.org/current/javadoc/org/gradle/process/JavaForkOptions.html#getJvmArgumentProviders--

ferozco:
@carterkozak change looks sane to me!

carterkozak:
Thanks, @ferozco! I'll put together some test coverage tomorrow.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
```diff
@@ -284,19 +297,32 @@ public void execute(Task _task) {
...
+                    manifest,
+                    manifestClassPathTask,
+                    javaAgentConfiguration);
         });
 
         project.afterEvaluate(_p -> launchConfigTask.configure(task -> {
+            task.getJavaAgents().setFrom(javaAgentConfiguration);
             if (distributionExtension.getEnableManifestClasspath().get()) {
-                task.setClasspath(manifestClassPathTask.get().getOutputs().getFiles());
+                task.getClasspath().builtBy(manifestClassPathTask.get());
```
ferozco: why do we need to set the classPath to be builtBy?

carterkozak: No idea, it seemed like the right thing to do to source the objects, I assumed that helped build a cleaner task dependency graph. I can remove these if that's not what it's for.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
```diff
@@ -254,6 +258,19 @@ public void execute(Task _task) {
             task.setGroup(JavaServiceDistributionPlugin.GROUP_NAME);
             task.setDescription("Runs the specified project using configured mainClass and with default args.");
             task.dependsOn("jar");
+            task.dependsOn(javaAgentConfiguration);
+            task.jvmArgs(ImmutableList.<String>builder()
+                    .addAll(distributionExtension.getDefaultJvmOpts().get())
+                    .addAll(distributionExtension.getGcJvmOptions().get())
+                    .build());
+            task.getJvmArgumentProviders().add(new CommandLineArgumentProvider() {
```
CRogers, line 271: Hmm, I didn't know you could use `jvmArgs` and `getJvmArgumentProviders` together - I thought they overwrote each other (perhaps they would if they were declared with `jvmArgs` last?)

carterkozak: huh, I'll update the test to verify gc args are present

carterkozak: hmm, this is the run task, not sure we test it. Might as well combine the two into a single jCommandLineArgumentProvider

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
```diff
@@ -89,9 +90,12 @@ public void apply(Project project) {
             }));
         });
 
+        Configuration runtimeClasspath = project.getConfigurations().getByName("runtimeClasspath");
+        Configuration javaAgentConfiguration = project.getConfigurations().create("javaAgent");
+        javaAgentConfiguration.setTransitive(false);
```
CRogers: It's generally best to configure configurations in an action, so

```java
Configuration javaAgentConfiguration = project.getConfigurations().create("javaAgent", javaAgent -> javaAgent.setTransitive(false));
```

Probably not a huge deal here, but in general this can allow Gradle to avoid doing work when it doesn't need to.

carterkozak: Is this one of those things that needs to be an anonymous class? We appear to do that for some other Action types.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -170,7 +167,8 @@ public final void createConfig() throws IOException {
                         .serviceName(serviceName.get())
                         .javaHome(javaHome.getOrElse(""))
                         .args(args.get())
-                        .classpath(relativizeToServiceLibDirectory(classpath))
+                        .classpath(relativizeToServiceLibDirectory(getClasspath()))
+                        .addAllJvmOpts(javaAgentArgs())
```
CRogers, line 171: Just to check, because I'm not sure, we had a problem with deployments overriding particular jvm args in the past, these here are in a place where they won't get overridden, right (as we want agents - or at least your new agent - to always be applied)?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
```diff
@@ -89,9 +91,20 @@ public void apply(Project project) {
             }));
         });
 
+        Configuration runtimeClasspath = project.getConfigurations().getByName("runtimeClasspath");
+        Configuration javaAgentConfiguration = project.getConfigurations()
+                .create("javaAgent", new Action<Configuration>() {
```
CRogers, line 96: oof I always forget how you can't use lambdas.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
carterkozak: Yep, much like the classpath I don't want to allow users to modify agents after build time. If it's included in the distribution we should be able to rely on it always executing.

carterkozak:
oof, builds are sad with `org.gradle.api.UncheckedIOException: de.undercouch.gradle.tasks.download.org.apache.http.client.ClientProtocolException: Not Found (HTTP status code: 404, URL: https://www.yourkit.com/download/YourKit-JavaProfiler-2020.9-b416.zip)`

Looks like I may need to retrigger later

CRogers:
ugh, that has happened before in the past, yourkit removes old versions from their website. I thought I had done something to fix this in the past, like an excavator to update it.

CRogers:
I think we were also debating whether or not we needed to package yourkit anymore too.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
CRogers: I think if you use these methods then you can avoid having to set `dependsOn` everywhere as Gradle can work it what it need to run to build these files.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
CRogers: We have an error prone rule that will warn you (or perhaps a the gradle plugin validation task?). I can never remember the particular cases so I just use a lambda until it complains at me.

ferozco: I believe that you only want an Action when doing a `doFirst`/`doLast` so that the task is properly cached.

I don't think there is performance benefit for configuring a Configuration using a lambda since the configuration is always eagerly created. It is definitely true that for lazy task (i.e. `TaskProvider<?>`) that you want to configure them using a lambda instead of calling `get()` and then configuring the return value

CRogers (approved):

PR #1084, opened 2021-02-25
The new `javaAgent` configuration is no longer non-transitive

carterkozak:
This works around a GCV bug which fails to apply constraints to
non-transitive configurations.
See https://github.com/palantir/gradle-consistent-versions/issues/677

==COMMIT_MSG==
The new `javaAgent` configuration is no longer non-transitive
==COMMIT_MSG==

## Possible downsides?
Agents shouldn't ever have dependencies, however it's somewhat fragile to rely on that being the case. We may end up with more `-javaagent:<path/to/*.jar>` lines args than we want, preventing the JVM from starting. I've added validation to prevent this scenario from shipping.

### DIFF
--- a/changelog/@unreleased/pr-1084.v2.yml
+++ b/changelog/@unreleased/pr-1084.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: The new `javaAgent` configuration is no longer non-transitive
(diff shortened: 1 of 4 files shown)


fawind (approved):
Makes sense!

PR #1085, opened 2021-02-26
Allow more lenient minimum version compares for intra-module product …

esword:
…dependencies.

## Before this PR
We have a very large mono-repo with an ecosystem of related products that use recommended product dependencies between them. Due to the way our build system works, we cannot ever lock in a minimum version for these because we get conflicts with the product-dependencies.lock files. There is no way for us to know what the specific version should be at "commit" time.

This change also addresses an issue where you would get "lock file is out of date" messages when producing builds from branches or that had not been rebased prior to merging.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

## Possible downsides?
None. It's perfect.

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/BaseDistributionExtension.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/BaseDistributionExtension.java
@@ -52,6 +52,7 @@ public class BaseDistributionExtension {
     private final Property<String> serviceName;
     private final Property<String> podName;
     private final Property<ProductType> productType;
+    private final Property<Boolean> lenientInternalLocks;
     private final ListProperty<ProductDependency> productDependencies;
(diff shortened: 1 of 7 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/BaseDistributionExtension.java
```diff
@@ -133,6 +137,14 @@ public final void setProductType(ProductType productType) {
         this.productType.set(productType);
     }
 
+    public final Provider<Boolean> getLenientInternalLocks() {
```
ferozco: you can expose a `Property<Boolean>` avoid the setter and allow for lazy for configuration

ferozco:
👍 LGTM thanks @esword

carterkozak (approved):
lgtm, thanks!

PR #1095, opened 2021-03-15
Improvement: Add workaround to JDK-8253566 for java 15 runtimes

robert3005:
## After this PR

==COMMIT_MSG==
Add workaround for JDK-8253566 by default when target runtime is java 15
==COMMIT_MSG==

even though this has been fixed in java 15.0.2 we can't guarantee that we aren't running with 15.0.0 or 15.0.1

### DIFF
--- a/changelog/@unreleased/pr-1095.v2.yml
+++ b/changelog/@unreleased/pr-1095.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add workaround for JDK-8253566 by default when target runtime is java 15
+  links:
+  - https://github.com/palantir/sls-packaging/pull/1095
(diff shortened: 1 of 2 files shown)


carterkozak (dismissed):
We decided to block java 15 rollout until this was resolved rather than hard-coding experimental options. Has something changed? I'd be okay with this if we could detect impacted versions of the jre, but using `UnlockDiagnosticVMOptions` everywhere doesn't exactly inspire confidence.

robert3005:
We already run this on one of our services (the one that exihibited the bug) in production, hence why I think this is ok to do.

tpetracca (approved):
Discussed quite a bit offline. We're going to move forward with this.
We're going to re-assess in May if its feasible to unwinde this given the java15 jdk infra rollout across the fleet at that point in time.

PR #1144, opened 2021-05-20
allow discovered dependencies to be set

esword:
if none are set then the default mechanism of scanning jars kicks in.

## Before this PR
We have an external tool that can modify the list of discovered (recommended) dependencies.  There was no way to override that process in the CreateManifest task

## After this PR

==COMMIT_MSG==
allow discovered dependencies to be set rather than processed in CreateManifestTask
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/BaseDistributionExtension.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/BaseDistributionExtension.java
@@ -53,6 +53,7 @@ public class BaseDistributionExtension {
     private final Property<String> podName;
     private final Property<ProductType> productType;
     private final ListProperty<ProductDependency> productDependencies;
+    private final ListProperty<ProductDependency> discoveredDependencies;
     private final SetProperty<ProductId> optionalProductDependencies;
     private final SetProperty<ProductId> ignoredProductDependencies;
     private final ProviderFactory providerFactory;
@@ -68,6 +69,7 @@ public class BaseDistributionExtension {
         podName = project.getObjects().property(String.class);
         productType = project.getObjects().property(ProductType.class);
         productDependencies = project.getObjects().listProperty(ProductDependency.class);
+        discoveredDependencies = project.getObjects().listProperty(ProductDependency.class);
         optionalProductDependencies = project.getObjects().setProperty(ProductId.class);
         ignoredProductDependencies = project.getObjects().setProperty(ProductId.class);
 
@@ -208,6 +210,14 @@ public class BaseDistributionExtension {
             }
             return dep;
         }));
+    }
+
+    public final void discoveredProductDependency(ProductDependency productDependency) {
+        discoveredDependencies.add(productDependency);
+    }
+
+    public final ListProperty<ProductDependency> getDiscoveredProductDependencies() {
+        return discoveredDependencies;
     }
 
     public final Provider<Set<ProductId>> getOptionalProductDependencies() {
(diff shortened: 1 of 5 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/BaseDistributionExtension.java
```diff
@@ -68,6 +69,7 @@ public BaseDistributionExtension(Project project) {
         podName = project.getObjects().property(String.class);
         productType = project.getObjects().property(ProductType.class);
         productDependencies = project.getObjects().listProperty(ProductDependency.class);
+        discoveredDependencies = project.getObjects().listProperty(ProductDependency.class);
```
ferozco, line 72: why does this have to be different from the existing productDependencies list?

ferozco (commented):
I think you can accomplish what you want without any changes. You can see an example here where we discover product dependencies separately from the createManifestTask and have them injected into the final manifest: https://github.com/gradlets/gradle-typescript/blob/develop/gradle-typescript/src/main/java/com/gradlets/gradle/typescript/pdeps/Packaging.java#L57-L83

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/BaseDistributionExtension.java
esword: the purpose of this is to be set from another tool/task rather than manually.  That task takes the recommended dependencies (doing a similar jar scan as the CreateManifestTask) and may alter the minimum versions to be earlier than those set by default in the jar manifest's recommended-dependencies block.  If we do not override the jar-scanning behavior of the CreateManifestTask the merging functionality will select the higher minimum version from the set of PDs that the CreateManifestTask finds.

ferozco: ok you want to completely replace the system that CreateManifest task uses internally. You could do that without an API change by configuring createManfiest task to scan an empty configuration and then providing your own product dependencies through the extension like in the example I linked

esword: That is a possibility, but it seems a bit obtuse.  Creating and setting an empty configuration to indirectly disable the default behavior has a code smell.  I'd rather have the process have an explicit way of changing behavior rather than attempting subvert the current one.

Ideally, one could even think of this as a case where a provider pattern could be used.  Rather than all the merging logic happening in the task itself, the task is given a ProductDependenciesProvider interface.  The default one would follow the current logic, while other implementations could replace said logic.  

I wanted to make this change smaller, so did not go that route, but I'm open to it if other's think it's a good idea.

ferozco: To me it seems clear that an empty configuration would result in nothing being discovered and has the nice benefit that it avoids adding another branch in the logic of the task but could see it being a taste thing.

If we did want to make a bigger change I would suggest pulling out the discovery logic into a separate task, outputting a file with the resulting product dependencies and then plumbing those dependencies into `createManifest`. It would be a nice improvement since you could separately cache product dependency discovery from writing the manifest

esword: that would be a nice improvement.  I can put that together.  Do we need to worry at all about this breaking any projects that customize the CreateManifest task at all?  I searched for mentions of that task using SourceGraph and the only project that does anything to CreateManifest other than add to a task-dependency (which wouldn't break) is pgdev.  I can handle fixing that when I bump the upgrade.

esword:
abandoning in favor of https://github.com/palantir/sls-packaging/pull/1146

PR #1146, opened 2021-05-25
move product dep calcs into own task

esword:
makes it easier to both cache and override the logic

## After this PR

==COMMIT_MSG==
move product dep calcs into own task
==COMMIT_MSG==


gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskTest.groovy
```diff
@@ -45,22 +45,4 @@ class CreateManifestTaskTest extends ProjectSpec {
         Exception exception = thrown()
         exception.message.contains("Project version must be a valid SLS version: 1.0.0foo")
     }
-
```
esword, line 48: I moved this into the integration test

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
```diff
@@ -200,6 +195,26 @@ class CreateManifestTaskIntegrationSpec extends GradleIntegrationSpec {
         buildResult.output.contains('Encountered product dependency declaration that was also declared as optional')
     }
 
+    def "throws if user declares dependency on the same product"() {
```
esword: moved from the CreateManifestTaskTest

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/CreateManifestTask.java
```diff
@@ -97,53 +80,14 @@
...
-                            return Stream.of(
-                                    dependencyProject.getTasks().withType(ConfigureProductDependenciesTask.class));
-                        }
-                        return Stream.empty();
-                    })
-                    .forEach(emptyFileCollection::builtBy);
-            return emptyFileCollection;
-        });
-    }
+    public CreateManifestTask() {}
```
ferozco: nit: remove unnecessary constructor

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/CreateManifestTask.java
```diff
@@ -166,37 +110,14 @@ public CreateManifestTask() {
...
-                .map(ResolvedComponentResult::getId)
-                .map(ComponentIdentifier::getDisplayName)
-                .collect(Collectors.toSet());
-    }
-
-    final void setConfiguration(Provider<Configuration> config) {
-        this.productDependenciesConfig.set(config);
+    @InputFile
+    @PathSensitive(PathSensitivity.RELATIVE)
+    public final RegularFileProperty getProductDependenciesFile() {
```
ferozco, line 112: you could conver this to an anonymous class/method and save on boilerplate

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
```diff
@@ -0,0 +1,347 @@
...
+import org.gradle.api.provider.Provider;
+import org.gradle.api.provider.SetProperty;
+import org.gradle.api.tasks.CacheableTask;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.OutputFile;
+import org.gradle.api.tasks.TaskAction;
+import org.gradle.jvm.tasks.Jar;
+
+@CacheableTask
+public class ResolveProductDependenciesTask extends DefaultTask {
```
ferozco: make this an abstract class and all of the inputs/ouputs can just be anonymous methods which gradle will inject implementations for

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
```diff
@@ -0,0 +1,347 @@
...
+    private final SetProperty<ProductId> ignoredProductIds =
+            getProject().getObjects().setProperty(ProductId.class);
+    private final Property<Configuration> productDependenciesConfig =
+            getProject().getObjects().property(Configuration.class);
+
+    private final RegularFileProperty outputFile = getProject().getObjects().fileProperty();
+
+    public ResolveProductDependenciesTask() {
+        dependsOn(otherProjectProductDependenciesTasks());
+        outputFile.convention(() -> new File(getTemporaryDir(), "endpoint-minimum-versions.json"));
```
ferozco: avoid using conventions in favour of property/provider. You should be able to just set the value of the property

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
```diff
@@ -0,0 +1,347 @@
...
+        return inRepoProductIds;
+    }
+
+    /**
+     * Contents of the given configuration.  Cannot list the configuration itself as an input property because the
+     * caching calculations attempt to resolve the configuration at config time.  This can lead to an error for
+     * configs that connot be resolved directly at that stage.  Caching and up-to-date calcs thus use this property.
+     */
+    @Input
+    final Set<String> getProductDependenciesConfig() {
```
ferozco, line 154: can we move this to use the artifactTransfomer pattern that we use for diagnostics?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
```diff
@@ -0,0 +1,347 @@
...
+                    } catch (IOException | IllegalArgumentException e) {
+                        log.debug(
+                                "Failed to load product dependency for artifact '{}', file '{}', '{}'",
+                                artifactName,
+                                artifact,
+                                e);
+                        return Stream.empty();
+                    }
+                })
+                .filter(this::isNotSelfProductDependency)
```
ferozco: we should probably do the filtering inside of createManifest task. It will also allow us to remove a few inputs into this task

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
```diff
@@ -27,6 +29,16 @@ import spock.lang.Unroll
 
 class CreateManifestTaskIntegrationSpec extends GradleIntegrationSpec {
```
ferozco: should we pull the tests for the product dependency discovery out into a seperate test suite? it might be easier to mock and verify the behaviour we want

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
esword: why? I like conventions in cases where you want to know if set. I don't need it here necessarily, but curious as to why.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/CreateManifestTask.java
esword: not sure I understand - how exactly do you mean? and could we do that with the caching-related annotationsn

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestTaskIntegrationSpec.groovy
esword: yes, in the process of doing so

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
esword: eh - we still need the serviceName and serviceGroup parameters in the resolving task because we check and thow error if declare self reference explicitly.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
esword: I see how it would be done (neat!), but I am nervous about time it will take and getting it right. All the current changes are pretty much just refactoring with the one new additional feature of being able to inject the discovered product dependencies. Can we please hold off on that improvement so that the rest of the change can move forward? I’ll circle back to it in the next couple of weeks as we are in a “soaking” period with the new features in some projects.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/SlsManifest.java
```diff
@@ -24,7 +24,7 @@
 
 @Value.Immutable
 @JsonSerialize(as = ImmutableSlsManifest.class)
-@JsonDeserialize(as = ImmutableSlsManifest.class)
+@JsonDeserialize(builder = SlsManifest.Builder.class)
```
ferozco, line 27: why was this changed?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
```diff
@@ -0,0 +1,350 @@
...
+                .sorted(Comparator.comparing(ProductDependency::getProductGroup)
+                        .thenComparing(ProductDependency::getProductName))
+                .collect(ImmutableList.toImmutableList());
+
+        Serializations.writeProductDependencyReport(
+                ProductDependencyReport.of(productDeps),
+                getOutputFile().getAsFile().get());
+    }
+
+    private Set<ProductDependency> dedupDiscoveredProductDependencies() {
```
ferozco, line 242: I think we want all of this logic to occur inside of CreateManifestTask so that regardless of how product dependencies are received we don't add self-dependencies or duplicate ones

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/AbstractTaskSpec.groovy
```diff
@@ -0,0 +1,95 @@
...
+
+package com.palantir.gradle.dist.tasks
+
+import java.nio.file.Files
+import com.palantir.gradle.dist.GradleIntegrationSpec;
+import java.nio.file.StandardCopyOption;
+import nebula.test.dependencies.DependencyGraph;
+import nebula.test.dependencies.GradleDependencyGenerator;
+
+abstract class AbstractTaskSpec extends GradleIntegrationSpec {
```
ferozco: I would really prefer avoiding a test class hierarchy and instead pull things out into shared utilities. Otherwise it can be very confusing to understand the setup of any given test and inadvertently break the tests assumptions

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/CreateManifestTask.java
```diff
@@ -515,28 +259,30 @@ private void validateProjectVersion() {
     }
 
     public static TaskProvider<CreateManifestTask> createManifestTask(Project project, BaseDistributionExtension ext) {
+        Provider<Set<ProductId>> provider = project.provider(
```
fawind: nit while we iterate on this: Would give this a more descriptive name, e.g. `inRepoProductIds`.

Same for the parameter name in: https://github.com/palantir/sls-packaging/blob/cb76b23c0d540da4725d566ab015e52e690e199e/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java#L81-L82

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestIntegrationSpec.groovy
```diff
@@ -0,0 +1,247 @@
...
+
+class CreateManifestIntegrationSpec extends AbstractTaskSpec {
+
+    File manifestFile
+
+    def setup() {
+        manifestFile = new File(projectDir, 'build/deployment/manifest.yml')
+    }
+
+    def 'fails if lockfile is not up to date'() {
```
ferozco, line 37: instead of creating dependencies (and implicitly testing the pdep discovery) why don't you wire up the input to the task with the pdeps you want. Even better you could refactor the createManfest task so that it can be tested using a unit test since it no longer really relies on gradle details

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
ferozco: conventions are generally unsafe and it is considered best practice to use properties instead since they provide a richer API to consume and modify

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/SlsManifest.java
esword: Because I centralized read/write calls into the Serializations class and added a reader metthod for SlsManifest.  Jackson throws exception without the builder - 
Caused by: java.lang.IllegalArgumentException: Cannot find a deserializer for non-concrete Map type [map type; class com.google.common.collect.ImmutableMap, [simple type, class java.lang.String] -> [simple type, class java.lang.Object]]
	at com.fasterxml.jackson.databind.deser.BasicDeserializerFactory.createMapDeserializer(BasicDeserializerFactory.java:1358)
	at com.fasterxml.jackson.databind.deser.DeserializerCache._createDeserializer2(DeserializerCache.java:387)
	at com.fasterxml.jackson.databind.deser.DeserializerCache._createDeserializer(DeserializerCache.java:349)
	at com.fasterxml.jackson.databind.deser.DeserializerCache._createAndCache2(DeserializerCache.java:264)
	... 13 more

It works in other test classes (see JavaServiceDistributionPluginTests) because they add a GuavaModule() to the object mapper configuration, but I didn't want to alter the production one.

ferozco: I think you should just add the guava module to the object mapper then :)

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
esword: Still confused, sorry. This isn't setting a Convention object (which have been deprecated), but the default value for the property per docs here -https://docs.gradle.org/current/userguide/lazy_configuration.html#applying_conventions.  It's like setting a value, but you can tell if something else changes it.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
esword: that really feels like we're splitting the logic and constraints for the same unit of work into two places. Why pull stuff out to another task if we're going to still have to have code in CreateManifest to double check things?

ferozco: we pulled it out so that it was easy to replace/provide an additional implementation. It seems reasonable for createManifest to ensure that the preconditions it requires for writing to manifest/lock file are met/fixing things if they are not

ferozco: Otherwise you'd have to implement the same deduplication/filtering logic in each seperate mechanism for discovering dependencies

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestIntegrationSpec.groovy
esword: I didn't create any of these tests - I just split them out from the original integration spec and cleaned them up a bit. I also find that integration tests are better for gradle tasks because it runs them in an environment that is closer to how they are actually used. Often the internal logic of a task seems to be fine in an isolated test, but the actual behavior of the task is not what is expected or fails outright when run for real.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/AbstractTaskSpec.groovy
esword: replaced with delegate fixture

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/CreateManifestTask.java
esword: done

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/SlsManifest.java
esword: not that I'm an expert on jackson deserialization, but if the tool provides a built in way to specify a builder, and the instances are supposed to be constructed with a builder, shouldn't we, uh, ya know, use a builder?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
esword: the way the resolve task is written, if someone uses it and overrides the discovered dependencies, it will still provide errors and warnings about duplicates and self-dependencies. So we have the same protections (and useful messages) as now. 

I could see the createManifest task doing a final validation on the provided list and throwing errors if there is a self-reference or a duplicate. That way if someone workflow doesn't use the resolve task at all, the system will fail fast and force them to deconflict things with conscious decisions rather than by hiding potential errors. 

Good?

ferozco: ya thats great thanks!

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/SlsManifest.java
ferozco: I think its just a gotcha to remember specify the builder especially since it would fail at runtime. Its pretty standard for people to do `@JsonDeserialize(as = Immutable..)` and basically all of our object mappers are configured with the guava module. don't care too much but just seems easier to make the change once in the mapper

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
ferozco: ok, lets keep it as is

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestIntegrationSpec.groovy
ferozco: ok, it seemed like pushing to unit test was a nice way to get rid of the test class hierarchy and speed up tests but its fine this way too

ferozco:
👍

fawind (approved):
Another +1 for the pdep validation in https://github.com/palantir/sls-packaging/pull/1146/commits/42c3154fa96f058019a3c611280d77f438f0709b

PR #1147, opened 2021-05-27
[WIP] Enable container support.

jkozlowski:
## Before this PR

Developers have to manually specify heap size when running in containers.

## After this PR
==COMMIT_MSG==
If containerSupport: true AND running inside container, java heap is configured automatically.
==COMMIT_MSG==


jkozlowski:
Decided against this approach.

PR #1150, opened 2021-06-04
resolve race condition with jar manifest manipulation

esword:
The minimum required change is for the task to dependOn the configuration.  The rest of the
changes simplify the jar-scanning logic since we can now assume all jars are present, regardless
of whether the dependency is in the same project or external.

## Before this PR
The ResolveProductDependenciesTask can run before or during tasks that modify the contents of the Manifest produced by the Jar task of dependent projects.  This can lead to ConcurrentModificationExceptions.  This was a potential problem even when the logic was in the CreateManifest task.  It is unclear why it did not manifest.

## After this PR

==COMMIT_MSG==
resolve race condition with jar manifest manipulation
==COMMIT_MSG==

## Possible downsides?
Adding the dependency on the configuration does mean that incoming dependencies will be resolved (i.e. the jars built).  That is true regardless of whether the logic to change how the artifact is scanned is changed or not.  This is normally not an issue since createManifest is generally triggered by the dist target and everything needs to be built anyway.  It is different when only --write-locks is run and there are no recommended product dependencies.  We may want to find an optimization for that case if it becomes significant.

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
@@ -31,12 +31,12 @@ import com.palantir.gradle.dist.Serializations;
 import java.io.File;
 import java.io.IOException;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 import java.util.jar.Attributes;
 import java.util.jar.Manifest;
@@ -93,6 +93,9 @@ public abstract class ResolveProductDependenciesTask extends DefaultTask {
                     task.getOptionalProductIds().set(ext.getOptionalProductDependencies());
                     task.getIgnoredProductIds().set(ext.getIgnoredProductDependencies());
                     task.getInRepoProductIds().set(provider);
+
+                    // Need the dependsOn because the configuration is not set as an Input property and
+                    // we need the artifacts in it to be resolved.
                     task.dependsOn(configProvider);
                 });
         return depTask;
@@ -270,24 +273,22 @@ public abstract class ResolveProductDependenciesTask extends DefaultTask {
 
         return artifactStream
                 .map(ResolveProductDependenciesTask::getProductDependenciesFromArtifact)
-                .filter(Optional::isPresent)
-                .map(Optional::get)
                 .flatMap(Collection::stream)
                 .distinct()
                 .collect(Collectors.toList());
     }
 
-    static Optional<Collection<ProductDependency>> getProductDependenciesFromArtifact(ResolvedArtifact artifact) {
+    private static Collection<ProductDependency> getProductDependenciesFromArtifact(ResolvedArtifact artifact) {
         File jarFile = artifact.getFile();
         try (ZipFile zipFile = new ZipFile(jarFile)) {
             String artifactName = artifact.getId().getDisplayName();
             ZipEntry manifestEntry = zipFile.getEntry("META-INF/MANIFEST.MF");
             if (manifestEntry == null) {
-                return Optional.empty();
+                return Collections.emptySet();
             }
             Attributes attrs = new Manifest(zipFile.getInputStream(manifestEntry)).getMainAttributes();
             if (!attrs.containsKey(RecommendedProductDependencies.SLS_RECOMMENDED_PRODUCT_DEPS_ATTRIBUTE)) {
-                return Optional.empty();
+                return Collections.emptySet();
             }
 
             Set<ProductDependency> recommendedDeps = Serializations.jsonMapper
@@ -296,7 +297,7 @@ public abstract class ResolveProductDependenciesTask extends DefaultTask {
                             RecommendedProductDependencies.class)
                     .recommendedProductDependencies();
             if (recommendedDeps.isEmpty()) {
-                return Optional.empty();
+                return Collections.emptySet();
             }
             log.info(
                     "Product dependency recommendation made by artifact '{}', file '{}', "
@@ -304,10 +305,10 @@ public abstract class ResolveProductDependenciesTask extends DefaultTask {
                     artifactName,
                     artifact,
                     recommendedDeps);
-            return Optional.of(recommendedDeps);
+            return recommendedDeps;
         } catch (IOException e) {
             log.warn("Failed to load product dependency for artifact '{}', file '{}', '{}'", artifact, jarFile, e);
-            return Optional.empty();
+            return Collections.emptySet();
         }
     }
 
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
@@ -20,13 +20,11 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.palantir.gradle.dist.BaseDistributionExtension;
-import com.palantir.gradle.dist.ConfigureProductDependenciesTask;
 import com.palantir.gradle.dist.ProductDependency;
 import com.palantir.gradle.dist.ProductDependencyMerger;
 import com.palantir.gradle.dist.ProductDependencyReport;
 import com.palantir.gradle.dist.ProductId;
 import com.palantir.gradle.dist.RecommendedProductDependencies;
-import com.palantir.gradle.dist.RecommendedProductDependenciesPlugin;
 import com.palantir.gradle.dist.Serializations;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
```diff
@@ -82,13 +85,15 @@ public ResolveProductDependenciesTask() {
...
+                    Provider<Configuration> configProvider = project.provider(ext::getProductDependenciesConfig);
                     task.getServiceName().set(ext.getDistributionServiceName());
                     task.getServiceGroup().set(ext.getDistributionServiceGroup());
                     task.getDeclaredProductDependencies().set(ext.getAllProductDependencies());
-                    task.setConfiguration(project.provider(ext::getProductDependenciesConfig));
+                    task.setConfiguration(configProvider);
                     task.getOptionalProductIds().set(ext.getOptionalProductDependencies());
                     task.getIgnoredProductIds().set(ext.getIgnoredProductDependencies());
                     task.getInRepoProductIds().set(provider);
+                    task.dependsOn(configProvider);
```
carterkozak, line 87: Based on reviews from @CRogers I had thought that the Provider would automatically set task dependencies. Is that not the case?

esword: Good question.  A property/provider that is the output of task A will create a dependency if it is set to an input property of task B.  However, because this provider is not associated with a task, I do not know if the same logic applies.  I can say the behavior was different once I added the explicit dependsOn statement - the upstream jar tasks were run.

fawind, line 87: Can we add a small comment here on why we need this dep?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
```diff
@@ -254,93 +259,56 @@ final void computeProductDependencies() throws IOException {
...
-                                .getValue(RecommendedProductDependencies.SLS_RECOMMENDED_PRODUCT_DEPS_KEY));
-                    }
-                    if (!pdeps.isPresent()) {
-                        log.debug(
-                                "No product dependency found for artifact '{}', file '{}'",
-                                artifactName,
-                                artifact.getFile());
-                        return Stream.empty();
-                    }
+    static Optional<Collection<ProductDependency>> getProductDependenciesFromArtifact(ResolvedArtifact artifact) {
```
fawind: Why not return an empty collection instead of `Optional.empty()`? That would simplify your stream logic above.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
fawind: That might be a stupid question, but what does it actually mean to depend on a provider? The provider will be evaluated before the task runs? How is that different from evaluating the provider passed through as a task input?

esword: actually, the right answer to this may be that the Configuration is not listed as an Input property to the task.  See - https://github.com/palantir/sls-packaging/blob/856fba273b088d83731675895923f7f39cfb9968/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java#L153-L164

While I didn't set things up this way here (predated me), I have encountered this issue before with in similar situations.  I do not know if it is still a problem of if newer versions of gradle have some way to get around it.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
esword: doh.  yes, that is cleaner.  That's what I get for reusing code and not looking closely at mods.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java
esword: More data - The task has this method to try to enforce the dependencies between the tasks as well - https://github.com/palantir/sls-packaging/blob/856fba273b088d83731675895923f7f39cfb9968/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ResolveProductDependenciesTask.java#L106-L130

This was added with this commit - 
https://github.com/palantir/sls-packaging/commit/3649f1b38a18abc2fc180413fb577d9958789973

I didn't change the functionality of this method when I extracted the task, but it clearly is not doing what we hoped.  I have a gradle build scan where configureProductDependencies is starting after resolveProductDependencies

esword: just writing notes to myself - this commit was to try to avoid building jars - https://github.com/palantir/sls-packaging/commit/b855a774d28d8d7ac618f2cce5215cbbc1dfdb41#diff-ffb538d33fb6230122fe54f923c3b269363d3fd26e7f1f38c58b9750aacf9e82R136-R143

It was about a month before the one that added the otherProjectProductDependenciesTasks  method above.  The problem with all of these is that they assume they have complete knowledge of what could write dependency info into the jar manifest (i.e. the ConfigureProductDependenciesTask).  Though I still don't know why I have seen the task ordering difference noted above.

fawind: Discussed offline and we'll move forward with this for now. If the overhead by compiling jars becomes too bad, we can look into other options.

fawind (approved):

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/CreateManifestIntegrationSpec.groovy
```diff
@@ -185,58 +185,6 @@ class CreateManifestIntegrationSpec extends GradleIntegrationSpec {
         file('bar-server/product-dependencies.lock').text.contains 'com.palantir.group:foo-service ($projectVersion, 1.x.x)'
     }
 
-    def "createManifest does not force compilation of sibling projects"() {
```
iamdanfox, line 188: @fawind @esword have we really lost this guarantee? IMO it's pretty critical that running `./gradlew --write-locks` remains fast, partly because it gets run a lot locally (e.g. when resolving merge conflicts or cherry picking things) and also gets run a ton on excavator.

robert3005: Agree with Dan. Is there another solution? Maybe using variants?

ferozco: I had suggested using variants and artifactTransformers offline. I think a solution similar to what we do for diagnostics could work here.

The only issue I see is that it is more challenging to get access to a manifest than to a file in the resource directory. I think we could work around this though by producing _both_ an entry in the JARs manifest (for backwards compatability) and a file in the resources directory that contained the Pdeps of a given JAR. We could then cleanly re-use the existing transforms (https://github.com/palantir/sls-packaging/blob/develop/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/DiagnosticsManifestPlugin.java#L165-L233) and leverage regular gradle cache validation

esword: I believe that writing the information to resource files is the way to go (it was the original design for the additional conjure information which is now being written), though that won't necessarily clear all of this up, especially if we are also writing to the meta-inf file (because we still have the existing problem).

However, there may be an optimization (a.k.a. hack) we can appy for in-project dependencies for lock files.  We do not need to actually compute the minimum version for such when writing lock files - we always write the template variable '$projectVersion'.  So if we think of updating locks separately, then we may be able to avoid compiling and writing jars.  I'll ponder a bit more.

fawind: I like Felipe's idea of also producing a pdeps artifact in addition to writing it in the JAR manifest as this would potentially simplify configuration.

> though that won't necessarily clear all of this up, especially if we are also writing to the meta-inf file (because we still have the existing problem).

Wouldn't project dependencies use the same version of sls-packaging though so we could rely on them using the new code path and producing an artifact containing the pdeps that does not require compiling the source?

> We do not need to actually compute the minimum version for such when writing lock files - we always write the template variable '$projectVersion'

Maybe I misunderstand but I think we need to know the concrete minimum version for writing the "product-dependencies.lock" file ([ref](https://github.com/palantir/sls-packaging/blob/d5b3149fd2867e9a09b50fb689bac0d1de61ac90/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/CreateManifestTask.java#L227))?

esword: in-project dependencies are always changed to the template ([ref](https://github.com/palantir/sls-packaging/blob/d5b3149fd2867e9a09b50fb689bac0d1de61ac90/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/ProductDependencyLockFile.java#L77)), unless there is some comparison I am missing?

pkoenig10: I think we need to revert this. After this PR, excavator checks will not open PRs if the resulting code does not compile.

Here's an example of a dependency upgrade excavator run on our internal server repo that failed because `compileJava` was run as part of `--write-locks`. This run did not open a PR.
https://e.p.b/api/coordinator/checkoutput/c2vrbesi5esaano38i40.stderr

PR #1155, opened 2021-06-08
Revert "move product dep calcs into own task (#1146)"

esword:
This reverts commit d5b3149f

## Before this PR
Can get race condition with writing jar manifest file

## After this PR

==COMMIT_MSG==
Revert "move product dep calcs into own task (#1146)"
==COMMIT_MSG==


ferozco:
👍

PR #1158, opened 2021-06-10
Step 2: Discover pdeps using artifact transformers

ferozco:
## After this PR

==COMMIT_MSG==
Discover product dependencies using artifact transformers
==COMMIT_MSG==


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ResolveProductDependenciesTask.java
```diff
@@ -0,0 +1,171 @@
...
+    abstract RegularFileProperty getManifestFile();
+
+    @TaskAction
+    public final void resolve() throws IOException {
+        Map<ProductId, ProductDependency> allProductDependencies = new HashMap<>();
+        Set<ProductId> allOptionalDependencies =
+                new HashSet<>(getOptionalProductIds().get());
+
+        getProductDependencies().get().forEach(declaredDep -> {
+            ProductId productId = new ProductId(declaredDep.getProductGroup(), declaredDep.getProductName());
```
esword: ProductId.of(declaredDep)

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ResolveProductDependenciesTask.java
```diff
@@ -0,0 +1,171 @@
...
+
+    private Multimap<ProductId, ProductDependency> discoverProductDependencies() {
+        Stream<ProductDependency> discoveredDependencies = getProductDependenciesFiles().getFiles().stream()
+                .map(ResolveProductDependenciesTask::safeDeserialize)
+                .flatMap(pdeps -> pdeps.recommendedProductDependencies().stream());
+        return discoveredDependencies.collect(
+                ImmutableSetMultimap.toImmutableSetMultimap(ProductId::of, Function.identity()));
+    }
+
+    private ProductDependency mergeDependencies(ProductId productId, ProductDependency dep1, ProductDependency dep2) {
```
esword, line 157: can replace method with ProductDependencyMerger::merge

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ProductDependencies.java
```diff
@@ -0,0 +1,74 @@
...
+import org.gradle.api.tasks.TaskProvider;
+
+public final class ProductDependencies {
+
+    public static TaskProvider<ResolveProductDependenciesTask> registerProductDependencyTasks(
+            Project project, BaseDistributionExtension ext) {
+        Provider<Directory> pdepsDir = project.getLayout().getBuildDirectory().dir("product-dependencies");
+        Configuration pdepsConfig = DependencyDiscovery.copyClasspath(project, "productDependencies");
+
+        // Register compatability rule to ensure that ResourceTransform is applied onto project dependencies so we
```
fawind:
```suggestion
        // Register compatibility rule to ensure that ResourceTransform is applied onto project dependencies so we
```

ferozco:
@fawind @CRogers mind taking a look?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ProductDependencies.java
```diff
@@ -0,0 +1,79 @@
...
+import org.gradle.api.Project;
+import org.gradle.api.artifacts.ArtifactView;
+import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.file.Directory;
+import org.gradle.api.provider.Provider;
+import org.gradle.api.tasks.TaskProvider;
+
+public final class ProductDependencies {
+
+    public static TaskProvider<ResolveProductDependenciesTask> registerProductDependencyTasks(
```
fawind, line 36: We still need to wire up this task and rip-out the pdeps logic from the current `createManifest` task, right?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ResolveProductDependenciesTask.java
```diff
@@ -0,0 +1,178 @@
...
+            });
+        });
+
+        allOptionalDependencies.stream()
+                .map(productId -> Optional.ofNullable(allProductDependencies.get(productId))
+                        .orElseThrow(() -> new IllegalStateException(String.format(
+                                "Unable to mark missing product dependency '%s' as optional", productId))))
+                .forEach(dep -> dep.setOptional(true));
+
+        CreateManifestTask.jsonMapper.writeValue(
```
fawind, line 141: nit: Would pull out this object mapper to not have cross-task references. Maybe into an `ObjectMappers` utils class?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/DiagnosticsManifestPlugin.java
```diff
@@ -160,75 +143,4 @@ private static Configuration createConsumableRuntimeConfiguration(Project projec
         project.getDependencies().add(consumable.getName(), project);
         return consumable;
     }
-
-    @CacheableTransform
-    public abstract static class ExtractFileFromJar implements TransformAction<ExtractFileFromJar.Parameters> {
```
fawind, line 165: Pretty nice how reusable transforms are!

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ProductDependencies.java
```diff
@@ -0,0 +1,79 @@
...
+            Project project, BaseDistributionExtension ext) {
+        Provider<Directory> pdepsDir = project.getLayout().getBuildDirectory().dir("product-dependencies");
+        Configuration pdepsConfig = DependencyDiscovery.copyClasspath(project, "productDependencies");
```
fawind: Why do we need to copy the classpath here instead of directly operating on it?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/PreferProjectCompatabilityRule.java
```diff
@@ -0,0 +1,62 @@
...
+        this.external = external;
+    }
+
+    @Override
+    public void execute(CompatibilityCheckDetails<Category> details) {
+        Category consumerValue = details.getConsumerValue();
+        Category producerValue = details.getProducerValue();
+        if (consumerValue == null) {
+            details.compatible();
+        } else if (project.equals(consumerValue) && external.equals(producerValue)) {
```
fawind: Can you elaborate a bit on how this prevents compilation, especially this statement?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ProductDependencies.java
ferozco: yup thats the next step!

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ResolveProductDependenciesTask.java
ferozco: for sure

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/PreferProjectCompatabilityRule.java
ferozco: Sure i'll leave a bit of a comment

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ProductDependencies.java
fawind: Makes sense!

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ResolveProductDependenciesTask.java
esword: if you pull from the set of changes I had that we rolled back, I created a Serializations class for this already.  Also had some enhancements to the test classes and such that would be nice to reuse.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ResolveProductDependenciesTask.java
esword: you don't need these extra checks.  InrepoProduct dependencies are always set to the projectVersion.  That's why I took out this method in the previous changes.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/DependencyDiscovery.java
```diff
@@ -0,0 +1,94 @@
...
+import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.artifacts.transform.TransformAction;
+import org.gradle.api.artifacts.transform.TransformParameters;
+import org.gradle.api.artifacts.type.ArtifactTypeDefinition;
+import org.gradle.api.attributes.Attribute;
+import org.gradle.api.attributes.Category;
+import org.gradle.util.GUtil;
+
+public final class DependencyDiscovery {
+    public static final Attribute<String> ARTIFACT_FORMAT = Attribute.of("artifactType", String.class);
```
CRogers, line 33: Just for my own understanding, is this `artifactType` string a built in bit of gradle or something custom we've made?

I see it [in the docs](https://docs.gradle.org/current/userguide/variant_model.html#sec:variant-visual) (that I have just spent a good long time grokking!) for something attached to artifacts, but not sure if that's the same thing as a variant attribute?

esword (changes requested):

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/DependencyDiscovery.java
```diff
@@ -0,0 +1,94 @@
...
+
+    public static Configuration copyClasspath(Project project, String name) {
+        Configuration conf = project.getConfigurations()
+                .create(GUtil.toLowerCamelCase("runtimeClasspathFor " + name), c -> {
+                    c.setCanBeConsumed(true);
+                    c.setCanBeResolved(true);
+                    c.setVisible(false);
+                });
+        Map<String, String> projectDependency =
+                ImmutableMap.of("path", project.getPath(), "configuration", "runtimeElements");
```
esword: nit: do you want to hardcode the "runtimeElements" string vs the nice-but-verbose:
project.getConvention().getPlugin(JavaPluginConvention.class).getSourceSets().getByName(SourceSet.MAIN_SOURCE_SET_NAME).getRuntimeElementsConfigurationName()

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ProductDependencies.java
ferozco: This is a nice optimisation to making discovery of "X" (where x is monitors, metrics, pdeps, etc.) from the current project flow through the same code path as other dependent projects. For pdeps, it isn't strictly necessary since we consume the extension explicitly but it is for diagnostics

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/DependencyDiscovery.java
CRogers: I see, it is a built in thing (just no public api to get the magic string/Attribute object) and just filters the things you want to resolve via the configuration.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/DependencyDiscovery.java
CRogers: If you want to avoid hardcoding it you can do `JavaPlugin.RUNTIME_ELEMENTS_CONFIGURATION_NAME`

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/DependencyDiscovery.java
```diff
@@ -0,0 +1,94 @@
...
+            v.getAttributes().attribute(ARTIFACT_FORMAT, targetArtifact);
+            Category projectCategory =
+                    project.getObjects().named(Category.class, PreferProjectCompatabilityRule.PROJECT);
+            v.getAttributes().attribute(Category.CATEGORY_ATTRIBUTE, projectCategory);
+        });
+    }
+
+    public static Configuration copyClasspath(Project project, String name) {
+        Configuration conf = project.getConfigurations()
+                .create(GUtil.toLowerCamelCase("runtimeClasspathFor " + name), c -> {
```
CRogers: Is it worth making this configuration name a bit more specific to sls-packaging/productdeps? I could just imagine someone else making one with the same name in a different plugin.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/DependencyDiscovery.java
ferozco: it is something that gradle defines, the variable is exposed as part of their internal API

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/ExtractSingleFileOrManifest.java
```diff
@@ -0,0 +1,72 @@
...
+                    Files.copy(is, outputFile.toPath());
+                }
+                return;
+            }
+
+            ZipEntry manifestEntry = zipFile.getEntry(MANIFEST);
+            if (manifestEntry != null) {
+                Manifest manifest = new Manifest(zipFile.getInputStream(manifestEntry));
+                File outputFile = outputs.file("manifest.json");
+                GFileUtils.writeFile(manifest.getMainAttributes().getValue(manifestKey), outputFile);
```
CRogers: `GFileUtils.writeFile` is a footgun as it uses the default charset.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ProductDependencies.java
```diff
@@ -0,0 +1,79 @@
...
+        // avoid compilation
+        PreferProjectCompatabilityRule.configureRule(project);
+
+        DependencyDiscovery.configureJarTransform(
+                project, ExtractSingleFileOrManifest.class, DependencyDiscovery.PRODUCT_DEPENDENCIES, params -> {
+                    params.getPathToExtract().set(RecommendedProductDependenciesPlugin.RESOURCE_PATH);
+                    params.getKeyToExtract().set(RecommendedProductDependencies.SLS_RECOMMENDED_PRODUCT_DEPS_KEY);
+                });
+
+        DependencyDiscovery.configureResourceTransform(
```
CRogers, line 52: I'm probably missing something, but if we can extract the pdeps from the manifest in the above transform, why do we need to put it in/extract it from the resources as well?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/ExtractSingleFileOrManifest.java
```diff
@@ -0,0 +1,72 @@
...
+    @InputArtifact
+    public abstract Provider<FileSystemLocation> getInputArtifact();
+
+    @Override
+    public final void transform(TransformOutputs outputs) {
+        File jarFile = getInputArtifact().get().getAsFile();
+        String pathToExtract = getParameters().getPathToExtract().get();
+        String manifestKey = getParameters().getKeyToExtract().get();
+
+        try (ZipFile zipFile = new ZipFile(jarFile)) {
```
CRogers: We're probably going to need to update dev-env to support get pdeps from resources as well as manifest too: https://github.palantir.build/foundry/dev-env/blob/e690c9a01b68297b59269febca4f4a9df0ff8dcc/dev-env-common/src/main/java/com/palantir/devenv/sls/PdepJarExtractor.java.

I wonder if we should make that library to reduce duplication.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/ExtractSingleFileOrManifest.java
```diff
@@ -0,0 +1,72 @@
...
+                            + pathToExtract.replaceAll("/", "-");
+                    File outputFile = outputs.file(newFileName);
+                    Files.copy(is, outputFile.toPath());
+                }
+                return;
+            }
+
+            ZipEntry manifestEntry = zipFile.getEntry(MANIFEST);
+            if (manifestEntry != null) {
+                Manifest manifest = new Manifest(zipFile.getInputStream(manifestEntry));
```
CRogers: I think you can avoid doing this by using `JarFile` that extends from `ZipFile`.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/ExtractSingleFileOrManifest.java
ferozco: good catch

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ProductDependencies.java
ferozco: we do this to avoid compilation since to get the manifest we need to create a JAR and a JAR requires compilation

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/ExtractSingleFileOrManifest.java
ferozco: the existing `gradle-recommended-product-dependencies` should expose enough to make it easy for people to consistently access pdeps however they are embedded

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ProductDependencies.java
CRogers: Ah, of course!

esword (approved):

PR #1160, opened 2021-06-14
Disable biased locking

carterkozak:
==COMMIT_MSG==
Disable biased locking by providing `-XX:-UseBiasedLocking` to jvm args for java versions prior to jdk15.
==COMMIT_MSG==

Biased locking is disabled on java 15+ https://openjdk.java.net/jeps/374

We've found a lot of safepoints due to `RevokeBias`, and several projects have observed a performance improvement by disabling biased locking. We're not aware of any cases in which disabling biased locks reduces performance because our applications tend to be multi-threaded, and modern CPUs handle the CAS elegantly.
biased locking can result in non-obvious safepoints when `System.identityHashCode` is called because the identity hashcode and lock bias are stored in the same portion of the object header.

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
@@ -47,7 +47,7 @@ import org.immutables.value.Value;
 
 public abstract class LaunchConfigTask extends DefaultTask {
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper(new YAMLFactory());
-    private static final List<String> java8gcLoggingOptions = ImmutableList.of(
+    private static final ImmutableList<String> java8gcLoggingOptions = ImmutableList.of(
             "-XX:+PrintGCDateStamps",
             "-XX:+PrintGCDetails",
             "-XX:-TraceClassUnloading",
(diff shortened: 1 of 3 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -183,6 +185,14 @@ public final void createConfig() throws IOException {
                                 javaVersion.get().compareTo(JavaVersion.toVersion("15")) == 0
                                         ? java15Options
                                         : ImmutableList.of())
+                        // Biased locking is disabled on java 15+ https://openjdk.java.net/jeps/374
+                        // We disable biased locking on all releases in order to reduce safepoint time,
+                        // revoking biased locks requires a safepoint, and can occur for non-obvious
+                        // reasons, e.g. System.identityHashCode.
```
schlosna, line 191: The JEP 374 didn't explicitly list these, but relevant references of interest for folks that run across this:
https://shipilev.net/jvm/objects-inside-out/#_observation_identity_hashcode_disables_biased_locking & https://wiki.openjdk.java.net/display/lilliput/Main#Main-Thelock/markword

Original motivation for biased locking circa 2006 https://blogs.oracle.com/dave/biased-locking-in-hotspot & ["Quickly Reacquirable Locks"](https://cdn.app.compendium.com/uploads/user/e7c690e8-6ff9-102a-ac6d-e4aebca50425/f4a5b21d-66fa-4885-92bf-c4e81c06d916/File/ccd39237cd4dc109d91786762fba41f0/qrl_oplocks_biasedlocking.pdf) paper.

schlosna (approved):

PR #1159, opened 2021-06-11
Step3: Plumb  in new discovery mechanism 

ferozco:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

PR #1162, opened 2021-06-15
Step 2.5: misc tweaks for cleanup and add tests for resolve task

esword:
Also fix one bug in resolve - it was allowing self dependencies

The tests for the resolve task are taken straight from the createManifest spec.
Once we wire the two tasks together, will delete the duplicate ones from the manifest spec.

## After this PR

==COMMIT_MSG==
misc tweaks for cleanup and add tests for resolve task
==COMMIT_MSG==


esword:
changed merged into the #1159

PR #1164, opened 2021-06-16
Feature: Discover product dependencies in a lazy, pluggable way

ferozco:
## After this PR

==COMMIT_MSG==
Discover product dependencies in a lazy, pluggable way
==COMMIT_MSG==

### DIFF
--- a/gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/CompileRecommendedProductDependencies.java
+++ b/gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/CompileRecommendedProductDependencies.java
@@ -1,0 +1,47 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.dist;
+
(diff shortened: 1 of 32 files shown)


ferozco:
Pending #1159

gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/ConfigureProductDependenciesTask.java
```diff
@@ -1,5 +1,5 @@
 /*
- * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
```
carterkozak:
```suggestion
 * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
```

gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/CompileRecommendedProductDependencies.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Set;
+import org.gradle.api.DefaultTask;
+import org.gradle.api.file.RegularFileProperty;
+import org.gradle.api.provider.SetProperty;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.OutputFile;
+import org.gradle.api.tasks.TaskAction;
+
+public abstract class CompileRecommendedProductDependencies extends DefaultTask {
+    static final ObjectMapper MAPPER = new ObjectMapper();
```
carterkozak, line 30: Do we need to worry about this ObjectMapper being configured differently from `ObjectMappers.jsonMapper`?

ferozco: good catch, we should use the shared `ObjectMappers` mapper

ferozco: err actually they're in separate packages. I don't think it ends up mattering but if it did we have tests that exercise that the JSON produced by this task are consumable in the main plugin

carterkozak (approved):
lgtm

PR #1165, opened 2021-06-16
the conjure-gradle plugin needs to be able to set the inputs.

esword:
making the output also public for completeness.

## After this PR

==COMMIT_MSG==
the conjure-gradle plugin needs to be able to set the inputs
==COMMIT_MSG==


fawind (approved):

fawind:
We're using the extension instead: https://github.com/palantir/gradle-conjure/pull/794

PR #1166, opened 2021-06-17
Handle missing manifest when transforming artifacts

fawind:
## Before this PR
FLUP on https://github.com/palantir/sls-packaging/pull/1164

For some external jars, the manifest file is not present which results in a NPE when trying to retrieve an attribute from it.

Example stacktrace:

```
Caused by: org.gradle.api.internal.artifacts.transform.TransformException: Failed to transform trove3-3.0.3-p5.jar
    [...]
Caused by: org.gradle.api.internal.artifacts.transform.TransformException: Execution failed for ExtractSingleFileOrManifest: /home/circleci/.gradle/caches/.../trove3-3.0.3-p5.jar.
    [...]
Caused by: java.lang.NullPointerException: Cannot invoke "java.util.jar.Manifest.getMainAttributes()" because "manifest" is null
    [...]
```

## After this PR
==COMMIT_MSG==
Handle missing manifest when transforming artifacts
==COMMIT_MSG==


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/ExtractSingleFileOrManifest.java
```diff
@@ -57,11 +57,11 @@ public final void transform(TransformOutputs outputs) {
                 return;
             }
 
-            Manifest manifest = jar.getManifest();
-            String value = manifest.getMainAttributes().getValue(key);
-            if (value != null) {
+            Optional<String> value = Optional.ofNullable(jar.getManifest())
+                    .map(manifest -> manifest.getMainAttributes().getValue(key));
```
CRogers, line 61: It always freaks me out slightly how returning `null` inside a map gets changed into an optional

CRogers (approved):

CRogers:
It might be nice to have a test for this if it's reasonable to do (on the actual `trove` dep)

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/ExtractSingleFileOrManifest.java
fawind: Yeah, also had to double check that it actually does so!

CRogers: It actually means that `map` is really `flatMap` as you can change whether the optional is present or empty. This makes haskell programmer inside me die a little.

fawind: But imagine the bugs you run into otherwise, when your `Optional<T>` is suddenly `null` :P

fawind:
Added a test case

CRogers:
:+1:

PR #1167, opened 2021-06-17
fix check for duplicated product dependencies

esword:
## Before this PR
would log error message if more than one version of the same recommended dep

## After this PR

==COMMIT_MSG==
logs message only if a declared and a recommended overlap
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ResolveProductDependenciesTask.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ResolveProductDependenciesTask.java
@@ -121,6 +121,16 @@ public abstract class ResolveProductDependenciesTask extends DefaultTask {
                 log.trace("Product dependency for '{}' declared as optional", productId);
                 allOptionalDependencies.add(productId);
             }
+            if (discoveredDependencies.containsKey(productId)) {
+                log.error(
+                        "Please remove your declared product dependency on '{}' because it is"
+                                + " already provided by a jar dependency:\n\n"
+                                + "\tProvided:     {}\n"
+                                + "\tYou declared: {}",
+                        productId,
+                        discoveredDependencies.get(productId),
+                        declaredDep);
+            }
         });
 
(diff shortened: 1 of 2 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ResolveProductDependenciesTask.java
```diff
@@ -135,16 +145,6 @@ public final void resolve() throws IOException {
             allProductDependencies.merge(productId, discoveredDependency, (declaredDependency, _newDependency) -> {
                 ProductDependency mergedDependency =
                         mergeDependencies(productId, declaredDependency, discoveredDependency);
-                if (mergedDependency.equals(discoveredDependency)) {
```
ferozco, line 138: I don't think this is quiet right, you still want this error to be produced. The problem is that you want to merge the discovered dependencies before attempting to merge them with the declared dependencies

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ResolveProductDependenciesTask.java
```diff
@@ -123,6 +123,9 @@ public final void resolve() throws IOException {
             }
         });
 
+        // Perform initial merge of the discovered dependencies so that if there are overlaps with declared
+        // dependencies we only log a single time.
+        Map<ProductId, ProductDependency> mergedDiscoveredDependencies = new HashMap<>();
```
ferozco: instead of constructing a new map can we merge as we iterate? something like: https://github.com/palantir/sls-packaging/compare/fo/example?expand=1

esword: changed as requested, except I took out the "if (mergedDependency.equals(discoveredDependency)))" because there are times when it is not an exact match - e.g. the recommended dependency did not contain a recommended version and that got merged in.

ferozco (approved):

PR #1174, opened 2021-07-02
Enable -XX:NativeMemoryTracking=summary

iamdanfox:
## Before this PR

In WC 1.71.0, I added a bunch of metrics (~60) extracted from running `jcmd VM.native_memory`.  This only works if you have  `-XX:NativeMemoryTracking=summary` enabled.

I was initially quite cautious about the extra 'overhead' that so many blogs talked about, but after Carter's benchmarks and seeing timelock run with it on an early-adopter stack, we haven't been able to see any noticeable performance impact. The space overhead was also very small (~8MiB for timelock's 4g heap).

## After this PR
==COMMIT_MSG==
Enable `-XX:NativeMemoryTracking=summary` for all
==COMMIT_MSG==

`do not merge` until we've come to a decision about the DD $ cost of exposing lots more metrics.

### DIFF
--- a/changelog/@unreleased/pr-1174.v2.yml
+++ b/changelog/@unreleased/pr-1174.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Enable `-XX:NativeMemoryTracking=summary` for all
+  links:
+  - https://github.com/palantir/sls-packaging/pull/1174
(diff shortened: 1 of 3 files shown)


jkozlowski (approved):
I think this will be useful for validating the 75% guesstimate and overall getting a better understanding of our native memory usage. I am curious if it will be at all helpful to debug OOMkills, since I suspect those are quite sudden and the metric might not get sampled?

iamdanfox:
Ok given the discussion on https://internal-github/foundry/witchcraft/pull/3392, I think I'm happy to take the `do not merge` label off.

Feel free to release at your leisure

PR #1176, opened 2021-07-06
Add jvm args to jdk16+ to support safepoint metrics

carterkozak:
They're some of our most actionable metrics which would be painful
to lose.

Not seeing progress here:
https://mail.openjdk.java.net/pipermail/serviceability-dev/2021-June/038164.html

## After this PR
==COMMIT_MSG==
Add jvm args to jdk16+ to support safepoint metrics
==COMMIT_MSG==

## Possible downsides?
Ideally we wouldn't rely on `--add-exports`

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
@@ -56,10 +56,15 @@ public abstract class LaunchConfigTask extends DefaultTask {
             "-XX:NumberOfGCLogFiles=10",
             "-Xloggc:var/log/gc-%t-%p.log",
             "-verbose:gc");
-    private static final ImmutableList<String> java14Options =
+    private static final ImmutableList<String> java14PlusOptions =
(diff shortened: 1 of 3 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -56,10 +56,15 @@
...
             "-Xloggc:var/log/gc-%t-%p.log",
             "-verbose:gc");
-    private static final ImmutableList<String> java14Options =
+    private static final ImmutableList<String> java14PlusOptions =
             ImmutableList.of("-XX:+ShowCodeDetailsInExceptionMessages");
     private static final ImmutableList<String> java15Options =
             ImmutableList.of("-XX:+UnlockDiagnosticVMOptions", "-XX:+ExpandSubTypeCheckAtParseTime");
+    // Support safepoint metrics from the internal sun.management package in production. We prefer not
+    // to use '--illegal-access=permit' so that we can avoid unintentional and unbounded illegal access
+    // that we aren't aware of.
```
fawind, line 65: This sounds good to me!

fawind (approved):

PR #1178, opened 2021-07-09
Add a marker task for writing just product dependencies locks

AlexLandau:
## Before this PR
In some circumstances, it is useful to do just one type of lock
writing at a time, without running everything that keys off of
`--write-locks`. This is less common for `product-dependencies.lock`
updates, which are now usually slower than other types, but this
still comes in useful in certain situations.

Our particular situation involves a GCV variant that ends up with
different classpath orderings when writing vs. not writing locks, so
JavaCompile tasks don't stay up-to-date. Thus, we want to keep
the locking of Java dependencies in different Gradle invocations
from Java compilation, which we can't do if writing product
dependency locks requires both `--write-locks` and compilation.

## After this PR
==COMMIT_MSG==
Add a new `writeProductDependenciesLocks` task that allows writing out the `product-dependencies.lock` files without running `--write-locks` and possibly triggering undesirable code paths from other gradle plugins.
==COMMIT_MSG==

This matches the [convention established in gradle-consistent-versions](https://github.com/palantir/gradle-consistent-versions/pull/723#discussion_r652680116), used by the `writeVersionsLocks` task.


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/TaskNameMatcher.java
```diff
@@ -0,0 +1,64 @@
...
+
+package com.palantir.gradle.dist.tasks;
+
+import com.google.common.collect.Streams;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+
+// Copied from gradle-consistent-versions
```
AlexLandau: Note: I copied this from gradle-consistent-versions but did not copy over the test, hoping this is a temporary situation. It didn't seem like a good idea to add a full `implementation` dependency for the sake of this class, and I don't know that we currently have any Gradle utilities library for this to go into.

As an alternative: Since this plugin writes locks in a task action rather than during configuration time, maybe we can check the task graph directly instead of checking the start parameters. I haven't tried that yet.

readme.md
```diff
@@ -85,7 +85,7 @@ distribution {
 }
 ```
 
-sls-packaging also maintains a lockfile, `product-dependencies.lock`, which should be checked in to Git.  This file is an accurate reflection of all the inferred and explicitly defined product dependencies. Run **`./gradlew --write-locks`** to update it. e.g.
+sls-packaging also maintains a lockfile, `product-dependencies.lock`, which should be checked in to Git.  This file is an accurate reflection of all the inferred and explicitly defined product dependencies. Run **`./gradlew --write-locks`** or **`./gradlew writeProductDependenciesLocks`** to update it. e.g.
```
AlexLandau, line 80: There are also assorted error messages prompting the user to run `--write-locks` (and the comment at the top of the `product-dependencies.lock` files). I didn't change those, since most users likely won't benefit much from this marker task.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/TaskNameMatcher.java
AlexLandau: Looks like the task graph approach works.

ferozco (commented):
Instead of a creating an indicator task could we add an [option](https://docs.gradle.org/current/userguide/custom_tasks.html#sec:using_task_option_command_line) to the existing createManifest task to cause the lock file to be re-written. I'm imagining something like:
`./gradlew createManifest --updateLock`.

Alternatively we could entirely pull out the lock file work from createManifest task and have a separate task for writing the lock file with its own option. i.e:
`./gradlw checkPdepLock --update`

in either case, we would set the default value of the option to be whether `--write-locks` is set. Effectively making `./gradlew --write-locks` equivalent to `./gradlew createManifest --updateLock`

tpetracca:
Hey @AlexLandau, ended up discussing this a bit between the infra buddies today. I think we've landed on what we'd like the full structure of how this should work to be (see below, and please let me know if its unclear), but @CRogers (who ultimately owns the tech here) is out this week and either way I'd like to wait for him to be able to review and approve said proposal before we actually move forward. That sound good?

Proposal:
- entirely move product dependency lock writing out to its own task, let's call it `writeProductDependenciesLock` though I don't feel strongly on name
- executing `writeProductDependenciesLock` only writes said lock, it does not invoke the `createManifest` task
- `createManifest` should check locks file and fail if not up to date
- if `createManifest` and `writeProductDependenciesLock` are both in the task graph, `writeProductDependenciesLock` should explicitly happen before `createManifest` (such that it is then impossible for the locks check to fail)

as such:
```
# this just updates locks; doesn't write manifest
./gradlew writeProductDependenciesLock

# this checks pdeps locks but does not write them
./gradlew createManifest

# this updates locks and creates manifest; should never fail checking the locks
./gradlew createManifest writeProductDependenciesLock

# in some future world we'll have something like this
# this updates all locks (depends on writeProductDependenciesLock, writeVersionsLock from GCV, etc.)
./gradlew writeAllLocks
```

tpetracca (changes requested):
see my big comment

AlexLandau:
I'm happy to wait until there's a consensus.

My main concern with the idea of splitting the work between two tasks was that I saw some existing users with code like `tasks.createManifest.dependsOn <...>` that would need to be updated in conjunction with this change to avoid race conditions or other problems.

CRogers:
On a purely making-the-implementation-easy side of things, I like @ferozco's adding an argument to the task idea. But it's probably going to be nicer for the user to have this be the same as GCV with a `writeProductDependencyLocks` task, even if the implementation is a bit more involved.

Regarding `writeAllLocks` - I'm not sure this something we're going to be able to move to. We need to keep supporting `--write-locks` for backcompat in most of our plugins so we might as well just keep using that instead. GCV (as always) makes this particularly hard to do `writeAllLocks` nicely, as write locks action is actually performed at configuration time before the task graph is calculated.

davidscottcohen:
is there any movement here?

esword:
@CRogers - did you have any time to apply some brain power on this like you mentioned a couple of weeks ago?

CRogers (approved):
Apologies, this review slipped amongst other fires - I think the current implementation looks good with the marker task, to make it line up with the `writeVersionsLocks` task in GCV.

PR #1183, opened 2021-07-20
Copy pdeps config lazily

fawind:
## Before this PR
We have an internal project that tries to explicitly overwrite the minimum version of a product dependency by doing the following:

```gradle
apply plugin: 'com.palantir.sls-asset-distribution'

configurations {
    overriddenProductDependencies {
        extendsFrom assetBundle
        // Brings in a pdep with a higher version of 'com.palantir.my-service'
        exclude group: project.group, module: 'my-api-conjure-objects'
        exclude group: project.group, module: 'my-api-conjure-jersey'
    }
}

distribution {
    // We don't want to pick up the higher pdep from my-api-conjure-objects or -jersey
    // but instead hardcode the min version to 1.0.0.
    productDependenciesConfig = configurations.overriddenProductDependencies
    productDependency {
        productGroup = 'com.palantir'
        productName = 'my-service'
        minimumVersion = '1.0.0'
    }
}

```

However the problem is that sls-packaging copies the `productDependencyConfig` ([ref](https://github.com/palantir/sls-packaging/blob/4870f4363c886e304401a78c3662db8f3d4732dc/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ProductDependencies.java#L41-L42)) *before* the project build file is configured and the `overriddenProductDependencies` configuration got created.
Based on that, the default `assetBundle` configuration is used for resolving product dependencies ([ref](https://github.com/palantir/sls-packaging/blob/4ec8b15b65c1e02f3c4faeb86079a5f6497bfbce/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/asset/AssetDistributionPlugin.java#L58)) and the `overriddenProductDependencies` with its exclude rules are ignored.
As a result, the minimum version coming from `my-api-conjure-jersey` is picked up and written into the manifest.

## After this PR

FLUP on https://github.com/palantir/sls-packaging/pull/1164.

Copy and wire up the `productDependencyConfig` lazily, allowing consumers to configure the extension at any point during configuration.

Note that this increases the minimum Gradle version to 6.1.

==COMMIT_MSG==
Copy pdeps config lazily
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/DependencyDiscovery.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/DependencyDiscovery.java
@@ -16,6 +16,7 @@
 
 package com.palantir.gradle.dist.artifacts;
 
+import java.util.Optional;
 import java.util.function.Consumer;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.ArtifactView;
@@ -40,17 +41,20 @@ public final class DependencyDiscovery {
     }
 
     public static Configuration copyConfiguration(Project project, String configurationName, String name) {
-        Configuration consumable = project.getConfigurations()
-                .create(GUtil.toLowerCamelCase(configurationName + " for " + name), conf -> {
-                    conf.extendsFrom(project.getConfigurations().getByName(configurationName));
-                    conf.setDescription("DiagnosticsManifestPlugin uses this configuration to extract single file");
-                    conf.setCanBeConsumed(true);
-                    conf.setCanBeResolved(true);
-                    conf.setVisible(false);
-                });
+        String newConfigName = GUtil.toLowerCamelCase(GUtil.toLowerCamelCase(configurationName + " for " + name));
+        return Optional.ofNullable(project.getConfigurations().findByName(newConfigName))
+                .orElseGet(() -> {
+                    Configuration consumable = project.getConfigurations().create(newConfigName, conf -> {
+                        conf.extendsFrom(project.getConfigurations().getByName(configurationName));
+                        conf.setDescription("DiagnosticsManifestPlugin uses this configuration to extract single file");
+                        conf.setCanBeConsumed(true);
+                        conf.setCanBeResolved(true);
+                        conf.setVisible(false);
+                    });
 
-        project.getDependencies().add(consumable.getName(), project);
-        return consumable;
+                    project.getDependencies().add(consumable.getName(), project);
+                    return consumable;
+                });
     }
 
     public static <T extends TransformAction<P>, P extends TransformParameters> void configureJarTransform(
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/DependencyDiscovery.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/DependencyDiscovery.java
@@ -16,7 +16,6 @@
 
 package com.palantir.gradle.dist.artifacts;
 
-import java.util.Optional;
 import java.util.function.Consumer;
 import org.gradle.api.Project;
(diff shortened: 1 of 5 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/DependencyDiscovery.java
```diff
@@ -40,17 +41,20 @@ public static ArtifactView getFilteredArtifact(Project project, Configuration co
     }
 
     public static Configuration copyConfiguration(Project project, String configurationName, String name) {
```
fawind: I haven't fully figured out why but the `pdepsConfigProvider` get resolved twice for some projects and thus we need to make sure that we don't crash on subsequent calls.

It would be nicer to somehow cache the provider but don't think there is a good way in Gradle.

ferozco: did you see this happening in tests? You could temporarily capture and log the stacktrace to figure out what is calling the provider. It definitely seems weird its being invoked multiple times

fawind: Can verify it with e.g. [this unit test](https://github.com/palantir/sls-packaging/blob/545a84381ce843cf6a8b34c6c3309725fb906567/gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/pdeps/ResolveProductDependenciesIntegrationSpec.groovy#L50).

These are the stacktraces of the two `DependencyDiscovery#copyConfiguration` invocations. Initial version is the first invocation and the first revision the second invocation: https://gist.github.com/fawind/43830abfe89a33a7ffab85dd22ec4425/revisions?diff=split

Some highlights from the fist invocation stacktrace:
```
org.gradle.api.internal.tasks.CachingTaskDependencyResolveContext
org.gradle.execution.plan.LocalTaskNode.resolveDependencies
org.gradle.initialization.BuildOperationFiringTaskExecutionPreparer$CalculateTaskGraph
```

and from the second invocation stacktrace:
```
org.gradle.api.internal.file.collections.DefaultConfigurableFileCollection.lambda$calculateFinalizedValue
org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$TaskExecution.visitInputs
org.gradle.internal.execution.steps.CaptureStateBeforeExecutionStep.captureExecutionState
```

So it looks like the first invocation happens when Gradle builds its task graph and then the second one when the actual task is executed?

Relavant issue: https://github.com/gradle/gradle/issues/6787

fawind: I think we can instead of the provider use a property with `Property#finalizeValueOnRead` as this will only get resolved once. Pushed the code for this and reverted the changes to `DependencyDiscovery#copyConfiguration`.

Let me add a test overwriting the pdeps configuration though.

fawind: Hmm, finalizeValueOnRead only exists for Gradle 6.1+ :(

carterkozak: we don't need to support older versions anymore, do we?

ferozco: Its probably fine to rely on gradle 6.1 features since it was released over a year and a half ago

fawind: Agree. Increased the min version to 6.1 and changed the changelog to break to be extra correct.

ferozco (approved):

PR #1184, opened 2021-07-21
Exclude AtlasDB schema from main class resolution

pkoenig10:
Attempting to use V2 of our internal AtlasDB Gradle plugin in some internal repos.

Unfortunately, because the schema files are part of the main source set, `MainClassResolver` finds both the schema class and the server class.

This PR updates `MainClassResolver` to ignore any files that import the AtlasDB `Schema` class. Server classes should never need to reference the `Schema` class directly - I've verified this is true in all internal repos.

I've verified this works as expected using `publishToMavenLocal` with an internal repo.


fawind (approved):
Sounds reasonable, thanks for fixing!

PR #1191, opened 2021-08-11
Include launcher-static.yml in config tars

CRogers:
## Before this PR
`launcher-static.yml` is not included in config tgzs, meaning in dev-env we must use artifcatory paths like `/internal-dist/group/name/version/big-dist.tgz!service/bin/launcher-static.yml`. This means we put a lot of cpu load on artifactory and it can take a long time to download.

## After this PR
==COMMIT_MSG==
Include `service/bin/launcher-static.yml` in java config tars.
==COMMIT_MSG==

This will allow dev-env to download a much smaller config tgz instead, which should be pretty much instant.

### DIFF
--- a/gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/ConfigTarTaskIntegrationSpec.groovy
+++ b/gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/ConfigTarTaskIntegrationSpec.groovy
@@ -16,16 +16,17 @@
 
 package com.palantir.gradle.dist.tasks
 
-import com.palantir.gradle.dist.GradleIntegrationSpec
 
-class ConfigTarTaskIntegrationSpec extends GradleIntegrationSpec {
+import nebula.test.IntegrationSpec
+
+class ConfigTarTaskIntegrationSpec extends IntegrationSpec {
 
     def 'configTar task exists for services'() {
         setup:
         createUntarBuildFile(buildFile, "java-service", "service", "foo-service")
 
         when:
-        runTasks(':configTar')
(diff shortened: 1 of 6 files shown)


gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/tasks/ConfigTarTaskIntegrationSpec.groovy
```diff
@@ -47,22 +48,23 @@ class ConfigTarTaskIntegrationSpec extends GradleIntegrationSpec {
...
+        runTasksSuccessfully(':configTar', ':untar')
 
         then:
         def files = directory('dist/foo-service-0.0.1/', projectDir).list()
-        files.length == 1
+        files.length == 2
         files.contains('deployment')
         def manifest = file('dist/foo-service-0.0.1/deployment/manifest.yml', projectDir).text
         manifest.contains('service.v1')
+        file('dist/foo-service-0.0.1/service/bin/launcher-static.yml')
```
carterkozak: I believe `file` will always create a new file if it doesn't exist. I'm not very familiar with groovy, but I'd assume that would evaluate to true even if a new empty file was created.
```suggestion
        fileExists('dist/foo-service-0.0.1/service/bin/launcher-static.yml')
```

CRogers: Good spot - the number of times I fall into this trap is actually incredible. Also I didn't even check for existence 🤦🏻 .

carterkozak (approved):

PR #1205, opened 2021-09-07
Allow product dependency to also be declared as optional

pkoenig10:
## Before this PR
To declare an optional dependency that is not discovered, you have to do:
```
productDependency {
    productGroup = 'com.palantir.foo'
    productName = 'foo-service'
    minimumVersion = '1.2.3'
    optional = true
}
```

It would be much nicer if we could use the shorthand that looks similar to dependencies for these optional dependencies.

## After this PR
To declare an optional dependency that is not discovered, you can do:
```java
productDependency 'com.palantir.foo:foo-service:1.2.3'

optionalProductDependency 'com.palantir.foo:foo-service'
```

Alternatively we could allow something like:
```
optionalProductDependency 'com.palantir.foo:foo-service:1.2.3'
```

But this is trickier because it conflicts with the existing method.

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ResolveProductDependenciesTask.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ResolveProductDependenciesTask.java
@@ -109,12 +109,6 @@ public abstract class ResolveProductDependenciesTask extends DefaultTask {
                                 + "dependency or ignore",
                         productId));
             }
-            if (getOptionalProductIds().get().contains(productId)) {
-                throw new IllegalArgumentException(String.format(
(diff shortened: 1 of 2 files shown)


fawind:
Sorry, not super sure I follow the intend of this PR. Can you add a bit more context, you want to add a shorthand syntax for declaring optional pdeps?

pkoenig10:
Yes that's correct.

pkoenig10:
We already have shorthand for:
- Declaring a non-optional product dependency
- Making a discovered product dependency optional

But we don't have shorthand for:
- Declaring an optional product dependency

It's nice for product dependency configuration to be consistent.

fawind:
Thanks, I see it now. We would combine two statements to mark something as optional as in your example. No strong opinion here but sounds reasonable to me!

fawind (approved):

PR #1215, opened 2021-10-29
Apply `--add-exports` to Java 16+ services based on manifest entries

carterkozak:
This provides the same functionality as
https://github.com/palantir/gradle-baseline/pull/1944/

We're piggy-backing off the `Add-Exports` manifest property to collect exports required for any given application based on the libraries it uses.

==COMMIT_MSG==
Apply `--add-exports` to Java 16+ services based on manifest entries
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/ModuleExports.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/ModuleExports.java
@@ -27,7 +27,11 @@ import org.gradle.api.Project;
 import org.gradle.api.file.FileCollection;
 
 /**
- * Utility functionality which mirrors {@code Add-Exports} plumbing from
+ * This utility class reuses the {@code Add-Exports} manifest entry defined in
+ * <a href="https://openjdk.java.net/jeps/261">JEP-261</a> to collect required exports
+ * from the runtime classpath so they may be applied to the static configuration.
+ *
+ * Note that this mirrors {@code Add-Exports} plumbing from
  * <a href="https://github.com/palantir/gradle-baseline/pull/1944">gradle-baseline#1944</a>.
  */
 final class ModuleExports {
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/ModuleExports.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/ModuleExports.java
@@ -44,7 +44,7 @@ final class ModuleExports {
 
     static ImmutableList<String> getExports(Project project, JavaVersion javaVersion, FileCollection classpath) {
         // --add-exports is unnecessary prior to java 16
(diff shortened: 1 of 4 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/ModuleExports.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.stream.Stream;
+import org.gradle.api.JavaVersion;
+import org.gradle.api.Project;
+import org.gradle.api.file.FileCollection;
+
+/**
+ * Utility functionality which mirrors {@code Add-Exports} plumbing from
+ * <a href="https://github.com/palantir/gradle-baseline/pull/1944">gradle-baseline#1944</a>.
+ */
+final class ModuleExports {
```
fawind, line 37: I think it would be nice to to also have a small explanation here instead of just linking to the PR. Something similar to what you added in:

https://github.com/palantir/gradle-baseline/blob/1a052650da4af39680a7dd3f790a5b291086158e/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java#L42-L47

fawind (approved):
Looks good, I think the code duplication with the gradle-baseline feature is actually not too bad!

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/ModuleExports.java
```diff
@@ -0,0 +1,84 @@
...
+        // --add-exports is unnecessary prior to java 16
+        if (javaVersion.compareTo(JavaVersion.toVersion("16")) < 0) {
+            return ImmutableList.of();
+        }
+
+        return Stream.concat(
+                        classpath.getFiles().stream().flatMap(file -> {
+                            try {
+                                if (file.getName().endsWith(".jar") && file.isFile()) {
+                                    JarFile jar = new JarFile(file);
```
iamdanfox: driveby: shouldn't JarFiles be in a try-with-resources block?


```java
    public void close() throws IOException {
        if (closeRequested) {
            return;
        }
        closeRequested = true;

        synchronized (this) {
            // Close streams, release their inflaters, release cached inflaters
            // and release zip source
            try {
                res.clean();
            } catch (UncheckedIOException ioe) {
                throw ioe.getCause();
            }
        }
    }
```

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/ModuleExports.java
carterkozak: Good call, updated!

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/ModuleExports.java
carterkozak: Good shout, we should have an error-prone rule for this!

PR #1221, opened 2021-11-03
Module export discovery handles jars with no manifest gracefully

carterkozak:
## Before this PR
NPE!

## After this PR
==COMMIT_MSG==
Module export discovery handles jars with no manifest gracefully
==COMMIT_MSG==


carterkozak:
:+1:

PR #1223, opened 2021-11-04
Support `--add-opens` in addition to `--add-exports` based on classpath

carterkozak:
==COMMIT_MSG==
Support `--add-opens` in addition to `--add-exports` based on classpath
==COMMIT_MSG==


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/ModuleArgs.java
```diff
@@ -0,0 +1,142 @@
...
+    // Exists for backcompat until infrastructure has rolled out with Add-Exports manifest values.
+    // Support safepoint metrics from the internal sun.management package in production. We prefer not
+    // to use '--illegal-access=permit' so that we can avoid unintentional and unbounded illegal access
+    // that we aren't aware of.
+    private static final ImmutableList<String> DEFAULT_EXPORTS = ImmutableList.of("java.management/sun.management");
+
+    static ImmutableList<String> collectClasspathArgs(
+            Project project, JavaVersion javaVersion, FileCollection classpath) {
+        // --add-exports is unnecessary prior to java 16
+        if (javaVersion.compareTo(JavaVersion.toVersion("16")) < 0) {
```
fawind, line 59: Just double checking, this is also true for `add-opens`?

fawind (approved):

PR #1280, opened 2022-01-26
Ensure object mappers are configured with ignore unknown for forward compatibility

CRogers:
## Before this PR
Currently, we read json values with object mappers that do not ignore unknown properties. This means it is hard/impossible to add new fields to objects in the recommened-product-dependencies plugin (for example: to mark pdeps as optional in libraries).

## After this PR

==COMMIT_MSG==
Ignore unknown properties in serialized data to support forward compatibility.
==COMMIT_MSG==


gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/RecommendedProductDependencies.java
```diff
@@ -29,6 +30,7 @@
 @Value.Style(jdkOnly = true)
 @JsonSerialize(as = ImmutableRecommendedProductDependencies.class)
 @JsonDeserialize(as = ImmutableRecommendedProductDependencies.class)
+@JsonIgnoreProperties(ignoreUnknown = true)
```
CRogers, line 33: I've done these as well as internal repos use these objects with their own object mappers (this will also complicate the rollout of any serialisation change...)

carterkozak:
Don't we use the client objectmappers which ignore unknowns by default? In some cases we may wish to fail on unknowns (for validation)

CRogers:
@carterkozak for deserialising these, we do not use client/conjure object mappers. I also can't think of a place we want to fail with new properties - it effectively means we can't change anything?

carterkozak:
I think the default mapper does what we want, but the cases where we want to validate against unknowns likely aren't worth a risk of failing.
:+1:

CRogers:
@carterkozak The default object mapper unfortunately doesn't ignore unknown properties - I sadly know this due to former $internal-product I worked on in pre-conjure times that failed to set ignoreUnknown everywhere, so nothing could ever change.

[Here is the a test that verifies that off develop](https://app.circleci.com/pipelines/github/palantir/sls-packaging/1721/workflows/d5f94b83-bd91-4021-af28-cf2b1e25b5f4/jobs/14746/tests):

```
Caused by: com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException:
Unrecognized field "something-else" (class com.palantir.gradle.dist.ImmutableRecommendedProductDependencies$Json),
not marked as ignorable (one known property: "recommended-product-dependencies"])
```

PR #1282, opened 2022-01-27
Fix #1230 Increase default JFR stack depth to 256 frames (from 64)

### LINKED ISSUE #1230: Increase default JFR stack depth beyond the default (conservative) 64 frames
## What happened?

2021-11-11_510x195_screen

I'd like to capture better profiling information when jfr events are collected. Perhaps `-XX:FlightRecorderOptions=stackdepth=256`

carterkozak:
Fix #1230
==COMMIT_MSG==
Increase default JFR stack depth to 256 frames (from 64)
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
@@ -70,7 +70,11 @@ public abstract class LaunchConfigTask extends DefaultTask {
             // Set DNS cache TTL to 20s to account for systems such as RDS and other
             // AWS-managed systems that modify DNS records on failover.
             "-Dsun.net.inetaddr.ttl=20",
-            "-XX:NativeMemoryTracking=summary");
+            "-XX:NativeMemoryTracking=summary",
+            // Increase default JFR stack depth beyond the default (conservative) 64 frames.
+            // This can be overridden by user-provided options.
+            // See sls-packaging#1230
+            "-XX:FlightRecorderOptions=stackdepth=256");
 
     // Reduce memory usage for some versions of glibc.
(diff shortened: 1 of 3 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -70,7 +70,11 @@
             // Set DNS cache TTL to 20s to account for systems such as RDS and other
             // AWS-managed systems that modify DNS records on failover.
             "-Dsun.net.inetaddr.ttl=20",
-            "-XX:NativeMemoryTracking=summary");
+            "-XX:NativeMemoryTracking=summary",
+            // Increase default JFR stack depth beyond the default (conservative) 64 frames.
+            // This can be overridden by user-provided options.
+            // See sls-packaging#1230
+            "-XX:FlightRecorderOptions=stackdepth=256");
```
carterkozak, line 77: I considered a separate optional integer property to opt out of the setting or provide a different value, but I don't think it will be common enough to modify to be worthwhile.

schlosna: Agreed

schlosna (approved):

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
fawind: I guess the small? downside is that capturing a JFR might have a slightly larger perf overhead? But going from 64 to 256 sounds reasonable and when you use JFR you probably care most about the additional info gained through this.

PR #1290, opened 2022-02-10
Read go-java-launcher version from properties

fawind:
## After this PR
Move the go-java-launcher version to a properties file. This will make it easier for us to automatically update this via excavator.

==COMMIT_MSG==
Read go-java-launcher version from properties
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/resources/go-java-version.properties
+++ b/gradle-sls-packaging/src/main/resources/go-java-version.properties
@@ -1,0 +1,3 @@
+# Version of https://github.com/palantir/go-java-launcher to use. Kept up-to-date by excavator.
+# If you move this file, please make sure to update the respective excavator.
+go-java-version=1.17.0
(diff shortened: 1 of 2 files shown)


gradle-sls-packaging/src/main/resources/go-java-version.properties
```diff
@@ -0,0 +1,3 @@
+# Version of https://github.com/palantir/go-java-launcher to use. Kept up-to-date by excavator.
+# If you move this file, please make sure to update the respective excavator.
+go-java-version=1.17.0
```
ferozco, line 3: should we just get people to declare the version of go-java-launcher in versions.props like we do for gofigure and sls-validation?

fawind: Agree that this is nicer. I started writing this first put then got put off as it would require a migration of existing consumers to add the version prop. However on a second thought, it's probably the same effort as writing the excavator that updates this property.
Will try to take a stab at this next week!

fawind: Added in https://github.com/palantir/sls-packaging/pull/1294 for future reference

ferozco: FWIW you can set a default dependency to keep upgrades flowing and then roll out the versions.props change separately/async

PR #1292, opened 2022-02-14
Update go-java-launcher to 1.18.0 (was 1.17.0)

fawind:
## Before this PR
Flup on https://github.com/palantir/sls-packaging/pull/1289. We actually need 1.18.0 to fix CVE-2022-23772 and CVE-2022-23806 as this version updates go to 1.17.7 (https://github.com/palantir/go-java-launcher/pull/197).

## After this PR

==COMMIT_MSG==
Update go-java-launcher to 1.18.0 (was 1.17.0)
==COMMIT_MSG==

PR #1294, opened 2022-02-15
Get go-java-launcher version from versions.props

fawind:
## Before this PR
The go-java-launcher version was hardcoded, forcing us to manually update this every time we want to include a newer version of the go-java-launcher ([example](https://github.com/palantir/sls-packaging/pull/1292)).

## After this PR

We can declare the version used through gradle. In most cases, this should be done through GCV by adding a pin to `versions.props`:

```properties
# versions.props
com.palantir.launching:* = 1.18.0
```

If the pin is not provided, we fall back to a hardcoded version. We might remove this feature in the future and require people to declare the version in `versions.props`.

This was actually possible before, so this PR is primarily adding a test for this behavior.

==COMMIT_MSG==
Get go-java-launcher version from versions.props
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
@@ -56,6 +56,8 @@ import org.gradle.process.CommandLineArgumentProvider;
 import org.gradle.util.GradleVersion;
 
 public final class JavaServiceDistributionPlugin implements Plugin<Project> {
+    // Used as fallback version if no higher version is specified in 'versions.props'.
+    private static final String FALLBACK_GO_JAVA_VERSION = "1.18.0";
     private static final String GO_JAVA_LAUNCHER = "com.palantir.launching:go-java-launcher";
     private static final String GO_INIT = "com.palantir.launching:go-init";
     public static final String GROUP_NAME = "Distribution";
@@ -105,11 +107,6 @@ public final class JavaServiceDistributionPlugin implements Plugin<Project> {
         project.getDependencies().add(launcherConfig.getName(), getGoJavaLauncherCoordinate(project, GO_JAVA_LAUNCHER));
         Configuration initConfig = project.getConfigurations().create("goInitBinary");
         project.getDependencies().add(initConfig.getName(), getGoJavaLauncherCoordinate(project, GO_INIT));
-
-        project.getRepositories().mavenCentral(m -> {
-            m.setName("go-java-launcher-repository");
-            m.content(content -> content.onlyForConfigurations(launcherConfig.getName(), initConfig.getName()));
-        });
 
         TaskProvider<Copy> copyLauncherBinaries = project.getTasks()
                 .register("copyLauncherBinaries", Copy.class, task -> {
@@ -342,10 +339,10 @@ public final class JavaServiceDistributionPlugin implements Plugin<Project> {
         Files.writeString(windowsScript, cleanedText);
     }
 
-    /** To make our unit-test setup simpler, we allow hard-coding a go-java-launcher version, which circumvents GCV. */
+    /** To make our unit-test setup simpler, we allow hard-coding a specific go-java-launcher version. */
     private static String getGoJavaLauncherCoordinate(Project project, String coordinate) {
         if (!project.hasProperty(TEST_GO_JAVA_LAUNCHER_VERSION_OVERRIDE_PROP)) {
-            return coordinate;
+            return coordinate + ":" + FALLBACK_GO_JAVA_VERSION;
         }
         String versionOverride =
                 project.property(TEST_GO_JAVA_LAUNCHER_VERSION_OVERRIDE_PROP).toString();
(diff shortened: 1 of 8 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
```diff
@@ -99,9 +102,15 @@ public void apply(Project project) {
 
         // Create configuration to load executable dependencies
         Configuration launcherConfig = project.getConfigurations().create("goJavaLauncherBinary");
-        project.getDependencies().add(launcherConfig.getName(), GO_JAVA_LAUNCHER);
+        project.getDependencies().add(launcherConfig.getName(), getGoJavaLauncherCoordinate(project, GO_JAVA_LAUNCHER));
         Configuration initConfig = project.getConfigurations().create("goInitBinary");
-        project.getDependencies().add(initConfig.getName(), GO_INIT);
+        project.getDependencies().add(initConfig.getName(), getGoJavaLauncherCoordinate(project, GO_INIT));
+
+        project.getRepositories().mavenCentral(m -> {
```
ferozco: any reason we need this instead of relying on whatever repository the project has configured? I can imagine this could be problematic for internal usage since we don't want to be loading directly from maven central

fawind: You are right, it's actually not needed and will clash with internal usage. Will remove!

fawind:
> any reason we need this instead of relying on whatever repository the project has configured? I can imagine this could be problematic for internal usage since we don't want to be loading directly from maven central

From [here](https://github.com/palantir/sls-packaging/pull/1294#discussion_r806850785): Thinking about this more, I agree that this is preferable compared to breaking and forcing consumers to adopt the versions.props pin. We can add the hard-requirement at a later time if wanted but right now we probably want to optimize for a smooth roll-out. Will update accordingly!

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
```diff
@@ -99,9 +104,10 @@ public void apply(Project project) {
 
         // Create configuration to load executable dependencies
         Configuration launcherConfig = project.getConfigurations().create("goJavaLauncherBinary");
-        project.getDependencies().add(launcherConfig.getName(), GO_JAVA_LAUNCHER);
+        project.getDependencies().add(launcherConfig.getName(), getGoJavaLauncherCoordinate(project, GO_JAVA_LAUNCHER));
         Configuration initConfig = project.getConfigurations().create("goInitBinary");
-        project.getDependencies().add(initConfig.getName(), GO_INIT);
+        project.getDependencies().add(initConfig.getName(), getGoJavaLauncherCoordinate(project, GO_INIT));
```
fawind, line 109: Not sure there is a better way for providing a default fallback version. Right now, we add a pin on `go-java-launcher:1.18.0` which acts as a lower version bound and then if consumers add a higher pin to `versions.props`, it gets overwritten.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
fawind: removed

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
CRogers: This makes sense.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
```diff
@@ -332,4 +338,15 @@ private static void replaceManifestClasspath(Path windowsScript, String manifest
 
         Files.writeString(windowsScript, cleanedText);
     }
+
+    /** To make our unit-test setup simpler, we allow hard-coding a specific go-java-launcher version. */
+    private static String getGoJavaLauncherCoordinate(Project project, String coordinate) {
+        if (!project.hasProperty(TEST_GO_JAVA_LAUNCHER_VERSION_OVERRIDE_PROP)) {
```
CRogers: The name of this confused me a bit - thought we were overriding the version set by GCV at first. Maybe it makes sense to call this `TEST_GO_JAVA_LAUNCHER_FALLBACK_VERSION_OVERRIDE`? To make it obvious that it's about the fallback version.

CRogers (approved):

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
fawind: Makes sense, updated.

PR #1349, opened 2022-05-11
Make ModuleArgs expand manifest classpath JARs for exports and opens

mglazer:
## Before this PR

We only read the JARs on the classpath. This was
fine if we didn't use manifest classpath JARs.

After this PR: We now read the JARs in the manifest classpath JAR
and then include those in our search for opens and exports.

## After this PR

==COMMIT_MSG==
Make ModuleArgs expand manifest classpath JARs for exports and opens
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
@@ -306,6 +306,11 @@ public final class JavaServiceDistributionPlugin implements Plugin<Project> {
 
         project.afterEvaluate(_p -> launchConfigTask.configure(task -> {
             task.getJavaAgents().setFrom(javaAgentConfiguration);
+            task.getFullClasspath()
+                    .from(jarTask.get()
+                            .getOutputs()
+                            .getFiles()
+                            .plus(distributionExtension.getProductDependenciesConfig()));
             task.getClasspath()
                     .from(
                             distributionExtension.getEnableManifestClasspath().get()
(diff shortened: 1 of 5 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -157,6 +157,15 @@ public final MapProperty<String, String> getEnv() {
     @InputFiles
     public abstract ConfigurableFileCollection getClasspath();
 
+    /**
+     * The difference between fullClasspath and classpath is that classpath is what is written
+     * to the launcher-static.yml file, this <i>might</i> in some cases be the manifest classpath
+     * JAR. Full Classpath on the other hand is always going to be the full set of JARs which may
+     * be the same as classpath if manifest classpath JARs are not used.
```
carterkozak, line 164: I appreciate the javadoc :-)

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
```diff
@@ -306,6 +306,11 @@ public void execute(Task _task) {
 
         project.afterEvaluate(_p -> launchConfigTask.configure(task -> {
             task.getJavaAgents().setFrom(javaAgentConfiguration);
+            task.getFullClasspath()
+                    .from(jarTask.get()
+                            .getOutputs()
+                            .getFiles()
+                            .plus(distributionExtension.getProductDependenciesConfig()));
```
carterkozak: minor nit: This duplicates the ternary else component on 318-322, since we're always building that provider now, we might as well build a single provider and pass it to both places. I think we always want the value to be identical.

carterkozak (approved):
lgtm, one suggestion.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
mglazer: Cool, wasn't sure if this was expensive or not...easy to fix.

PR #1365, opened 2022-07-29
Add the `--enable-preview` flag based on the Baseline-Enable-Preview jar manifest attribute

iamdanfox:
## Before this PR

I've recently (https://github.com/palantir/gradle-baseline/pull/2322) added functionality to gradle-baseline so that folks can opt in to java's `--enable-preview` functionality by setting something like `javaVersions { distributionTarget = '17_PREVIEW' }`.  Baseline applies the flag to all the compile time tasks, but we _also_ need this to propagate to the production runtime jvm args.

We decided to convey the information using a jar attribute like `Baseline-Enable-Preview: 17`.  This way if different jars tried to select conflicting runtime values we can provide clear errors.

See https://openjdk.org/jeps/12 for details of the behaviour of `--enable-preview`.

## After this PR
==COMMIT_MSG==
`createLaunchConfig` task automatically adds the `--enable-preview` flag based on the `Baseline-Enable-Preview` jar manifest attribute
==COMMIT_MSG==


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/ModuleArgs.java
```diff
@@ -92,23 +100,40 @@ static ImmutableList<String> collectClasspathArgs(
...
-                .orElseGet(ImmutableList::of);
+    private static Stream<String> enablePreview(
+            JavaVersion javaVersion, Map<File, JarManifestModuleInfo> parsedJarManifests) {
+        Map<String, Collection<String>> enablePreviewFromJar = parsedJarManifests.entrySet().stream()
+                .filter(entry -> entry.getValue().enablePreview().isPresent())
+                .collect(Multimaps.toMultimap(
+                        entry -> entry.getValue().enablePreview().get(),
+                        entry -> entry.getKey().getName(),
+                        () -> MultimapBuilder.hashKeys().arrayListValues().build()))
+                .asMap();
```
iamdanfox: this is where I'd intuitively reach for streamex and an entrystream, but I think we've found that including too many deps in our gradle plugins causes sad times right?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/ModuleArgs.java
```diff
@@ -92,23 +100,40 @@ static ImmutableList<String> collectClasspathArgs(
...
+
+        if (enablePreviewFromJar.size() == 1) {
+            String enablePreviewVersion = Iterables.getOnlyElement(enablePreviewFromJar.keySet());
+            Preconditions.checkState(
+                    enablePreviewVersion.equals(javaVersion.toString()),
+                    "Runtime java version (" + javaVersion + ") must match version from embedded "
+                            + JarManifestModuleInfo.ENABLE_PREVIEW_ATTRIBUTE + " attribute (" + enablePreviewVersion
+                            + ") from:\n" + Iterables.getOnlyElement(enablePreviewFromJar.values()));
+            return Stream.of("--enable-preview");
+        }
```
carterkozak, line 128: This method is pretty complicated, I think we want to stream through the jars, fail if we see any with ENABLE_PREVIEW_ATTRIBUTE != javaVersion, and add the flag if we've seen any ENABLE_PREVIEW_ATTRIBUTE attrs?

imo the `Map<String, Collection<String>> enablePreviewFromJar` makes things difficult to grok

iamdanfox: Sorry for the delay, I've updated to use less stream-y stuff... think this should be a bit more readable now!

PR #1370, opened 2022-08-03
Disable C2 compiler for Java 11

atrocities:
Disable the C2 compiler using the -XX:TieredStopAtLevel=1 flag. This prevents a potential memory leak issue https://bugs.openjdk.org/browse/JDK-8291665

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
@@ -61,6 +61,7 @@ public abstract class LaunchConfigTask extends DefaultTask {
     private static final ImmutableList<String> java15Options =
             ImmutableList.of("-XX:+UnlockDiagnosticVMOptions", "-XX:+ExpandSubTypeCheckAtParseTime");
     private static final ImmutableList<String> disableBiasedLocking = ImmutableList.of("-XX:-UseBiasedLocking");
+    private static final ImmutableList<String> disableC2 = ImmutableList.of("-XX:TieredStopAtLevel=1");
 
     private static final ImmutableList<String> alwaysOnJvmOptions = ImmutableList.of(
             "-XX:+CrashOnOutOfMemoryError",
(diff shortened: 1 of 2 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -191,6 +192,12 @@ public final void createConfig() throws IOException {
                         .addAllJvmOpts(javaAgentArgs())
                         .addAllJvmOpts(alwaysOnJvmOptions)
                         .addAllJvmOpts(addJava8GcLogging.get() ? java8gcLoggingOptions : ImmutableList.of())
+                        // Java 11u6 and later have potential memory leak issues when using the C2
```
carterkozak: I'm not familiar with u-based versioning in newer java releases (google search doesn't give me many hits for it)
```suggestion
                        // Java 11.0.16 introduced a potential memory leak issues when using the C2
```

carterkozak:
Thanks, Yun! :+1:

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -61,6 +61,7 @@ public abstract class LaunchConfigTask extends DefaultTask {
     private static final ImmutableList<String> java15Options =
             ImmutableList.of("-XX:+UnlockDiagnosticVMOptions", "-XX:+ExpandSubTypeCheckAtParseTime");
     private static final ImmutableList<String> disableBiasedLocking = ImmutableList.of("-XX:-UseBiasedLocking");
+    private static final ImmutableList<String> disableC2 = ImmutableList.of("-XX:TieredStopAtLevel=1");
```
iamdanfox, line 64: hang on guys - if you're just trying to turn off the C2 compiler, shouldn't this say `-XX:TieredStopAtLevel=3`?? I think right now you're limiting the JVM to "simple" C1 and preventing it from doing "limited C1" and "Full C1" ??

https://www.baeldung.com/jvm-tiered-compilation#4-level-3--full-c1-compiledcode

https://community.oracle.com/tech/developers/discussion/4499712/jvm-c1-c2-compiler-thread-high-cpu-consumption

image

carterkozak: Try some benchmarks ;-)

carterkozak: I’ve forwarded some info internally

iamdanfox: Fascinating! Is it worth calling out the expected performance impact here though?? If an application was CPU bound then disabling C2 could make it run ~10x slower than before? Although maybe IO bound stuff wouldn't see as much of a slowdown?

iamdanfox: For any future readers - we found a better way of doing this:

```
-XX:CompileCommand=exclude,sun/security/ssl/SSLEngineInputRecord.decodeInputRecord
```

https://github.com/palantir/sls-packaging/pull/1372

PR #1371, opened 2022-08-04
Upgrade Gradle wrapper to the latest version

xRuiAlves:
## Before this PR
Upgrading to gradle wrapper version `7.5`.

After a lot of digging up and trying to make sense of how attribute matching works in variant disambiguation, looks like there were changes in gradle `>=7.4` that make variant resolution a bit different.

BLUF is that a custom disambiguation rule we added for compat with clients on older versions is causing trouble in variant resolution on consumers after gradle `7.4`. Removing this custom rule solves the issue, and I don't see / wasn't able to find a reason for us to have it.

## After this PR
==COMMIT_MSG==
Gradle wrapper is updated to latest version `7.5`
==COMMIT_MSG==


xRuiAlves:
Was doing this change on https://github.palantir.build/excavator-checks/excavator-gradle/pull/3417 as well, but excavator pushed over my commits 😬 Moved the change request to this PR instead

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/SlsBaseDistPlugin.java
```diff
@@ -50,27 +46,5 @@ public final void apply(Project project) {
...
                 .attribute(Usage.USAGE_ATTRIBUTE, project.getObjects().named(Usage.class, SLS_DIST_USAGE));
-
-        project.getDependencies()
-                .getAttributesSchema()
-                .attribute(Usage.USAGE_ATTRIBUTE, strategy -> strategy.getDisambiguationRules()
-                        .add(SlsDisambiguationRule.class));
-    }
-
-    /**
-     * Still support old consumers which don't declare a required usage, such as gradle-docker's docker configuration.
```
CRogers, line 61: This is a big wat - no idea what this is actually doing. Since we don't really use gradle-docker internally, at least not for anything that matters, I'm ok attempting to remove this.

CRogers (approved):

PR #1376, opened 2022-08-06
Improve the overall fidelity of JFR profiles and reduce sampling bias

schlosna:
## Before this PR
sls-packaging did not enable `DebugNonSafepoints` by default, so JFR profiles may exhibit [safepoint sampling bias](https://psy-lob-saw.blogspot.com/2016/02/why-most-sampling-java-profilers-are.html) for method profiling.

## After this PR
==COMMIT_MSG==
Improve the overall fidelity of JFR profiles and reduce safepoint sampling bias

Note due to https://bugs.openjdk.org/browse/JDK-8201516 there may still
be potentially small incorrect attribution, but overall this tradeoff is
worth the sampling bias impact.

See references:
* http://hirt.se/blog/?p=609
* https://docs.oracle.com/javacomponents/jmc-5-5/jfr-runtime-guide/about.htm#JFRRT111
* https://psy-lob-saw.blogspot.com/2016/02/why-most-sampling-java-profilers-are.html
* https://jpbempel.github.io/2022/06/22/debug-non-safepoints.html
==COMMIT_MSG==

## Possible downsides?
Enables diagnostic JVM options by default now, and also JFR profiles may exhibit some incorrect attribution due to https://bugs.openjdk.org/browse/JDK-8201516


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -74,6 +74,11 @@ public abstract class LaunchConfigTask extends DefaultTask {
             // AWS-managed systems that modify DNS records on failover.
             "-Dsun.net.inetaddr.ttl=20",
             "-XX:NativeMemoryTracking=summary",
+            // Improve the overall fidelity of JFR profiles and reduce sampling bias,
+            // see http://hirt.se/blog/?p=609 & https://jpbempel.github.io/2022/06/22/debug-non-safepoints.html
+            // Per https://bugs.openjdk.org/browse/JDK-8201516 may still be potentially incorrect attribution.
+            "-XX:+UnlockDiagnosticVMOptions",
+            "-XX:+DebugNonSafepoints",
```
schlosna, line 81: Consider testing impact of preserving frame pointers for native and kernel traces, though that may be too expensive to enable by default:

* https://github.com/jvm-profiling-tools/async-profiler
* http://techblog.netflix.com/2015/07/java-in-flames.html
* https://bugs.openjdk.org/browse/JDK-8068945


```suggestion
            "-XX:+DebugNonSafepoints",
            "-XX:+PreserveFramePointer",
```

carterkozak: I haven't had a chance to benchmark myself, do you know if there's a measurable runtime cost associated with `+ DebugNonSafepoints`?

schlosna: I haven't had cycles yet to benchmark with/without `-XX:+DebugNonSafepoints` on current JDKs  for our various workflows such as WC infra. The JDK ticket and [Jean-Phillipe Bempel's `Why JVM modern profilers are still safepoint biased?` blog post](https://jpbempel.github.io/2022/06/22/debug-non-safepoints.html) show ~2% overhead. If you have some pointers to spin those WC benchmarks up, I can try to give them a shot to understand potential fleet implications.

Async-profiler enables DebugNonSafepoints via JVMTI agent registering [CompileMethodLoad](https://github.com/jvm-profiling-tools/async-profiler/blob/master/src/profiler.h#L222-L228)

Also worth reading the note http://hirt.se/blog/?p=609 from Marcus Hirt, one of the original JRockit and JFR authors.

carterkozak: Benchmarks don't appear to be impacted (results are within margin of error)

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -74,6 +74,11 @@ public abstract class LaunchConfigTask extends DefaultTask {
             // AWS-managed systems that modify DNS records on failover.
             "-Dsun.net.inetaddr.ttl=20",
             "-XX:NativeMemoryTracking=summary",
+            // Improve the overall fidelity of JFR profiles and reduce sampling bias,
+            // see http://hirt.se/blog/?p=609 & https://jpbempel.github.io/2022/06/22/debug-non-safepoints.html
+            // Per https://bugs.openjdk.org/browse/JDK-8201516 may still be potentially incorrect attribution.
+            "-XX:+UnlockDiagnosticVMOptions",
```
carterkozak, line 80: Some danger opting into this by default: jvm option overrides may rely upon this without realizing, meaning it cannot safely be removed from sls-packaging. What do you think?

schlosna: Yeah, adding this now would make removing any diagnostic VM arg that depends on it a break. We could mitigate a bit with `-XX:+IgnoreUnrecognizedVMOptions` where startup would no longer fail on those args, but we could silently end up in a situation where we assume args are in effect when they're not actually.

schlosna:
@carterkozak curious for your thoughts on this again. I can rebase on develop if interested.

carterkozak:
I think last time we tried this, it produced method and allocation stack traces that pointed near the points we were measuring, but off by enough to cause confusion. It's certainly better to avoid sampling bias, but I'm not sure whether or not it's net positive if the stack traces lead us down rabbit holes.
I'd certainly be willing to try it more broadly for a time to collect more data, the cost of regret should be very low.

PR #1374, opened 2022-08-05
Use Async JVM Unified Logging on JDK 17+

schlosna:
## Before this PR
https://github.com/palantir/sls-packaging/issues/1373

## After this PR

==COMMIT_MSG==
Use Async JVM Unified Logging on JDK 17+

Avoid fflush from Unified JVM Logging (e.g. GC logs), see:

https://aws.amazon.com/blogs/developer/asynchronous-logging-corretto-17/
https://bugs.openjdk.org/browse/JDK-8291898
https://bugs.openjdk.org/browse/JDK-8229517
https://github.com/openjdk/jdk/commit/41185d38f21e448370433f7e4f1633777cab6170
==COMMIT_MSG==

PR #1380, opened 2022-08-11
Enable String Deduplication on JDK 17+

### LINKED ISSUE #1378: Enable UseStringDeduplication where feasible
OpenJDK 17.0.2 fixed https://bugs.openjdk.org/browse/JDK-8277981 so worth re-testing and consider enabling `-XX:+UseStringDeduplication` by default when using supported GC on JDK 17+ to both reduce string heap bloat and streamline comparisons in string heavy services.

https://github.com/openjdk/jdk/blame/8d7958e46969fd59df7bc6dd1cbe31ae63686b05/src/hotspot/share/gc/shared/stringdedup/stringDedupConfig.cpp#L116-L155

schlosna:
## Before this PR
JVM string deduplication was not enabled

## After this PR
Reduce heap size by allowing JVM garbage collectors to deduplicate strings.
==COMMIT_MSG==
Enable String Deduplication on JDK 17+

G1 and Shenandoah GC support string deduplication as of JDK 17
* https://openjdk.java.net/jeps/192
* https://bugs.openjdk.org/browse/JDK-8264718

Only enable on JDK 17+ due to fix
https://bugs.openjdk.org/browse/JDK-8277981

JDK 18+ enable string deduplication for SerialGC, ParallalGC, and ZGC.
https://malloc.se/blog/zgc-jdk18
==COMMIT_MSG==

Closes https://github.com/palantir/sls-packaging/issues/1378

## Possible downsides?
Enabling this by default may trigger some unknown JDK bug(s).

### DIFF
--- a/changelog/@unreleased/pr-1380.v2.yml
+++ b/changelog/@unreleased/pr-1380.v2.yml
@@ -1,0 +1,19 @@
+type: improvement
+improvement:
+  description: |-
+    Enable String Deduplication on JDK 17+
+
+    G1 and Shenandoah GC support string deduplication as of JDK 17
+    * https://openjdk.java.net/jeps/192
+    * https://bugs.openjdk.org/browse/JDK-8264718
+
+    Only enable on JDK 17+ due to fix
+    https://bugs.openjdk.org/browse/JDK-8277981
+
+    JDK 18+ enable string deduplication for SerialGC, ParallalGC, and ZGC.
+    https://malloc.se/blog/zgc-jdk18
+  links:
+  - https://github.com/palantir/sls-packaging/pull/1380
+  - https://openjdk.java.net/jeps/192
+  - https://bugs.openjdk.org/browse/JDK-8264718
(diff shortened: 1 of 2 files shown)


carterkozak:
Let's pick a couple performance-sensitive canary services and try this out before including in sls-packaging

iamdanfox:
we've run with this one one apollo node for 24 hours and interestingly haven't seen a noticeable reduction in heap yet (but also haven't seen any _downside_ in terms of CPU either).  Perhaps we should trial it on a _really_ performance sensitive service like MP or timelock before blasting to everyone?

carterkozak:
@pkoenig10 Thoughts on giving this a shot in MP?

carterkozak:
Happy to merge this. Currently this branch targets #1374 which is still do-not-merge, up to you how you'd like to sequence things.

schlosna:
I will rebase off develop later tonight/tomorrow and update this to target develop

carterkozak (approved):
Thanks!

PR #1414, opened 2022-12-08
Enable 'UseContainerCpuShares' for java distributions

carterkozak:
https://bugs.openjdk.org/browse/JDK-8281181 was introduced in the October patch.
==COMMIT_MSG==
Enable 'UseContainerCpuShares' for java distributions
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
@@ -225,10 +225,10 @@ public abstract class LaunchConfigTask extends DefaultTask {
                                         : ImmutableList.of())
                         // https://bugs.openjdk.org/browse/JDK-8281181 stopped respecting cpu.shares for
                         // processor count. UseContainerCpuShares can be enabled for the time being, however it
-                        // is deprecated in jdk19 and expired in jdk21: https://bugs.openjdk.org/browse/JDK-8282684
+                        // is deprecated in jdk19 and obsoleted in jdk20: https://bugs.openjdk.org/browse/JDK-8282684
                         .addAllJvmOpts(
                                 javaVersion.get().compareTo(JavaVersion.toVersion("11")) >= 0
-                                                && javaVersion.get().compareTo(JavaVersion.toVersion("20")) <= 0
+                                                && javaVersion.get().compareTo(JavaVersion.toVersion("19")) <= 0
                                         ? forceUseContainerCpuShares
                                         : ImmutableList.of())
                         .addAllJvmOpts(
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
@@ -64,6 +64,8 @@ public abstract class LaunchConfigTask extends DefaultTask {
     // Disable C2 compilation for problematic structure in JDK 11.0.16, see https://bugs.openjdk.org/browse/JDK-8291665
     private static final ImmutableList<String> jdk11DisableC2Compile =
             ImmutableList.of("-XX:CompileCommand=exclude,sun/security/ssl/SSLEngineInputRecord.decodeInputRecord");
+    private static final ImmutableList<String> forceUseContainerCpuShares =
+            ImmutableList.of("-XX:+UseContainerCpuShares");
 
     private static final ImmutableList<String> alwaysOnJvmOptions = ImmutableList.of(
             "-XX:+CrashOnOutOfMemoryError",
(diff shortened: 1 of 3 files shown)


carterkozak:
Leaving this as a draft until I have data from an internal test.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -219,6 +221,14 @@ public final void createConfig() throws IOException {
                                 javaVersion.get().compareTo(JavaVersion.toVersion("15")) < 0
                                         ? disableBiasedLocking
                                         : ImmutableList.of())
+                        // https://bugs.openjdk.org/browse/JDK-8281181 stopped respecting cpu.shares for
+                        // processor count. UseContainerCpuShares can be enabled for the time being, however it
+                        // is deprecated in jdk19 and expired in jdk21: https://bugs.openjdk.org/browse/JDK-8282684
+                        .addAllJvmOpts(
+                                javaVersion.get().compareTo(JavaVersion.toVersion("11")) >= 0
+                                                && javaVersion.get().compareTo(JavaVersion.toVersion("20")) <= 0
```
schlosna: Should we only add `-XX:+UseContainerCpuShares` for JDK [11, 19] as JDK 20.x obsoletes the flag and will accept it as an argument without throwing, but it won't have any effect as it [has already been removed from jdk-20+12](https://github.com/openjdk/jdk/commit/db772276848f6ad2d4d13e892bcd0eb3123d030f) from the code outside https://github.com/openjdk/jdk/blob/jdk-20+26/src/hotspot/share/runtime/arguments.cpp#L552

```suggestion
                                                && javaVersion.get().compareTo(JavaVersion.toVersion("19")) <= 0
```

```
  { "UseContainerCpuShares",        JDK_Version::jdk(19), JDK_Version::jdk(20), JDK_Version::jdk(21) },
```

>      DEPRECATED: An option that is supported, but a warning is printed to let the user know that
>                  support may be removed in the future. Both regular and aliased options may be
>                  deprecated.
>   
>                  Add a deprecation warning for an option (or alias) by adding an entry in the
>                  "special_jvm_flags" table and setting the "deprecated_in" field.
>                  Often an option "deprecated" in one major release will
>                  be made "obsolete" in the next. In this case the entry should also have its
>                  "obsolete_in" field set.
>   
>        OBSOLETE: An option that has been removed (and deleted from globals.hpp), but is still accepted
>                  on the command line. A warning is printed to let the user know that option might not
>                  be accepted in the future.
>   
>                  Add an obsolete warning for an option by adding an entry in the "special_jvm_flags"
>                  table and setting the "obsolete_in" field.
>   
>         EXPIRED: A deprecated or obsolete option that has an "accept_until" version less than or equal
>                  to the current JDK version. The system will flatly refuse to admit the existence of
>                  the flag. This allows a flag to die automatically over JDK releases.
>   
>                  Note that manual cleanup of expired options should be done at major JDK version upgrades:
>                     - Newly expired options should be removed from the special_jvm_flags and aliased_jvm_flags tables.
>                     - Newly obsolete or expired deprecated options should have their global variable
>                       definitions removed (from globals.hpp, etc) and related implementations removed.

carterkozak: Good catch, thanks!

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java
```diff
@@ -64,6 +64,10 @@ public abstract class LaunchConfigTask extends DefaultTask {
     // Disable C2 compilation for problematic structure in JDK 11.0.16, see https://bugs.openjdk.org/browse/JDK-8291665
     private static final ImmutableList<String> jdk11DisableC2Compile =
             ImmutableList.of("-XX:CompileCommand=exclude,sun/security/ssl/SSLEngineInputRecord.decodeInputRecord");
+    // UseContainerCpuShares was added in a patch release, thus IgnoreUnrecognizedVMOptions is required to avoid
+    // breaking distributions running with older JDKs.
+    private static final ImmutableList<String> forceUseContainerCpuShares =
+            ImmutableList.of("-XX:+IgnoreUnrecognizedVMOptions", "-XX:+UseContainerCpuShares");
```
carterkozak, line 70: Updated with `IgnoreUnrecognizedVMOptions` because `UseContainerCpuShares` was added in the patch that changed behavior. It pains me a bit, but it's likely our best option.

oss latest-jdks is a bit behind, and didn't have the flag. Resolved that piece by merging https://github.com/palantir/gradle-jdks-latest/pull/59 but we need to be careful about implicit dependencies on specific jdk builds.

schlosna (commented):
Feel free to add the merge back when you're ready

carterkozak:
Internal results look great, pushing this through!

schlosna (approved):

PR #1416, opened 2022-12-13
Revert #1414: Stop adding -XX:+UseContainerCpuShares

iamdanfox:
## Before this PR

sls-packaging 7.25.0 introduced this new option: https://github.com/palantir/sls-packaging/pull/1414. Rolling it out has is correlated with some significant performance changes (e.g. PDS-320010 for timelock, and another thread ing #dev-backend-infra), which makes me concerned this might introduce a wider risk of problems over the break.

## After this PR
==COMMIT_MSG==
Revert #1414: Stop adding -XX:+UseContainerCpuShares
==COMMIT_MSG==

## Possible downsides?
- i don't have conclusive proof that this JVM option was problematic for timelock

### DIFF
--- a/changelog/@unreleased/pr-1416.v2.yml
+++ b/changelog/@unreleased/pr-1416.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Stop automatically adding `-XX:+UseContainerCpuShares` (introduced
+    in sls-packaging 7.25.0) to allow for a more conservative rollout.
+  links:
+  - https://github.com/palantir/sls-packaging/pull/1416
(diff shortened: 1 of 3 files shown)


iamdanfox:
cc @schlosna for SA

schlosna:
I think we should hold on this for now

carterkozak (changes requested):
Agreed, we have positive data from this change, but only one dubious data-point against it.

iamdanfox:
Ditching this because the specific installation I saw a perf hit on was _not_ running inside a container, so the correlation looks v. unlikely to be causation!

PR #1417, opened 2022-12-13
Annotate ConfigureProductDependenciesTask inputs

carterkozak:
==COMMIT_MSG==
Annotate ConfigureProductDependenciesTask inputs to ensure task dependencies are set as expected
==COMMIT_MSG==


bjlaub (approved):
lgtm

PR #1434, opened 2023-02-10
Parallelise some tasks

CRogers:
## Before this PR
None of the custom task types in this repo use the worker api - this means they cannot be parallelised within the same project and even worse, they can hold up longer parallelisable tasks (like `distTar`) from being run in parallel.

## After this PR
Use [auto-parallelizable](https://github.com/palantir/auto-parallelizable) to make some of these tasks run in parallel.

==COMMIT_MSG==
`createLaunchConfig`, `mergeDiagnosticsJson`, `createInitScript` and `createCheckScript` can all now run in parallel within the same project.
==COMMIT_MSG==


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/ModuleArgs.java
```diff
@@ -40,6 +41,7 @@
  * plugin <a href="https://github.com/palantir/gradle-baseline/pull/1944">gradle-baseline#1944</a>.
  */
 final class ModuleArgs {
+    private static final Logger log = Logging.getLogger(ModuleArgs.class);
```
felixdesouza, line 44: does gradle route this tot he project logger here? not sure how it would :S

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/util/EmitFiles.java
```diff
@@ -40,7 +42,11 @@ public static Path replaceVars(InputStream src, Path dest, Map<String, String> v
         dest.getParent().toFile().mkdirs();
 
         // write content
-        return Files.write(dest, text.getBytes(StandardCharsets.UTF_8));
+        try {
```
felixdesouza, line 45: I was thinking, what if we relaxed the "doesn't throw exception" constraint? if we were writing tasks normally, with the `@TaskAction` we could just slap a `throws Exception` on the end and everything would be fine. In the generated task, we just wrap invocations with a try catch? thoughts?

felixdesouza: only because it just makes code a bit bloated and distracting / or you have to make a million "unsafe write" type methods

felixdesouza:
looks good 👍

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/ModuleArgs.java
CRogers: Unfortunately there's no injectable logger you can use. This is the only real ways I found. https://github.com/gradle/gradle/issues/2678

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/util/EmitFiles.java
CRogers: I think the only downside would be another layer of rethrown exception in the stacktrace that does not have any useful information.

CRogers: I guess the RuntimeException I've made below is not really that useful either. But the autowrapper exception would be something like "Task LaunchConfigTask failed"

PR #1436, opened 2023-02-13
MergeDiagnosticJson should not be build cacheable

CRogers:
## Before this PR
`mergeDiagnosticJson` is cacheable, meaning that Gradle will spend 5 seconds looking for a cache entry to save doing ~10-20 milliseconds of actual work (given the jars it's scanning through is the runtime classpath, which is always needed and already cached). This is quite a common sight:

Screenshot 2023-02-13 at 17 50 52

## After this PR
==COMMIT_MSG==
`mergeDiagnosticJson` is no longer build cacheable, as it spends 100x more time accessing the cache than just running the task
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/MergeDiagnosticsJsonTask.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/MergeDiagnosticsJsonTask.java
@@ -16,7 +16,4 @@
 
 package com.palantir.gradle.dist.service;
 
-import org.gradle.api.tasks.CacheableTask;
-
-@CacheableTask
 public abstract class MergeDiagnosticsJsonTask extends MergeDiagnosticsJsonTaskImpl {}
--- a/changelog/@unreleased/pr-1436.v2.yml
+++ b/changelog/@unreleased/pr-1436.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: '`mergeDiagnosticJson` is no longer build cacheable, as it spends 100x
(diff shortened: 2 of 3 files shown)


carterkozak (approved):

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/MergeDiagnosticsJsonTask.java
```diff
@@ -16,7 +16,4 @@
 
 package com.palantir.gradle.dist.service;
 
-import org.gradle.api.tasks.CacheableTask;
-
-@CacheableTask
 public abstract class MergeDiagnosticsJsonTask extends MergeDiagnosticsJsonTaskImpl {}
```
carterkozak, line 19: Might be worth a comment so we don't attempt to re-"optimize" this in the future!

CRogers: hmm, I think in that case we would end up putting that comment on every single simple task we write...

felixdesouza: you can go even better: `@DisableCachingByDefault(because = "reasons")`, you see it on the `Copy` task.

carterkozak: I don't think it would be entirely unreasonable to explicitly declare whether each task is cacheable or not, but I defer to your judgement.

PR #1443, opened 2023-03-02
Discover all diagnostic manifests

pkoenig10:
Our internal diagnostic annotation processor generates the diagnostic manifest in the class directory, not the resource directory.

I've confirmed that this works on one of our internal projects.

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/DiagnosticsManifestPlugin.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/DiagnosticsManifestPlugin.java
@@ -17,7 +17,6 @@
 package com.palantir.gradle.dist.service;
 
 import com.palantir.gradle.dist.artifacts.ExtractFileFromJar;
-import com.palantir.gradle.dist.artifacts.SelectSingleFile;
 import com.palantir.gradle.dist.tasks.CreateManifestTask;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
@@ -38,7 +37,7 @@ public final class DiagnosticsManifestPlugin implements Plugin<Project> {
     static final String mergeDiagnosticsJson = "mergeDiagnosticsJson";
 
     private static final String FILE_TO_EXTRACT = "sls-manifest/diagnostics.json";
-    private static final String MY_ATTRIBUTE = "extracted-" + FILE_TO_EXTRACT;
+    private static final String DIAGNOSTICS = "sls-diagnostics";
 
     /**
      * This plugin uses a few slightly 'advanced' gradle features:
@@ -58,18 +57,17 @@ public final class DiagnosticsManifestPlugin implements Plugin<Project> {
     @Override
     @SuppressWarnings("RawTypes")
     public void apply(Project project) {
-        configureExternalDependencyTransform(project);
-        configureProjectDependencyTransform(project);
+        configureDiagnosticsTransform(project);
 
         Configuration consumableRuntimeConfiguration = createConsumableRuntimeConfiguration(project);
         ArtifactView attributeSpecificArtifactView = consumableRuntimeConfiguration
                 .getIncoming()
                 .artifactView(v -> {
-                    v.getAttributes().attribute(artifactType, MY_ATTRIBUTE);
+                    v.getAttributes().attribute(artifactType, DIAGNOSTICS);
                     v.getAttributes()
                             .attribute(
                                     LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE,
-                                    project.getObjects().named(LibraryElements.class, MY_ATTRIBUTE));
+                                    project.getObjects().named(LibraryElements.class, DIAGNOSTICS));
                 });
 
         TaskProvider<MergeDiagnosticsJsonTask> mergeDiagnosticsTask = project.getTasks()
@@ -95,13 +93,15 @@ public final class DiagnosticsManifestPlugin implements Plugin<Project> {
         });
     }
 
-    private static void configureExternalDependencyTransform(Project project) {
+    private static void configureDiagnosticsTransform(Project project) {
         project.getDependencies().registerTransform(ExtractFileFromJar.class, details -> {
             details.getParameters().getPathToExtract().set(FILE_TO_EXTRACT);
 
             details.getFrom().attribute(artifactType, ArtifactTypeDefinition.JAR_TYPE);
-            details.getTo().attribute(artifactType, MY_ATTRIBUTE);
+            details.getTo().attribute(artifactType, DIAGNOSTICS);
 
+            // It's not strictly necessary to define a mapping for the 'org.gradle.libraryelements' attribute,
+            // but it looks a bit weird for something to still be marked as a 'jar' when it's clearly not.
             details.getFrom()
                     .attribute(
                             LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE,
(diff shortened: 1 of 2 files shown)


carterkozak (dismissed):
Beautiful, thank you, @pkoenig10!

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/DiagnosticsManifestPlugin.java
```diff
@@ -117,7 +117,7 @@ private static void configureProjectDependencyTransform(Project project) {
         project.getDependencies().registerTransform(SelectSingleFile.class, details -> {
             details.getParameters().getPathToExtract().set(FILE_TO_EXTRACT);
 
-            details.getFrom().attribute(artifactType, ArtifactTypeDefinition.JVM_RESOURCES_DIRECTORY);
```
carterkozak, line 120: Will this removal impact folks who check in manually-generated resource manifests?

carterkozak: We may want to set up two transforms, one for each

pkoenig10: Yeah I'm trying to figure this out.

I had tried setting both here - poking around the Gradle code it looked like they would be combined. But that didn't appear to work.

https://github.com/gradle/gradle/blob/5d5de431f8d0c53f2c3e732753a066362eac2b20/subprojects/core/src/main/java/org/gradle/api/internal/attributes/DefaultMutableAttributeContainer.java#L58-L63

https://github.com/gradle/gradle/blob/5d5de431f8d0c53f2c3e732753a066362eac2b20/subprojects/core/src/main/java/org/gradle/api/internal/attributes/DefaultImmutableAttributesFactory.java#L98-L110

Might have to poke at this more.

pkoenig10: Yeah it's last write wins. So if you have

```
details.getFrom().attribute(artifactType, ArtifactTypeDefinition.JVM_CLASS_DIRECTORY);
details.getFrom().attribute(artifactType, ArtifactTypeDefinition.JVM_RESOURCES_DIRECTORY);
```

then it only picks up manfiests from the resources directory.

Maybe we just need to register a separate transform.

carterkozak:
👍 looks great, feel free to merge+release when you're happy with it. Thanks again!

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/DiagnosticsManifestPlugin.java
```diff
@@ -95,39 +93,23 @@ public void apply(Project project) {
...
+    private static void configureJarTransform(Project project) {
         project.getDependencies().registerTransform(ExtractFileFromJar.class, details -> {
             details.getParameters().getPathToExtract().set(FILE_TO_EXTRACT);
 
             details.getFrom().attribute(artifactType, ArtifactTypeDefinition.JAR_TYPE);
-            details.getTo().attribute(artifactType, MY_ATTRIBUTE);
+            details.getTo().attribute(artifactType, DIAGNOSTICS);
 
+            // It's not strictly necessary to define a mapping for the 'org.gradle.libraryelements' attribute,
+            // but it looks a bit weird for something to still be marked as a 'jar' when it's clearly not.
```
pkoenig10, line 104: I was a bit confused why we set the `org.gradle.libraryelements` attribute, since it didn't seem necessary.

Then I discovered that @iamdanfox's implementation had some good comments in https://github.com/palantir/sls-packaging/pull/1035 that were removed before the PR merged.

https://github.com/palantir/sls-packaging/blob/614f8a265085d3fcdf32e70c81b7c56ef4da9ac0/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/DiagnosticsManifestPlugin.java#L95-L96

Adding this back so future devs understand why we do this.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/DiagnosticsManifestPlugin.java
```diff
@@ -95,39 +93,23 @@ public void apply(Project project) {
...
                             LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE,
                             project.getObjects().named(LibraryElements.class, LibraryElements.JAR));
             details.getTo()
                     .attribute(
                             LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE,
-                            project.getObjects().named(LibraryElements.class, MY_ATTRIBUTE));
-        });
-    }
-
-    private static void configureProjectDependencyTransform(Project project) {
```
pkoenig10, line 116: My initial approach did not work because we were registering multiple artifact transforms. Both the resources directory and the classes directory transform would produce a dependency with the variant:

```
artifactType=diagnostics,org.gradle.libraryelements=diagnostics
```

During resolution, Gradle must select one. There are some docs that vaguely describe how this works. In my testing, the resources directory was chosen fairly consistently over the classes directory.
 
https://docs.gradle.org/7.6/userguide/artifact_transforms.html#artifact_transform_selection_and_execution

The only way to make this work would be to define separate attribute values, so the variants are different. For example, a dependency would produce both:
```java
// These are from the resources directory
artifactType=diagnostics,org.gradle.libraryelements=resource-diagnostics

// These are from the class directory
artifactType=diagnostics,org.gradle.libraryelements=class-diagnostics
```

_Interestingly, because a local dependency also produces a JAR, I suspect the JAR transform also conflicts with the resources directory transform._

This solution would also require us to define a separate artifact view for each variant and pass those all into the classpath for the `mergeDiagnosticsTask`, which is a bit gross.

Looking back at code comments from https://github.com/palantir/sls-packaging/pull/1035, it looks like this was done purely as an optimization to avoid needing to create a JAR to extract the diagnostic manifest.

https://github.com/palantir/sls-packaging/blob/614f8a265085d3fcdf32e70c81b7c56ef4da9ac0/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/DiagnosticsManifestPlugin.java#L107-L109

It feels reasonable to just eliminate this optimization, with the following justification:
- It is undesirable to have conflicting artifact transforms
- Removing it makes the plugin logic simpler and more consistent between libraries and local projects
- If we want to use the output from an annotation processor, we're going to have to perform compilation
- Most of the builds that run `createManifest` are probably also going to perform compilation

pkoenig10:
Note that you cannot have a diagnostic manifest in both the resources directory and class directory.

Both files will have the same path in the generated JAR, so only one must be chosen. In my testing, the resource manifest would override the generated one - I assume the resource manifest wins because resources are copied into the classes directory after compilation.

While this may seem undesirable, there's not really an alternative and it makes the behavior more consistent with the behavior when consuming manifest from libraries (ie. we just extract from the JAR).

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/DiagnosticsManifestPlugin.java
carterkozak: I agree, this is entirely reasonable. We have several other components which tie into distribution/manifest creation (e.g. minimum dependency calculation based on referenced rpc methods) which also require compilation.

CRogers: I'd need to check, but my one fear with making createManifest depend on compilation is that dev-env might delay the `resolveSlsConfigurations` task, which will slow down builds by minutes as we can't pull containers early (requires `resolveSlsConfigurations` to be run). It also means `resolveSlsConfigurations` and it successors (`gofigure` etc) cannot be run in parallel with compilation and must come after it.

CRogers: > We have several other components which tie into distribution/manifest creation (e.g. minimum dependency calculation based on referenced rpc methods) which also require compilation.

I'm pretty all the scary code involving artifacts transforms inside this repo was to avoid that.

CRogers: I confirmed by looking at a build scan that `createManifest` does not currently depend on compilation, and this would cause some slowdown in builds.

pkoenig10: I don't see how we can avoid compilation if we want to support the annotation processor.

I don't completely understand how we have this guarantee today. It's not obvious to me what would make Gradle reliably choose the resources directory artifact transform over the JAR artifact transform - I assume we're just getting lucky with a Gradle implementation detail.

Just looking at this code and the [Gradle docs](https://docs.gradle.org/7.6/userguide/artifact_transforms.html#artifact_transform_selection_and_execution), I would have thought that we'd fall into the "selection fails and an error is reported" case. But that's obviously not the case.

> Of all the found transform chains, Gradle tries to select the best one:
> - If there is only one transform chain, it is selected.
> - If there are two transform chains, and one is a suffix of the other one, it is selected.
> - If there is a shortest transform chain, then it is selected.
> - In all other cases, the selection fails and an error is reported.

CRogers: Talking @carterkozak, we're rolling out the conjure per-endpoint minimum version plugin atm, which will force compilation to occur anyway. This will slow down builds, but I think I can make another task in sls-packaging to spit out the pdeps to file, then have dev-env read that rather than `createManifest`. This way we can make builds fast again. So I think this is fine.

PR #1454, opened 2023-03-22
create schema-versions.lock file

rzpt:
## Before this PR

We want to be able to track the supported schema versions of dists, but the information is buried in the manifest, which is not checked in.

## After this PR

==COMMIT_MSG==
Create schema-versions.lock file to track supported schema versions
==COMMIT_MSG==

## Possible downsides?

There will be an extra file to maintain, but --write-locks will handle it.


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/SchemaMigration.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    default int compareTo(SchemaMigration other) {
+        int cmp = Integer.compare(fromVersion(), other.fromVersion());
+        if (cmp != 0) {
+            return cmp;
+        }
+        return type().compareTo(other.type());
+    }
+
+    static SchemaMigration fromObject(Object obj) {
```
ptaug: You should use Jackson annotations to parse the object instead of a `fromObject` method.

Adding:
```java
@JsonDeserialize(as = ImmutableSchemaMigration.class)
@JsonSerialize(as = ImmutableSchemaMigration.class)
```
and 
```java
@JsonProperty("from")
int fromVersion();
```
allows you to delete the `fromObject` method and rely on an ObjectMapper instead.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/CreateManifestTask.java
```diff
@@ -132,8 +155,17 @@ final void createManifest() throws IOException {
...
-    private void requireAbsentLockfile() {
-        File lockfile = getLockfile();
+    private List<SchemaMigration> getSchemaMigrations() {
+        Object raw = getManifestExtensions().get().get("schema-migrations");
+        if (raw == null) {
+            return null;
+        }
+        @SuppressWarnings("unchecked")
+        List<Object> list = (List<Object>) raw;
+        return list.stream().map(SchemaMigration::fromObject).collect(ImmutableList.toImmutableList());
```
ptaug: This can be replaced by an ObjectMapper you can have as a static field.
```java
return OBJECT_MAPPER.convertValue(raw, new TypeReference<>() {});
```
This would use the Jackson annotations from `SchemaMigration`.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/CreateManifestTask.java
```diff
@@ -132,8 +155,17 @@ final void createManifest() throws IOException {
                         .build());
     }
 
-    private void requireAbsentLockfile() {
-        File lockfile = getLockfile();
+    private List<SchemaMigration> getSchemaMigrations() {
+        Object raw = getManifestExtensions().get().get("schema-migrations");
+        if (raw == null) {
+            return null;
```
ptaug: Can you return an empty list instead?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/SchemaMigration.java
```diff
@@ -0,0 +1,46 @@
...
+@Value.Immutable
+public interface SchemaMigration extends Comparable<SchemaMigration> {
+    @Value.Parameter
+    String type();
+
+    @Value.Parameter
+    int fromVersion();
+
+    @Override
+    default int compareTo(SchemaMigration other) {
```
ptaug, line 39: These are two sequential comparators (version -> type). You can use comparator composition instead:
```java
Comparator<SchemaMigration> comparator = Comparator.comparing(SchemaMigration::fromVersion, Integer::compareTo)
  .thenComparing(SchemaMigration::type, String::compareTo);
return comparator.compare(this, other);
```

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/SchemaVersionLockFile.java
```diff
@@ -0,0 +1,57 @@
...
+    public static final String LOCK_FILE = "schema-versions.lock";
+
+    public static String asString(List<SchemaMigration> schemaMigrations) {
+        Preconditions.checkArgument(!schemaMigrations.isEmpty(), "Migrations must not be empty");
+        List<SchemaMigration> sorted = schemaMigrations.stream().sorted().collect(Collectors.toList());
+
+        List<SchemaMigrationRange> ranges = new ArrayList<>();
+        SchemaMigration firstOfRange = sorted.get(0);
+        SchemaMigration lastOfRange = firstOfRange;
+        for (int i = 1; i < sorted.size(); i++) {
```
ptaug: Iterating directly on the objects would be clearer:
```
for (SchemaMigration currentMigration : sorted) {
    ...
}
```
Also, we should aim to have explicit variable names, e.g. `currentMigration` or `current` instead of `cur`.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/SchemaVersionLockFile.java
```diff
@@ -0,0 +1,57 @@
...
+            SchemaMigration cur = sorted.get(i);
+            Preconditions.checkArgument(
+                    cur.fromVersion() != lastOfRange.fromVersion(),
+                    "Multiple migrations with the same from version are not allowed");
+            if (cur.type().equals(lastOfRange.type()) && cur.fromVersion() == lastOfRange.fromVersion() + 1) {
+                lastOfRange = cur;
+            } else {
+                ranges.add(ImmutableSchemaMigrationRange.of(
```
ptaug: We tend to avoid instantiating Immutables directly -- can you create a method `of()` in `SchemaMigrationRange` that returns the interface type?

Similar to `ProductDependencyManifest.class` in this repo.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/SchemaVersionLockFile.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+package com.palantir.gradle.dist;
+
+import com.palantir.logsafe.Preconditions;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public final class SchemaVersionLockFile {
```
ptaug: Could we explore making this an Immutable as well (that holds a List<SchemaMigration>), and generate the file using Jackson serialization somehow? This will make it more reliable to consume for end clients, and will make the serialization itself not rely on manual string construction.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/SchemaVersionLockFileTest.java
```diff
@@ -0,0 +1,81 @@
...
+package com.palantir.gradle.dist;
+
+import com.google.common.collect.ImmutableList;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class SchemaVersionLockFileTest {
```
ptaug: Can you write these as Spock tests? I believe all other tests in the repo use it.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/SchemaVersionLockFileTest.java
```diff
@@ -0,0 +1,81 @@
...
+        migrations.add(ImmutableSchemaMigration.of("online", 102));
+        String str = SchemaVersionLockFile.asString(migrations);
+        Assertions.assertEquals(SchemaVersionLockFile.HEADER + "online [100, 100]\nonline [102, 102]\n", str);
+    }
+
+    @Test
+    void testMultipleRanges() {
+        List<SchemaMigration> migrations = new ArrayList<>();
+        int ver = 100;
+        for (int i = 0; i < 10; i++, ver++) {
```
ptaug: It's a bit hard to read what you construct. Instead of iterating over 10 for each version, could you just write down a manual list of <10 items? This would improve readability of the test, and would not reduce the testing surface (afaict).
```
migrations.add(ImmutableSchemaMigration.of("offline", 101))
migrations.add(ImmutableSchemaMigration.of("online", 102))
migrations.add(ImmutableSchemaMigration.of("online", 103))
migrations.add(ImmutableSchemaMigration.of("offline", 104))
migrations.add(ImmutableSchemaMigration.of("online", 105))
```

ptaug (commented):
Thanks for the PR! Left a few comments. Could you add a short description of what this file will be used for?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/SchemaVersionLockFile.java
rzpt: Made it immutable. I don't think it's worth it to make it jackson serializable.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/SchemaVersionLockFileTest.java
rzpt: I thought we wanted to minimize groovy usage. I can't tell if java is supported by spock. The way this test is written is very typical for all our java tests.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/SchemaVersionLockFile.java
rzpt: Done, but iterated over a sublist.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/SchemaVersionLockFileTest.java
ptaug: In all the Gradle tooling, Spock is the preferred way to write tests (examples in internal gradle-plugins org), as it makes it easy to understand the shape of the files produced by the tooling. Mostly though, it's used everywhere in the repo, so for consistency new tests should use spock.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/SchemaVersionLockFileTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.junit.jupiter.api.Test;
+
+public class SchemaVersionLockFileTest {
+    @Test
+    void testEmpty() {
+        try {
+            SchemaVersionLockFile.of(ImmutableList.of());
+            Assertions.fail();
+        } catch (SafeIllegalArgumentException e) {
+            // expected
```
ptaug: Not relevant if we switch to spock, but we prefer the assertj framework (here, `assertThatThrownBy`).

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/SchemaVersionLockFile.java
```diff
@@ -0,0 +1,69 @@
...
+package com.palantir.gradle.dist;
+
+import com.palantir.logsafe.Preconditions;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface SchemaVersionLockFile {
```
ptaug, line 29: It'd be worthwhile to make it Jackson serializable, imo. YAML would be a good choice if we want something more human-readable than JSON.

The reason the deps lock files are in a custom lock format is because we copied it over from nebula. It also wasn't meant to be deserialized and consumed in tooling (at least initially). Here, we're building these files for the purpose of deserializing them, and with no concerns of backwards-compatibility, so having a custom format creates an additional maintenance load that's avoidable. For example, the revapi file, similar to a lock file, is a YAML file: https://github.com/palantir/conjure-java-runtime/blob/develop/.palantir/revapi.yml.

To make that serialization work OOTB with an Immutable, you can add a `@Value.Derived` attribute, `List<SchemaMigrationRange> migrationRanges`, where the derivation logic copies the current `asString()` logic. And that should make it possible to dump the file as a YAML.

rzpt: Okay, makes sense to me now. I wrote it without using `@Value.Derived`

ptaug (commented):
That looks ok to me now. One final question: Why are we storing ranges, instead of mirroring what's on the manifest? There are legitimate reasons for skipping a migration -- the non-contiguity of migrations is not necessarily a problem, and i'm not sure ranges convey that well.

And a non-blocking thought: It's a shame that we can't write the "comment" as an actual YAML comment. It doesn't seem feasible though (no way to add comments using Jackson; prepending the file contents with the header is clumsy).

rzpt:
Thanks! I decided to store ranges because if a product has many migrations, then the file would be much larger. Skipping a migration usually entails making the migration do nothing, which does not affect what schema versions are in the manifest, but it is possible other cases exist.

ptaug:
Looks good to me -- I'll ask for an approval from one of the maintainer.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/CreateManifestTask.java
```diff
@@ -114,11 +131,18 @@ final void createManifest() throws IOException {
 
         List<ProductDependency> productDependencies = productDependencyManifest.productDependencies();
         if (productDependencies.isEmpty()) {
-            requireAbsentLockfile();
+            requireAbsentLockfile(getLockfile());
         } else {
             ensureLockfileIsUpToDate(productDependencies);
```
fawind: small: Can we rename this method to something like `ensurePdepsLockfileIsUpToDate` to make it clear that we now manage two different lock files?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/CreateManifestTask.java
```diff
@@ -217,6 +252,16 @@ private Optional<String> diff(File existing, String upToDateContents) {
...
     }
 
+    private File getSchemaLockfile() {
+        return getProject().file(SchemaVersionLockFile.LOCK_FILE);
+    }
+
+    private void ensureSchemaLockfileIsUpToDate(List<SchemaMigration> schemaMigrations) throws IOException {
+        File lockfile = getSchemaLockfile();
+        String upToDateContents = ObjectMappers.writeSchemaVersionsAsString(SchemaVersionLockFile.of(schemaMigrations));
+        ensureFileIsUpToDate(lockfile, upToDateContents);
```
fawind: It feels a bit weird to me that `./gradlew writeProductDependenciesLocks` updates the schema lock file, given it has nothing to do with product dependencies. Could we split this up and maybe add something like `./gradlew writeSchemaMigrationLocks`?

Maybe we can parameterize the `CreateManifestTask#shouldWriteLocks` method to take take in the task name to check as an additional parameter and then pipe this information through from the respective `ensureXXXLockfileIsUpToDate` method?

readme.md
```diff
@@ -133,6 +133,24 @@ dependencies {
 }
 ```
 
+### Schema versions
+
+sls-packaging also maintains a lockfile, `schema-versions.lock`, which should be checked in to Git.
+This file is an accurate reflection of the schema versions specified in the manifest.
+The file can be used to easily determine what schema versions are supported by a particular version of the code.
+Run **`./gradlew --write-locks`** or **`./gradlew writeProductDependenciesLocks`** to update it.
```
fawind: Same as the previous comment

fawind (commented):
Looks good! Just a last small comment regarding the task name for writing the schema lock file.

rzpt:
@fawind Thanks for the review! I've addressed your comments in the new commit.

fawind (approved):
Looks good, thanks for iterating!

PR #1467, opened 2023-04-25
throw SafeRuntimeException instead of GradleException

rzpt:
## Before this PR
I am trying to use ObjectMappers from another project that cannot have gradle-api on the classpath. This does not work because GradleException is not on the classpath.

## After this PR

==COMMIT_MSG==
Throw SafeRuntimeException instead of GradleException in ObjectMappers
==COMMIT_MSG==

## Possible downsides?

Anyone catching a GradleException could have problems.
I did not modify all the other places that throw GradleException, just the ObjectMappers usages.

### DIFF
--- a/changelog/@unreleased/pr-1467.v2.yml
+++ b/changelog/@unreleased/pr-1467.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Throw SafeRuntimeException instead of GradleException in ObjectMappers
+  links:
+  - https://github.com/palantir/sls-packaging/pull/1467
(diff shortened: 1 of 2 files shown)


ptaug:
Could you do the object mapping in your other project instead? This gives you control over how you want to handle the JsonProcessingException/IOException directly, and we don't have to maintain and rely on object mappers that live in outside projects.

ptaug:
Ok, after talking internally:
- We should split the `SchemaVersionLockFile` in a separate `gradle-sls-packaging-api` project.
- Annotate it so that we can minimize ObjectMapper configuration, e.g. with `@JsonProperty` on properties to avoid `            .setPropertyNamingStrategy(PropertyNamingStrategies.KEBAB_CASE)`.
- Have an `ObjectMapper` in the consumer of the file.

Alternative would be to expose the `ObjectMapper` in `gradle-sls-packaging-api`, but we should create that separate project/jar.

ptaug:
Ok, after talking internally:
- We should split the `SchemaVersionLockFile` in a separate `gradle-sls-packaging-api` project.
- Annotate it so that we can minimize ObjectMapper configuration, e.g. with `@JsonProperty` on properties to avoid `            .setPropertyNamingStrategy(PropertyNamingStrategies.KEBAB_CASE)`.
- Have an `ObjectMapper` in the consumer of the file.

Alternative would be to expose the `ObjectMapper` in `gradle-sls-packaging-api`, but we should create that separate project/jar.

PR #1486, opened 2023-06-14
Remove UseAVX=2 overrides from JDK-8292158

carterkozak:
We've rolled out new JDKs patched for this issue. TLS implementations on newer hardware benefit from AVX-512 instructions.

==COMMIT_MSG==
Remove UseAVX=2 overrides from JDK-8292158
==COMMIT_MSG==

I'm going to plan to roll this out EoW once I've verified old jdks aren't being used any longer.


schlosna (approved):

schlosna:
Separate from this, but possibly as part of same release rollout -- do we want to remove the disabling of C2 for `sun/security/ssl/SSLEngineInputRecord.decodeInputRecord` on JDK 11?

https://github.com/palantir/sls-packaging/blob/4fc9cde1ee17ba48d11281988691c0f503a94676/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfig.java#L64-L66

carterkozak:
> do we want to remove the disabling of C2 for sun/security/ssl/SSLEngineInputRecord.decodeInputRecord on JDK 11?

Good call, thanks!

Current plan is not to merge this PR until we've begun to include JDKs in our distribution tarballs to avoid risk of reintroducing issues in certain environments.

carterkozak:
Closing in favor of #1517

PR #1505, opened 2023-08-01
Increase G1 default pause target from 200ms to 500ms

carterkozak:
We use 500ms by default, up from the JDK default value of 200ms. Using G1, eden space is dynamically chosen based on the amount of memory which can be collected within the pause time target. Higher pause target values allow for more eden space, resulting in more stable old generation in high-garbage or low-gc-thread scenarios. In the happy case, increasing the pause target increases both throughput and latency. In degenerate cases, a low target can cause the garbage collector to thrash and reduce throughput while increasing latency.

==COMMIT_MSG==
Increase G1 default pause target from 200ms to 500ms
==COMMIT_MSG==

The G1 pause target can be modified thusly:
```groovy
gc 'hybrid', {
    maxGCPauseMillis 250
}
```

## Possible downsides?
I considered making two PRs, one to make `MaxGCPauseMillis` configurable, and a second to increase the default value.
Increasing this value may increase latency in latency-sensitive apps. Those which are particularly latency-sensitive are already using the `response-time` profile, so there shouldn't be a great deal of risk.

### DIFF
--- a/changelog/@unreleased/pr-1505.v2.yml
+++ b/changelog/@unreleased/pr-1505.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Increase G1 default pause target from 200ms to 500ms
+  links:
+  - https://github.com/palantir/sls-packaging/pull/1505
--- a/changelog/@unreleased/pr-1505.v2.yml
+++ b/changelog/@unreleased/pr-1505.v2.yml
@@ -1,5 +1,13 @@
 type: improvement
 improvement:
-  description: Increase G1 default pause target from 200ms to 500ms
+  description: |-
+    Increase G1 default pause target from 200ms to 500ms
+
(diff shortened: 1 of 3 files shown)


iamdanfox:
The justification that particularly latency-sensitive applications should already using the response-time profile is pretty convincing to me.

From trying this on AS (which has a 12 GiB heap), it looks like the duration of young gen collections is now a lot more consistent. I've also tried it on another service (catalog-publisher) but haven't had similar success yet.

schlosna (approved):
500ms is a reasonable max pause bounds for the types of services running `hybrid` / default profile and should help avoid some of the excessive minor GC churn.

schlosna:
@parker89 for SA as this is relevant for some heap size tuning he's been working.

raiju:
How did we get to the 500ms number? (Note that this is the default strategy)

PR #1510, opened 2023-08-21
Include JDKs in Dists

CRogers:
## Before this PR
We want to include JDKs in dists (reasons in [this doc](https://pl.ntr/2hO)). There is currently no way to do this.

I had tried to do this all from an internal plugin, but it's impossible to set `javaHome` correctly and maintain laziness.

## After this PR
We add a way to include a number of JDKs to a dist, up to one per version, and then set `javaHome` in go-java-launcher to the correct one.

Multiple JDKs are required as some internal services run user code, which select from multiple JDK versions, and these should continue to work. I'm assuming we won't need multiple _variants_ of the same JDK version, as that would be hugely wasteful to deploy and we should avoid this as much as we can.

==COMMIT_MSG==
Add `jdks` property to `distribution` extension to help include JDKs into dists.
==COMMIT_MSG==


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java
```diff
@@ -85,7 +97,13 @@ public JavaServiceDistributionExtension(Project project) {
         excludeFromVar = objectFactory.listProperty(String.class);
         excludeFromVar.addAll("log", "run");
 
-        env = objectFactory.mapProperty(String.class, String.class).empty();
+        env = objectFactory.mapProperty(String.class, String.class);
```
felixdesouza, line 99: I think you want to separate out env that's editable by everyone else and this env. I think you're able to actually empty this `env` and everything is fine.

I think you'd want to have a resulting env, that first applies the jdks env and then the user env?

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java
```diff
@@ -85,7 +97,13 @@ public JavaServiceDistributionExtension(Project project) {
...
+        env.putAll(project.provider(() -> {
+            return jdks.get().entrySet().stream()
```
felixdesouza: can't you map `jdks`?

CRogers: Already have - I thought you'd notice that heh

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java
CRogers: True. No one uses `env` internally at the moment, but I guess it should still not be editable.

CRogers: fixed + test

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java
```diff
@@ -331,7 +331,7 @@ private static Provider<Map<String, String>> userConfiguredEnvWithJdkEnvVars(
         return distributionExtension.getEnv().zip(distributionExtension.getJdks(), (userConfiguredEnv, jdks) -> {
             Map<String, String> actualEnv = new HashMap<>(userConfiguredEnv);
```
felixdesouza: wouldn't you need a `LinkedHashMap` in addition to the below

CRogers: I just thought that... very tricksy

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPluginTests.groovy
```diff
@@ -179,7 +179,6 @@ class JavaServiceDistributionPluginTests extends GradleIntegrationSpec {
             }
 
             repositories {
-                jcenter()
```
CRogers, line 182: Deprecation warning with new higher version of gradle

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/SlsBaseDistPlugin.java
```diff
@@ -31,7 +31,7 @@ public class SlsBaseDistPlugin implements Plugin<Project> {
 
     public static final String SLS_DIST_USAGE = "sls-dist";
 
-    public static final GradleVersion MINIMUM_GRADLE = GradleVersion.version("6.1");
+    public static final GradleVersion MINIMUM_GRADLE = GradleVersion.version("7.6");
```
CRogers, line 34: `Provider#zip` is introduced in 6.6. However, it seems that gradle introduced plugin bytecode rewriting sometime between 6.1 and 6.6. This means 6.6 now falls over on the jackson 2.15.2 dep from baseline, so we need to go all the way to 7.6

felixdesouza: don't we need 7.6.2?

CRogers: 7.6 for Java 19: https://docs.gradle.org/current/userguide/compatibility.html

felixdesouza: https://docs.gradle.org/7.6.2/release-notes.html and https://github.com/gradle/gradle/issues/24390

felixdesouza: happy to merge if 7.6 doesn't break and have a lower min version

CRogers: Huh, I wonder why it works...

felixdesouza:
👍

PR #1516, opened 2023-09-01
Remove version from JDK path

CRogers:
## Before this PR
In #1510, we included the version in the path JDKs live in the dist to help people getting alerts from third party security scanners. However, adding a version here means that when container images are built, their layer caches are busted and docker images are recreated all the time.

## After this PR
==COMMIT_MSG==
Remove version from JDK path when JDKs are included in dists.
==COMMIT_MSG==

## Possible downsides?
Might be marginally harder to work out which exact service instance has the vuln (as in might need an extra step) - but the people who this would affect have agreed this is ok.

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java
@@ -273,11 +273,10 @@ public class JavaServiceDistributionExtension extends BaseDistributionExtension 
         // and you end up with a report like so:
         //      Path: /opt/palantir/services/.24710105/service/jdk17
         // rather than more useful:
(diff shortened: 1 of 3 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java
```diff
@@ -265,19 +265,18 @@ private static GcProfile getDefaultGcProfile(JavaVersion javaVersion) {
         return new GcProfile.Throughput();
     }
 
-    final String jdkPathInDist(JavaVersion javaVersionValue) {
+    public final String jdkPathInDist(JavaVersion javaVersionValue) {
```
CRogers, line 268: I need to use this in an internal gradle plugin.

fawind (approved):

PR #1517, opened 2023-09-05
Allow AVX-512 instructions when jdks are bundled with distributions

carterkozak:
## Before this PR
No avx512 anywhere.

## After this PR

==COMMIT_MSG==
Allow AVX-512 instructions when jdks are bundled with distributions
==COMMIT_MSG==
This builds upon #1510, relying on usage of the feature flag to drop arguments for compatibility with older jdk builds.

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfig.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfig.java
@@ -78,16 +78,17 @@ public final class LaunchConfig {
             // AWS-managed systems that modify DNS records on failover.
             "-Dsun.net.inetaddr.ttl=20",
             "-XX:+UnlockDiagnosticVMOptions",
-            // Disable AVX-512 intrinsics due to AES/CTR corruption bug in https://bugs.openjdk.org/browse/JDK-8292158
             "-XX:+IgnoreUnrecognizedVMOptions",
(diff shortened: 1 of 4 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfig.java
```diff
@@ -208,12 +214,25 @@ static void action(Params params) {
...
                         .addAllJvmOpts(javaAgentArgs(params))
                         .addAllJvmOpts(alwaysOnJvmOptions)
+                        .addAllJvmOpts(avxOptions)
                         .addAllJvmOpts(params.getDefaultJvmOpts().get())
                         .env(defaultEnvironment)
                         .build(),
                 params.getCheckLauncher().get().getAsFile());
     }
 
+    // When a specific jdk is provided, we can assume a modern versions including the
```
mpritham, line 224: I suppose it's possible for us to have a modern jdk bundled, but have an affected java version?

mpritham (approved):

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfig.java
carterkozak: we recall products which don't stay up to date, so they're guaranteed to have reasonable jdks bundled :-]

PR #1519, opened 2023-09-12
Unset the JDK-11 CompileCommand override for bundled JDKS

carterkozak:
The config option is only necessary for outdated JDKs with a known bug, thus does not impact products which are shipped with bundled JDKs.

==COMMIT_MSG==
Unset the JDK-11 CompileCommand override for bundled JDKS
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfig.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfig.java
@@ -168,9 +168,10 @@ public final class LaunchConfig {
                         .addAllJvmOpts(avxOptions)
                         .addAllJvmOpts(params.getAddJava8GcLogging().get() ? java8gcLoggingOptions : ImmutableList.of())
                         // Java 11.0.16 introduced a potential memory leak issues when using the C2
-                        // compiler
+                        // compiler, resolved in 11.0.16.1
                         .addAllJvmOpts(
                                 javaVersion.compareTo(JavaVersion.toVersion("11")) == 0
+                                                && !params.getBundledJdks().get()
                                         ? jdk11DisableC2Compile
                                         : ImmutableList.of())
                         .addAllJvmOpts(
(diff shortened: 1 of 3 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfig.java
```diff
@@ -168,9 +168,10 @@ static void action(Params params) {
                         .addAllJvmOpts(avxOptions)
                         .addAllJvmOpts(params.getAddJava8GcLogging().get() ? java8gcLoggingOptions : ImmutableList.of())
                         // Java 11.0.16 introduced a potential memory leak issues when using the C2
-                        // compiler
+                        // compiler, resolved in 11.0.16.1
```
schlosna, line 171: [JDK-8291665](https://bugs.openjdk.org/browse/JDK-8291665) for future reference if needed

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfig.java
```diff
@@ -168,9 +168,10 @@ static void action(Params params) {
...
                         .addAllJvmOpts(
                                 javaVersion.compareTo(JavaVersion.toVersion("11")) == 0
+                                                && !params.getBundledJdks().get()
                                         ? jdk11DisableC2Compile
                                         : ImmutableList.of())
```
schlosna, line 176: I'm fine with PR as it is now, but wondering if we should completely remove the compile command even on JDK 11 as anything picking up new sls-packaging release would have dependency on fixed JDKs at this point.

```suggestion
```

schlosna (approved):

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfig.java
carterkozak: The failure mode is a potentially old depot, however it is unlikely

PR #1520, opened 2023-09-12
JDK-21+ response-time gc profile uses generational zgc

carterkozak:
==COMMIT_MSG==
JDK-21+ response-time gc profile uses generational zgc
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/gc/GcProfile.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/gc/GcProfile.java
@@ -49,6 +49,16 @@ public interface GcProfile extends Serializable {
 
         @Override
         public final List<String> gcJvmOpts(JavaVersion javaVersion) {
+            // JDK-21+ uses generational ZGC as the response-time optimized garbage collector.
+            if (javaVersion.compareTo(JavaVersion.toVersion("21")) >= 0) {
+                return ImmutableList.of(
+                        "-XX:+UseZGC",
+                        // https://openjdk.org/jeps/439
+                        "-XX:+ZGenerational",
+                        // "forces concurrent cycle instead of Full GC on System.gc()"
+                        "-XX:+ExplicitGCInvokesConcurrent");
+            }
+
             // The CMS garbage collector was removed in Java 14: https://openjdk.java.net/jeps/363. Users are free to
             // use it up until this release.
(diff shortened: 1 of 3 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/gc/GcProfile.java
```diff
@@ -49,6 +49,16 @@ class ResponseTime implements GcProfile {
...
         @Override
         public final List<String> gcJvmOpts(JavaVersion javaVersion) {
+            // JDK-21+ uses generational ZGC as the response-time optimized garbage collector.
+            if (javaVersion.compareTo(JavaVersion.toVersion("21")) >= 0) {
+                return ImmutableList.of(
+                        "-XX:+UseZGC",
+                        // https://openjdk.org/jeps/439
+                        "-XX:+ZGenerational",
+                        // "forces concurrent cycle instead of Full GC on System.gc()"
+                        "-XX:+ExplicitGCInvokesConcurrent");
```
schlosna, line 59: notes for myself:

- `NUMA` is enabled by default:
https://github.com/openjdk/jdk/blob/a9679acd5c85b88afe827dc55dbe9a78b00c8918/src/hotspot/share/gc/z/zArguments.cpp#L117-L119
- [`-XX:+UseDynamicNumberOfGCThreads` is enabled by default](https://malloc.se/blog/zgc-jdk17)
- `-XX:+ClassUnloadingWithConcurrentMark` is not yet supported for ZGC from what I can tell.
- [`-XX:+UseStringDeduplication` is available for ZGC in JDK 18+](https://malloc.se/blog/zgc-jdk18), could enable separately https://github.com/palantir/sls-packaging/issues/1378

schlosna (approved):
This seems reasonable to me to get a baseline

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/gc/GcProfile.java
carterkozak: based on https://www.chriswhocodes.com/corretto_jdk11_options.html `-XX:+ClassUnloadingWithConcurrentMark` defaults to `true` (and has since jdk-11, I haven't verified earlier releases).

The zgc documentation suggested that numa support is enabled by default, but turned off if the jvm detects it's not needed, but that `UseNUMA` flag overrides this behavior: https://wiki.openjdk.org/display/zgc/Main#Main-EnablingNUMASupport

PR #1629, opened 2024-04-12
Allow plugins to configure where the configuration.yml is included from

CRogers:
## Before this PR
Internally, we have both:
1. Plugins that create tasks which template the `configuration.yml`. This means there is no `deployment/configuration.yml` and a task needs to run in order to create it. _Most of the time_ the plugin author remembers to make `distTar` in this repo depend on that templating task, but not always.
2. Plugins that have tasks which read the `configuation.yml` as a raw file input.

Unfortunately, when combined together there is no automatic task dep between the templating task and the configuration.yml consuming task, since they both just use the static file location between them.

Gradle 8 (often) verifies your task dependencies and will error out if they are not set up properly. This problem gets exposed to repos using Gradke 8.

People have unfortunately resorted to listing all the `dependsOn` between each of the task sets manually, which is error prone and full of clutter. 

At the moment, there is no way for the two different internal plugins (or even the templating plugin to these sls-plugins) to communicate the tasks needed to generate the `configuration.yml`. When a `configuration.yml` is generated, it must live in `deployment/configuration.yml`, and so must also be `.gitignore`d.

## After this PR
==COMMIT_MSG==
Allow plugins to configure/read where the configuration.yml is included from
==COMMIT_MSG==

By placing a configuration property in the appropriate distribution/asset extension, we allow the templating task to set the property to be it's output, and tasks that read the configuration to read from it the extension. Since it's a property, the task dep is propagated automatically provided plugins are written sensibly. Templating plugins can now output the configuration.yml to a build directory instead `deployment/configuration.yml` where it needs to be gitignored.

## Possible downsides?
Always the possibility of unintentional breaks, but he default value is `deployment/configuration.yml` so it all _should_ continue to work.


gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPluginTests.groovy
```diff
@@ -366,7 +394,7 @@ class JavaServiceDistributionPluginTests extends GradleIntegrationSpec {
         runTasks(':build', ':distTar', ':untar')
 
         then:
-        String startScript = file('dist/service-name-0.0.1/service/bin/service-name', projectDir).text
+        String startScript = new File(projectDir, 'dist/service-name-0.0.1/service/bin/service-name').text
```
CRogers, line 397: `file(` makes the file if it doesn't exist - makes the test very confusing, especially when using `file(...).exists()` as an assertion. Going to (eventually) write an automated check for this somehow.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPluginTests.groovy
```diff
@@ -354,10 +354,38 @@ class JavaServiceDistributionPluginTests extends GradleIntegrationSpec {
...
+            task createConfigurationYml {
+                outputs.file('build/some-place/configuration-but-called-something-else.yml')
+                
+                doFirst {
+                    file('build/some-place/configuration-but-called-something-else.yml').text = 'custom: yml'
+                }
+            }
+
+            distribution {
+                configurationYml.fileProvider(tasks.named('createConfigurationYml').map { it.outputs.files.singleFile })
```
CRogers, line 377: In java plugin code you're just going be doing `taskProvider.flatMap(CustomTask::getOutputFile)`, so not this ugly.

pkoenig10 (approved):

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPluginTests.groovy
pkoenig10: Alternatively we could stop using `nebula-test` and just write our tests in Java. We've decided to do this in one of our internal plugins and I find to test development UX to be meaningfully better.

I don't think `nebula-test` provides much value using BuildKit directly and I always find interacting with Groovy to be fairly painful.

CRogers: I actually quite like groovy/nebula tests; maybe I have stockholm syndrome from too many years of using it. I am open to making our own Java framework at some point, provided we can get all the auto-debug, performance, good text blocks, etc working. Even if we did that we're going to have a 9+ year corpus of nebula tests that aren't going away for a long time we need to maintain, though.

PR #1633, opened 2024-04-23
Allow plugins to configure where the configuration.yml is included from, take 2

CRogers:
## Before this PR
See original issue at https://github.com/palantir/sls-packaging/pull/1629

When we rolled this out internally, it interacted poorly with some internal code that does file deduplication in container image layers. This deduplication code uses the copyspec from this repo in order to avoid actually running `distTar` in order to build a container image, so it meant that adding the `.rename` messed with what it was doing. These interactions are extremely unfortunate, but there's no clear way to remove the copyspec copying without destroying build iteration performance.

## After this PR
==COMMIT_MSG==
Allow plugins to configure/read where the configuration.yml is included from
==COMMIT_MSG==

We just carefully avoid doing a `rename` this time and add a mega comment about it.

### DIFF
--- a/changelog/@unreleased/pr-1633.v2.yml
+++ b/changelog/@unreleased/pr-1633.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Allow plugins to configure/read where the configuration.yml is included
+    from
+  links:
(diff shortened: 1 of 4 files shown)


felixdesouza:
👍

felixdesouza:
👍

PR #1643, opened 2024-05-09
sls-packaging sets the individual-addressability extension

carterkozak:
We enforce this everywhere in code, however we need to make sure stragglers are picked up to avoid infrastructure failures. We could assume the manifest default is true, however that is risky for older product. This way new releases pick up the new value if it's not already specified.

==COMMIT_MSG==
sls-packaging sets the individual-addressability extension default value to `true` when not otherwise specified
==COMMIT_MSG==

## Possible downsides?
Ideally we excavate this into buildscripts first, which we've done in 99.9% of projects at this point. This acts as a backstop for one-off projects, assuming they take sls-packaging upgrades. This will apply to other languages as well as Java.

### DIFF
--- a/changelog/@unreleased/pr-1643.v2.yml
+++ b/changelog/@unreleased/pr-1643.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: sls-packaging sets the individual-addressability extension default
+    value to `true` when not otherwise specified
+  links:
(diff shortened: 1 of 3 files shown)


carterkozak:
(I haven't run tests yet, I suspect a test will fail until I update expectations)

CRogers (approved):

PR #1645, opened 2024-05-16
Add a helm-chart.v1 product type

sfackler:
## Before this PR
The createManifest task could only be used to create services, daemons, or assets.

## After this PR

==COMMIT_MSG==
Added the `HELM_CHART_V1` `ProductType`.
==COMMIT_MSG==

This can be used by downstream Gradle logic bundling Helm charts.


felixdesouza:
👍

PR #1646, opened 2024-05-16
`ConfigTarTask` supports custom `configuration.yml` location

CRogers:
## Before this PR
In https://github.com/palantir/sls-packaging/pull/1629, we added support for specifying a different location/task for configuraion.ymls to be generated.

Unfortunately, as part of the PR I forgot about the `configTar` task.

## After this PR
==COMMIT_MSG==
`ConfigTarTask` supports custom `configuration.yml` location
==COMMIT_MSG==

## Possible downsides?
The number of places where Gradle plugins yolo-create their own config tar task is kinda horrifying. This certainly hasn't been built "well". I don't think these should break, but you really never know.

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/DeploymentDirInclusion.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/DeploymentDirInclusion.java
@@ -1,0 +1,63 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/DeploymentDirInclusion.java
```diff
@@ -0,0 +1,63 @@
...
+
+package com.palantir.gradle.dist;
+
+import org.gradle.api.Action;
+import org.gradle.api.file.CopySpec;
+import org.gradle.api.file.ProjectLayout;
+
+public final class DeploymentDirInclusion {
+
+    public static void includeFromDeploymentDirs(
```
CRogers, line 25: This was all more or less moved exactly from `DistTarTask`.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/tasks/ConfigTarTask.java
```diff
@@ -30,19 +29,12 @@
...
-public class ConfigTarTask extends Tar {
+public abstract class ConfigTarTask extends Tar {
     @Override
     public final AbstractCopyTask from(Object... sourcePaths) {
         return this.from(sourcePaths, _ignored -> {});
     }
 
-    @Override
-    @SuppressWarnings({"RawTypes", "deprecation"})
-    public final AbstractCopyTask from(Object sourcePath, Closure closure) {
```
CRogers, line 41: This is auto-implemented by Gradle by making the task class abstract.

felixdesouza (approved):

PR #1662, opened 2024-07-01
Allow `ArtifactLocator` s to have inputs derived from task outputs.

felixdesouza:
## Before this PR
~When you use `whenObjectAdded` it will eagerly realise any objects that exist in the provider:~

> ```
> Adds an {@code Action} to be executed when an object is added to this collection.
> Like {@link #all(Action)}, this method will cause all objects in this container to be realized.
> ```

~So if a provider has a task output, it will complain because it's trying to realise it at configuration time, which is not possible.~

It turns out that the validation in `ArtifactLocator` would always call `get`, so it's never enough to do a `configureEach` or a `whenObjectAdded`, as we would eagerly realise it the moment a concrete version was added to the set.

An alternative is to wrap the `ArtifactLocator` into a provider, however, I believe that will lose task dependency information!

## After this PR

==COMMIT_MSG==
Allow `ArtifactLocator`s to have inputs derived from task outputs.
==COMMIT_MSG==

Use `@Nested` in `SetProperty` task input for `CreateManifest` task. The validation happens when converting to `JsonArtifactLocator`. Add integration tests that allow using derived task output providers.

### DIFF
--- a/changelog/@unreleased/pr-1662.v2.yml
+++ b/changelog/@unreleased/pr-1662.v2.yml
@@ -1,6 +1,5 @@
 type: fix
 fix:
-  description: Use `configureEach` instead of `whenObjectAdded` for the `artifacts`
-    block.
+  description: Allow `ArtifactLocator`s to have inputs derived from task outputs.
   links:
   - https://github.com/palantir/sls-packaging/pull/1662
--- a/changelog/@unreleased/pr-1662.v2.yml
+++ b/changelog/@unreleased/pr-1662.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Use `configureEach` instead of `whenObjectAdded` for the `artifacts`
+    block.
+  links:
+  - https://github.com/palantir/sls-packaging/pull/1662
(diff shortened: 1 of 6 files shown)


CRogers (dismissed):

sfackler:
Is this going to be sufficiently lazy? The case I ran into is one where an artifact property is derived from another task, so the property value can't actually be evaluated until the property's task dependencies have run rather than when the set of artifacts is realized.

felixdesouza:
> Is this going to be sufficiently lazy? The case I ran into is one where an artifact property is derived from another task, so the property value can't actually be evaluated until the property's task dependencies have run rather than when the set of artifacts is realized.

I think this should be fine? this is only ever passed into the output of another task, so it'd only be evaluated when trying to evaluate those inputs. It's a bit of a shame (having tighter validation loop).

I'll write an extra integration test to confirm, but feel free to clone and check this against your use case as well, then we can iterate from there.

sfackler (approved):

bmarcaur (approved):

PR #1668, opened 2024-07-09
assets can also have custom configuration.yml locations and fall back to `build/deployment/configuration.yml`

CRogers:
## Before this PR
In #1666, we tried to allow assets to have a configurable location for their `configuration.yml`. However, the new codepath did not support having a templating task output to `build/deployment/configuration.yml`, which some internal products did for assets. Some of these products are very "basic" assets containing nothing apart from templated configuration and lack any real tests, meaning they managed to slip through and get published without any `configuration.yml`, go straight to prod and caused PDS-560061 and PDS-560082. Detailed RCA [here](https://pl.ntr/2nx). As part of the remediation, this change was reverted in #1667.

## After this PR
==COMMIT_MSG==
assets can also have custom configuration.yml locations and the asset plugin will fall back to `build/deployment/configuration.yml` if the `deployment/configuration.yml` does not exist.
==COMMIT_MSG==

## Possible downsides?
I haven't yet worked out how to make the task fail if there is no `configuration.yml` at all

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/DeploymentDirInclusion.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/DeploymentDirInclusion.java
@@ -63,8 +63,6 @@ public final class DeploymentDirInclusion {
             // `build/deployment/configuration.yml`.
             t.from(projectLayout.getBuildDirectory().file("deployment/configuration.yml"));
         });
-
-        root.int
     }
 
     private DeploymentDirInclusion() {}
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/DeploymentDirInclusion.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/DeploymentDirInclusion.java
@@ -19,6 +19,7 @@ package com.palantir.gradle.dist;
 import org.gradle.api.Action;
 import org.gradle.api.file.CopySpec;
 import org.gradle.api.file.ProjectLayout;
+import org.gradle.api.file.RegularFile;
 
 public final class DeploymentDirInclusion {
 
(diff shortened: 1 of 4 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/DeploymentDirInclusion.java
```diff
@@ -56,6 +56,12 @@ public static void includeFromDeploymentDirs(
                 throw new IllegalStateException("The file set to be the value of getConfigurationYml() "
                         + "must be called configuration.yml. Instead, it was called " + file.getAsFile());
             }));
+
+            // This is a bit of hack to "fall back" to a `build/deployment/configuration.yml` if the
+            // `deployment/configuration.yml` does not exist. This can happen when there's a templating setup
+            // that pre-exists the being able to configuration.yml on the extension and it outputs to
+            // `build/deployment/configuration.yml`.
+            t.from(projectLayout.getBuildDirectory().file("deployment/configuration.yml"));
         });
```
CRogers, line 65: It would be really nice if there was a way to evaluate the files copied as part of this copyspec and check that we do actually have a `configuration.yml` in there. I don't think this is possible at the copyspec level (even using `eachFile` - there's no time for a check at the "end" to happen as we never get told when the "end" is).

I think it would need to be a check after the distTar task runs, which even then is dangerous as that check doesn't run before (and so prevent) the publish after it.

CRogers: I guess we could add a task action to all the tar tasks that checks the tar files, but this seems like it might be expensive to read all the way through the tar.gz and I'm concerned Gradle might detect that and turn off parallelisation for those tasks.

felixdesouza: > It would be really nice if there was a way to evaluate the files copied as part of this copyspec and check that we do actually have a `configuration.yml` in there. I don't think this is possible at the copyspec level (even using `eachFile` - there's no time for a check at the "end" to happen as we never get told when the "end" is).
> 
> I think it would need to be a check after the distTar task runs, which even then is dangerous as that check doesn't run before (and so prevent) the publish after it.

I was initially thinking of doing a finaliser task, but it's kinda weird because you might not want to run all the tasks that it might depend on.

You could potentially do something like:
* you have a checking task, it is configured as `mustRunAfter` for all the `distTar` tasks.
* you *always* run the checking task
* it's input should be the distTar output with `getLocationOnly` if possible or we can just annotate the `ConfigurableFileCollection` or whatever type with `@Internal` 
* Can potentially make it incremental

I believe these should allay your concerns? It doesn't feel great though, just a lot of weird hand holding. Might be worth just trying with the `doLast` task action and seeing whether we can get things to work in parallel that way.

felixdesouza: okay, incremental might not work due to needing to declare it as an input.

Maybe instead of one checking finaliser task, you could just have a finaliser task per `distTar` perhaps, idk

felixdesouza: you could also potentially have the `eachFile` write into some structure that's an internal property on the finaliser-type task

CRogers: I think finalizers don't work as the publish tasks after will still run if the finalizers fail (under the presence of `--continue`?)

CRogers: I'm going to try and see if adding a `doLast` breaks parallelism. I think inside that doLast we could check some value that's been set on a property when the main action runs...

felixdesouza: that's the behaviour regardless of `--continue`, I think you have to pass in the task state of the `distTar` and check that in the `onlyIf`

felixdesouza (approved):

PR #1673, opened 2024-07-31
Override the JNA temporary directory to point to var/data/tmp

SerialVelocity:
## Before this PR

The JNA temporary directory was left to it's default which for most OS's is `$USER/.cache`.

There are two issues with this:
- This can be mounted as `noexec` which means JNA will fail (seen in prod)
- If we fill up this folder, the user may not be able to SSH in as the disk is full (very unlikely)

## After this PR

==COMMIT_MSG==
Override the JNA temporary directory to point to var/data/tmp
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1673.v2.yml
+++ b/changelog/@unreleased/pr-1673.v2.yml
@@ -1,0 +1,5 @@
+type: feature
+feature:
+  description: Override the JNA temporary directory to point to var/data/tmp
+  links:
+  - https://github.com/palantir/sls-packaging/pull/1673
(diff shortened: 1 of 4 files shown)


CRogers:
The implementation looks fine, I thinks someone from the Runtime team should decide whether we run this in prod though.

schlosna (approved):
I'm supportive of this change.

An example of where this is needed is for Cassandra and Elasticsearch https://access.redhat.com/solutions/6991416 & https://davecturner.github.io/2021/08/30/seven-year-old-segfault.html

PR #1693, opened 2024-09-12
Stop mergeDiagnosticJson having @CacheableTransform; speed up build

CRogers:
## Before this PR
There's a task `mergeDiagnosticJson`, which gathers up all the different possible diagnostics listed in json files from libraries and combines them together into the SLS manifest.

It does this by using an [artifact transform](https://docs.gradle.org/8.10.1/userguide/artifact_transforms.html), which has been listed as build cacheable using `@CacheableTransform`. We use remote build caches in builds, so this will, for each jar in the runtime classpath, attempt to load the result of extracting a single file from this jar from the remote build cache.

The runtime of actually executing extracting a single small file at a known path from a zip is _milliseconds_. Doing a network request to the build cache is likely 100s of milliseconds. Since there are many jars, we're making many build cache requests which all take significantly longer than just doing the operation locally. We don't even get to avoid downloading the input jars by using the build cache - they still need to exist to have their hash calculated for cache keys.

The result is that on internal projects this is a common sight:

mergeDiagnosticJson

and we can check a build scan to see how bad this actually is: 372 cacheable transforms executed, taking a whopping 53 secs!

Screenshot 2024-09-12 at 17 25 36

## After this PR
==COMMIT_MSG==
Do not use the remote build cache for `mergeDiagnosticJson`; it takes far longer than just doing the operation locally. Avoids spamming output with "Requesting from remote build cache" too.
==COMMIT_MSG==

This speeds it up hugely: 372 non-cacheable transforms executed, taking under 3 seconds!

Screenshot 2024-09-12 at 17 25 43

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/ExtractFileFromJar.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/ExtractFileFromJar.java
@@ -29,12 +29,12 @@ import org.gradle.api.file.FileSystemLocation;
 import org.gradle.api.provider.Provider;
 import org.gradle.api.tasks.PathSensitive;
 import org.gradle.api.tasks.PathSensitivity;
+import org.gradle.work.DisableCachingByDefault;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-// Do *not* make this class cacheable using @CacheableTransform. It's far, far faster to just extract a single
-// file from a zip than to make network requrest to the remote build cache.
-// See https://github.com/palantir/sls-packaging/pull/1693
+@DisableCachingByDefault(
+        because = "Extracting a single file from a zip is much faster than making network requests to the build cache")
 public abstract class ExtractFileFromJar implements TransformAction<FileExtractParameter> {
     private static final Logger log = LoggerFactory.getLogger(ExtractFileFromJar.class);
 
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/ExtractFileFromJar.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/ExtractFileFromJar.java
@@ -32,6 +32,9 @@ import org.gradle.api.tasks.PathSensitivity;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+// Do *not* make this class cacheable using @CacheableTransform. It's far, far faster to just extract a single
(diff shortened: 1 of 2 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/ExtractFileFromJar.java
```diff
@@ -33,7 +32,6 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-@CacheableTransform
 public abstract class ExtractFileFromJar implements TransformAction<FileExtractParameter> {
```
carterkozak, line 38: Might be worth commenting so we don't try to get clever and add it back. Great PR description, fantastic find!

CRogers: Good idea, have done so.

felixdesouza: you should use `org.gradle.work.DisableCachingByDefault`. It also allows you to include a reason as well. Should also prevent a regression if by any chance they turn on caching automatically.

felixdesouza:
The savings!!! Nice find!

CRogers:
I think some of these transforms managed to run in parallel with other tasks; but the overall build time still improved from 2mins 7secs -> 1min 45secs, so about 20 secs overall build time improvement, which is still great.

felixdesouza:
> I think some of these transforms managed to run in parallel with other tasks; but the overall build time still improved from 2mins 7secs -> 1min 45secs, so about 20 secs overall build time improvement, which is still great.

That is strange though, I would have expected artifact transforms (cached or not) to be treated as a separate entity allowing it to run in parallel. One day I will understand Gradle's execution model 😢

CRogers:
I suspect there's a limited number of execution threads and the transform is consuming a huge number of them because there are so many files. It certainly seems like that with the big block of `Requesting from the remote build cache` you see.

CRogers:
Any other comments or is this good to merge?

felixdesouza:
I commented https://github.com/palantir/sls-packaging/pull/1693/files#r1757240550 but not sure if you saw it.

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/ExtractFileFromJar.java
CRogers: Done!

felixdesouza (approved):

PR #1703, opened 2024-09-24
Add Foundry Product Type

jakobjuelich:
## Before this PR
The createManifest task could only be used to create services, daemons, helm charts, or assets.

## After this PR

==COMMIT_MSG==
Add Foundry Product Type

Foundry Products were recently introduced as a concept in Apollo. With this PR, we can create products with this product type using the sls packaging plugin.
==COMMIT_MSG==

## Possible downsides?

This PR is analogous to how helm-charts were added here: https://github.com/palantir/sls-packaging/pull/1645

### DIFF
--- a/changelog/@unreleased/pr-1703.v2.yml
+++ b/changelog/@unreleased/pr-1703.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Added Foundry Product Type
+  links:
+  - https://github.com/palantir/sls-packaging/pull/1703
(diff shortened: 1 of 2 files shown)


TheKeveloper (dismissed):

iamdanfox (approved):
Looks fine to me - there is already corresponding support for `foundry-product.v1` in apollo-catalog.

Ricky-N (approved):
🎉

esword (approved):

PR #1706, opened 2024-09-25
Hybrid (G1) GC uses the defeault 200ms MaxGCPauseMillis on JDK-21+

carterkozak:
It seems that the heuristics have gone a bit sideways in JDK-21 causing degenerate full gc pauses unnecessarily when we configure `-XX:MaxGCPauseMillis=500`, however setting the default 200ms value resolves this behavior.

Note that we deployed the increase from 200ms to 500ms at a time when container cpu shares informed the JDKs processor count, thus gc threads/etc, which is no longer the case. As such, it should be safe (and generally more stable) to use the hardened default values from the jdk.

==COMMIT_MSG==
Hybrid (G1) GC uses the defeault 200ms MaxGCPauseMillis on JDK-21+
==COMMIT_MSG==

## Possible downsides?
Performance could change in unexpected ways!

## Alternatives:

We could change the default across the board, however it's a bit safer to apply this more precisely to the java version where our previous default value causes problems.

We could slowly ratchet the value down over time, however that makes it more difficult to root cause changes, as they would be less abrupt. We're currently rolling this out explicitly (via the configuration dsl) in a subset of services of varying sizes to validate that metrics look the same or better before rolling this out.

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/gc/GcProfile.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/gc/GcProfile.java
@@ -103,13 +103,13 @@ public interface GcProfile extends Serializable {
     // Match the MaxGCPauseMillis case
     @SuppressWarnings("AbbreviationAsWordInName")
     class Hybrid implements GcProfile {
-        // Applies to jdks older than 21:
-        // We use 500ms by default, up from the JDK default value of 200ms. Using G1, eden space is dynamically
-        // chosen based on the amount of memory which can be collected within the pause time target.
-        // Higher pause target values allow for more eden space, resulting in more stable old generation
-        // in high-garbage or low-gc-thread scenarios. In the happy case, increasing the pause target increases
-        // both throughput and latency. In degenerate cases, a low target can cause the garbage collector to
-        // thrash and reduce throughput while increasing latency.
+        // For Java versions older than 21 we use 500ms by default, up from the JDK default value of 200ms for
+        // historical reasons. Java 21 builds seem to interact poorly with this value, causing frequent full
+        // GC cycles. The 500ms value was chosen at a time when GC thread counts were based on cpu shares
+        // rather than the underlying physical cores (prior to https://bugs.openjdk.org/browse/JDK-8281181),
+        // so services tended to thrash in GC to meet the 200ms threshold, promoting data too quickly to old gen.
+        // This is no longer the case, however we are leaving the Java 17 values where they've been, and making
+        // changes targeting Java 21 and beyond where we've observed significant improvements.
         private static final int LEGACY_MAX_GC_PAUSE_MILLIS = 500;
 
         private OptionalInt maxGCPauseMillis = OptionalInt.empty();
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/gc/GcProfile.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/gc/GcProfile.java
@@ -22,6 +22,7 @@ import com.google.common.collect.ImmutableMap;
 import java.io.Serializable;
 import java.util.Collections;
 import java.util.List;
+import java.util.OptionalInt;
 import org.gradle.api.JavaVersion;
 
 public interface GcProfile extends Serializable {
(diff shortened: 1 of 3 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/gc/GcProfile.java
```diff
@@ -108,15 +109,30 @@ class Hybrid implements GcProfile {
...
+        private static final int JAVA_21_PLUS_MAX_GC_PAUSE_MILLIS = 200;
+
+        private OptionalInt maxGCPauseMillis = OptionalInt.empty();
 
         @Override
-        public final List<String> gcJvmOpts(JavaVersion _javaVersion) {
-            return ImmutableList.of("-XX:+UseG1GC", "-XX:+UseNUMA", "-XX:MaxGCPauseMillis=" + maxGCPauseMillis);
+        public final List<String> gcJvmOpts(JavaVersion javaVersion) {
+            return ImmutableList.of(
+                    "-XX:+UseG1GC", "-XX:+UseNUMA", "-XX:MaxGCPauseMillis=" + getMaxGCPauseMillis(javaVersion));
```
carterkozak: Perhaps it would be best to exclude this parameter entirely when it's not specified? That way if the default changes in the jdk, we follow it. I don't have a strong preference -- we could update this once we drop support for jdk17 to exclude the value unless specified.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPluginTests.groovy
```diff
@@ -1300,6 +1300,76 @@ class JavaServiceDistributionPluginTests extends GradleIntegrationSpec {
...
+        runTasks(':untar')
+
+        then:
+        def actualStaticConfig = OBJECT_MAPPER.readValue(
+                new File(projectDir, 'dist/service-name-0.0.1/service/bin/launcher-static.yml'), LaunchConfig.LaunchConfigInfo)
+        actualStaticConfig.jvmOpts.containsAll(['-XX:+UseG1GC', '-XX:+UseNUMA', '-XX:MaxGCPauseMillis=1234'])
+    }
+
+    def 'gc profile null configuration closure on java 21'() {
+        // We set a different different default MaxGCPauseMillis in JDK-21 than older JDKs.
```
mpritham: nit: I think we don't set the MaxGCPauseMillis in JDK-21

carterkozak: good call -- I updated the implementation but neglected to reflect that in the comment!

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/gc/GcProfile.java
```diff
@@ -102,21 +103,42 @@ public final void newRatio(int newerRatio) {
     // Match the MaxGCPauseMillis case
     @SuppressWarnings("AbbreviationAsWordInName")
     class Hybrid implements GcProfile {
+        // Applies to jdks older than 21:
         // We use 500ms by default, up from the JDK default value of 200ms. Using G1, eden space is dynamically
         // chosen based on the amount of memory which can be collected within the pause time target.
         // Higher pause target values allow for more eden space, resulting in more stable old generation
         // in high-garbage or low-gc-thread scenarios. In the happy case, increasing the pause target increases
         // both throughput and latency. In degenerate cases, a low target can cause the garbage collector to
         // thrash and reduce throughput while increasing latency.
```
bmarcaur: This comment is slightly misleading as its not the "default" and it might help to also throw in some information about why this is different surrounding the 21 boundary given that we were using `G1GC` before and after the upgrade.

carterkozak: Good call, I've replaced it with something that provides better context.

gradle-sls-packaging/src/test/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPluginTests.groovy
```diff
@@ -1300,6 +1300,76 @@ class JavaServiceDistributionPluginTests extends GradleIntegrationSpec {
         actualStaticConfig.jvmOpts.containsAll(['-XX:+UseG1GC', '-XX:+UseNUMA', "-XX:MaxGCPauseMillis=500"])
```
schlosna, line 1301: Should we update this test to use `javaVersion 17` to explicitly test JDK 17 and avoid breaking when default `javaVersion` for the project bumps to JDK 21?

carterkozak: good idea, updated!

pkoenig10:
@carterkozak are you still planning to push this through?

carterkozak:
Many services appear to perform worse with the default 200ms value than our override, I think we scrap this for now. Probably best to roll back the config in mp for consistency

PR #1709, opened 2024-10-04
remove accidental connection between task outputs

esword:
## Before this PR
If a project has both the com.palantir.recommended-product-dependencies plugin and one of the distribution plugins (asset or service) then gradle8+ will fail with an error like this:

```
 Reason: Task 'processResources' uses this output of task 'resolveProductDependencies' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
```

The reason is that both the resolveProductDependencies and the compileRecommendedProductDependencies tasks would write into a directory called build/product-dependencies.  The RecommendedProductDependenciesPlugin adds that directory as a resources source.  It properly adds compileRecommendedProductDependencies as a dependency of processResources.  The fact that resolveProductDependencies gets involved is a coincidence - we just happened to name it's output the same thing even though it's in a different project and plugin.

The two plugins should never be applied to the same project - something shouldn't be both an API jar with recommended dependencies and an SLS distribution that consumes them.  But many repos do have both applied (likely accidentally), so this just makes it clear that the two outputs have nothing to do with each other.  I may do another PR to enforce this (like we do for the Asset and Service plugins), but I want to get a release out that solves the problem first because it's blocking  some repos from upgrading to gradle8.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1709.v2.yml
+++ b/changelog/@unreleased/pr-1709.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: remove accidental connection between task outputs
+  links:
+  - https://github.com/palantir/sls-packaging/pull/1709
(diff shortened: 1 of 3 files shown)


carterkozak (dismissed):
lgtm.

Not a change for this PR necessarily, but I wonder if we should try to map out our expected file structure within the build directory, perhaps namespaced using the plugin, then task or extension name, etc.
That could also make the contents of the build director a bit more obvious, beyond helping us avoid clobbering

carterkozak (approved):

PR #1739, opened 2024-11-14
use new file permissions API

esword:
## Before this PR
the setFileMode of copy task is deprecated for gradle9.  Using it causes gradle deprecation warnings which cause any nebula integration tests to fail.

Note that this repo still has the ignoreDeprecations flag set to true because tests use the GCV plugin which also has gradle9 issues.  But at least this plugin will not directly cause tests in other repos to have deprecation issues.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


esword:
sigh... the new API is only available in >= 8.3 so this won't work while we support gradle7.  will need to do a switch to determine what version we are running.

PR #1762, opened 2025-01-27
Add getProductDependenciesFile() to CompileRecommendedProductDependencies

crogoz:
## Before this PR

Follow-up for: https://github.com/palantir/sls-packaging/pull/1760
Some internal consumers were expecting the `output.getSingleFile()` to be the `product-dependencies` file, with the change in https://github.com/palantir/sls-packaging/pull/1760, the output became a directory. 

## After this PR
Fixes: https://pl.ntr/2qh
Adds a method `getProductDependenciesFile()` that will surface the product-files that can be used by the consumers. 
==COMMIT_MSG==
Add getProductDependenciesFile() to CompileRecommendedProductDependencies 
==COMMIT_MSG==

### DIFF
--- a/gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/CompileRecommendedProductDependencies.java
+++ b/gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/CompileRecommendedProductDependencies.java
@@ -30,7 +30,6 @@ import org.gradle.api.tasks.TaskAction;
 
 public abstract class CompileRecommendedProductDependencies extends DefaultTask {
 
-
     public static final String RESOURCE_PATH =
             RecommendedProductDependencies.SLS_RECOMMENDED_PRODUCT_DEPS_KEY + "/product-dependencies.json";
     static final ObjectMapper MAPPER = new ObjectMapper();
@@ -42,12 +41,11 @@ public abstract class CompileRecommendedProductDependencies extends DefaultTask 
      * Ensure that the sourcesJar task in {@link RecommendedProductDependenciesPlugin} includes {@code getProductDependenciesFile()}.
      */
     @OutputDirectory
-    abstract DirectoryProperty getOutputDirectory();
-
+    abstract DirectoryProperty getOutputDir();
 
     @Internal
     public final File getProductDependenciesFile() {
-        return getOutputDirectory().file(RESOURCE_PATH).get().getAsFile();
+        return getOutputDir().file(RESOURCE_PATH).get().getAsFile();
     }
 
     @TaskAction
--- a/gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/CompileRecommendedProductDependencies.java
+++ b/gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/CompileRecommendedProductDependencies.java
@@ -17,13 +17,16 @@
 package com.palantir.gradle.dist;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import java.io.File;
 import java.io.IOException;
+import javax.inject.Inject;
 import org.gradle.api.DefaultTask;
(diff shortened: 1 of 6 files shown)


gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/CompileRecommendedProductDependencies.java
```diff
@@ -17,36 +17,46 @@
...
+
+    public static final String RESOURCE_PATH =
+            RecommendedProductDependencies.SLS_RECOMMENDED_PRODUCT_DEPS_KEY + "/product-dependencies.json";
     static final ObjectMapper MAPPER = new ObjectMapper();
 
     @Input
     abstract SetProperty<ProductDependency> getRecommendedProductDependencies();
 
+    @OutputFile
+    abstract RegularFileProperty getOutputFile();
```
CRogers: I don't even think we need to make this an `@OutputFile` - could just make it a helper method on the class to get the output.

CRogers: Actually, there is one possible benefit: the dev-env code that calls `outputs.singleFile` would fail harder with the `@OutputFile` setup I think that just a helper method.

gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/CompileRecommendedProductDependencies.java
```diff
@@ -17,36 +17,46 @@
...
+
     @OutputDirectory
-    abstract DirectoryProperty getOutputDir();
+    abstract DirectoryProperty getOutputDirectory();
```
CRogers: We should explain in a comment why this needs to be an `@OutputDirectory` - it confused me at first. We should also have a test that a task dependency is set up properly between the source jar and this task. Probably should have had them in [the previous PR](https://github.palantir.build/gradle-plugins/dev-env/pull/3546), but will do to have them here.

gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/RecommendedProductDependenciesPlugin.java
```diff
@@ -25,9 +25,6 @@
 
 public class RecommendedProductDependenciesPlugin implements Plugin<Project> {
 
-    public static final String RESOURCE_PATH =
-            RecommendedProductDependencies.SLS_RECOMMENDED_PRODUCT_DEPS_KEY + "/product-dependencies.json";
```
CRogers: This is used in at least this test in another repo: https://github.com/palantir/gradle-conjure/blob/947daccbc62a144195cca972a831882654aff10a/gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPluginIntegrationSpec.groovy#L211. I'd probably just keep it where it is to avoid any breaks.

CRogers (approved):

gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/CompileRecommendedProductDependencies.java
```diff
@@ -33,20 +35,25 @@ public abstract class CompileRecommendedProductDependencies extends DefaultTask
...
+     */
     @OutputDirectory
     abstract DirectoryProperty getOutputDir();
 
-    @TaskAction
-    final void action() throws IOException {
-        File outputFile = getOutputDir()
-                .get()
+    @Internal
+    public final File getProductDependenciesFile() {
```
pkoenig10, line 45: This should probably be a `Provider`, no?

Also, any reason not to make this an `@OutputFile`? That way Gradle will automatically create the parent directories.

PR #1767, opened 2025-02-03
remove output dir

esword:
## Before this PR
TODO

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


pkoenig10:
Why do we want to do this? `OutputDirectory` is simply the more correct thing to use here and doesn't require this parent directory hack.

PR #1772, opened 2025-02-13
Add enableAlwaysPreTouch option

parker89:
## Before this PR

Add enableAlwaysPreTouch option which enables the `-XX:+AlwaysPreTouch` JVM flag

Plan is to slowly roll this out to the fleet via excavator and eventually make it the default. 

Part of the java hugepages plan: https://palantir.quip.com/c66fAvkijzKX

## After this PR

==COMMIT_MSG==
Add enableAlwaysPreTouch option
==COMMIT_MSG==


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java
```diff
@@ -148,6 +151,14 @@ public final void enableManifestClasspath(boolean newEnableManifestClasspath) {
         this.enableManifestClasspath.set(newEnableManifestClasspath);
     }
 
+    public final Provider<Boolean> getEnableAlwaysPreTouch() {
+        return enableAlwaysPreTouch;
+    }
+
+    public final void enableAlwaysPreTouch(boolean alwaysPreTouch) {
```
carterkozak: I recall that gradle does something special with `getEnableAlwaysPreTouch` and `setEnableAlwaysPreTouch` which impacts the way methods can be called.

I think `void enableAlwaysPreTouch(boolean alwaysPreTouch)` allows us to use `enableAlwaysPreTouch true` where `setEnableAlwaysPreTouch` would allow us to use `enableAlwaysPreTouch = true`

carterkozak: The build folks would have a better idea what best practice looks like for these things than I do

parker89: I added both types of setter - seems like some of the other options do this. I'll check the some build folks too

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java
```diff
@@ -148,6 +151,18 @@ public final void enableManifestClasspath(boolean newEnableManifestClasspath) {
         this.enableManifestClasspath.set(newEnableManifestClasspath);
     }
 
+    public final Provider<Boolean> getEnableAlwaysPreTouch() {
```
CRogers: Make this a `Property<Boolean>` then remove the getter and setter below. Gradle will use this `get...` property to make `enableAlwaysPreTouch = true` work.

CRogers: I might consider renaming this to property to `alwaysPreTouch` then having a void method `enableAlwaysPreTouch`. This means you can excavate out

```gradle
distribution {
    enableAlwaysPreTouch()
}
```

rather than `alwaysPreTouch = true`, which is a bit nicer/more in line with some of our other options in various plugins.

parker89: I updated it to use the `enableAlwaysPreTouch()` style setter and changed the name of the property to `alwaysPreTouch`. How does it look now?

carterkozak (dismissed):
lgtm, thanks!

readme.md
```diff
@@ -265,6 +265,7 @@ And the complete list of configurable properties:
    Java 9 or higher will default to `$JAVA_<majorversion>_HOME` ie for Java 11 this would be `$JAVA_11_HOME`.
  * (optional) `gc` override the default GC settings. Available GC settings: `throughput` (default for Java 14 and lower), `hybrid` (default for Java 15 and higher) and `response-time`. Additionally, there is also `dangerous-no-profile` which does not apply any additional JVM flags and allows you to fully configure any GC settings through JVM options (not recommended for normal usage!).
  * (optional) `addJava8GcLogging` add java 8 specific gc logging options.
+ * (optional) `enableAlwaysPreTouch` adds the `-XX:+AlwaysPreTouch` JVM option.
```
carterkozak: unclear if we should document this as `alwaysPreTouch` or `enableAlwaysPreTouch()`

gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java
CRogers: looks good

readme.md
parker89: I was debating this too. The current docs don't differentiate between the different potential usage styles currently. eg `enableManifestClasspath true` vs `enableManifestClasspath = true`

parker89: I'll add the `()` to make it most clear - can't hurt

carterkozak (approved):

PR #1787, opened 2025-03-19
AlwaysPreTouch option includes UseTransparentHugePages

parker89:
## Before this PR

The enableAlwaysPreTouch() option will include the `+UseTransparentHugePages` JVM opt.

If the OS enables THP [madvise], this option will tell the JVM to start using madvise to use THP. 
If the OS enables THP [always], the JVM is already using THP in the background, and this option will tell the JVM to start using madvise to use THP instead, which is preferable
If the OS does not enable THP, this option will have no effect. 

`+UseTransparentHugePages` is bundled with `+AlwaysPreTouch` since using THP w/o AlwaysPreTouch can lead to increased memory fragmentation. 

## After this PR

==COMMIT_MSG==
The enableAlwaysPreTouch() option will include the `+UseTransparentHugePages` JVM opt.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1787.v2.yml
+++ b/changelog/@unreleased/pr-1787.v2.yml
@@ -1,0 +1,6 @@
+type: feature
+feature:
+  description: The enableAlwaysPreTouch() option will include the `+UseTransparentHugePages`
+    JVM opt.
+  links:
+  - https://github.com/palantir/sls-packaging/pull/1787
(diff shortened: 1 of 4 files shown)


carterkozak (approved):
Thanks! Adding my +1 here since we'd discussed this at length a couple weeks ago, no need to block on re-hashing.

PR #1821, opened 2025-06-27
Remove `isValid` call in `ProductDependency` and removed ConfigureProductDependenciesTask for configuration cache compatibility

FinlayRJW:
## Before this PR
When trying to fix other plugins' configuration cache issues, we kept hitting the following problem:

```
Configuration cache state could not be cached: field `__recommendedProductDependencies__` of task `:task:compileRecommendedProductDependencies` of type `com.palantir.gradle.dist.CompileRecommendedProductDependencies`: error writing value of type 'org.gradle.api.internal.provider.DefaultSetProperty'
> minimumVersion must be specified
```

It turns out this was triggered by the `ProductDependency` class calling `isValid` in the constructor, which caused validation to occur too early during configuration time, leading to serialization issues with Gradle's configuration cache.

A further problem was found in `ConfigureProductDependenciesTask` when trying to make configuration cache tests for these changes -> same issue as seen here [gradle-conjure](https://github.com/palantir/gradle-conjure/pull/1637/files#diff-257e5d4a39b7cde343dac08bd16d9ef30c0299543aecfd888f48ba24b8c1dc93R205):

```
org.gradle.api.InvalidUserCodeException: Invocation of 'Task.project' by task ':configureProductDependencies' at execution time is unsupported.
at org.gradle.api.DefaultTask.getProject(DefaultTask.java:59)
    at com.palantir.gradle.dist.ConfigureProductDependenciesTask.lambdaaction0(ConfigureProductDependenciesTask.java:58)
        at org.gradle.api.internal.tasks.DefaultTaskCollection$ExistingTaskProvider.configure(DefaultTaskCollection.java:217)
            at com.palantir.gradle.dist.ConfigureProductDependenciesTask.action(ConfigureProductDependenciesTask.java:53)
                at java.base/java.lang.Thread.run(Thread.java:829)
```

## After this PR
The call to `isValid` has been removed from the constructor of `ProductDependency` and a `ValidateProductDependenciesTask` has been added to ensure that validation still occurs. This resolves the initial configuration cache issues by deferring validation to a dedicated task that runs during the execution phase, avoiding serialization problems.

The `ConfigureProductDependenciesTask` has been completely removed, and its logic has been moved into a `doFirst` call in `RecommendedProductDependenciesPlugin`. This follows the pattern used in [Baseline](https://github.com/palantir/gradle-baseline/blob/1d5024af928c7c3353d03d9e4a742fd1fe5451f5/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java#L121). By using `doFirst` on the `Jar` task, we configure the manifest at execution time without accessing `Project` or other non-serializable objects in an unsupported way, ensuring compatibility with Gradle's configuration cache. The product dependencies are still resolved lazily via a captured `Provider`, maintaining the original behavior while avoiding serialization issues.

Additionally, integration tests have been updated to use Gradle's TestKit with configuration cache enabled, verifying that the plugin works correctly with the cache. A new test case has been added to confirm that invalid product dependencies cause the build to fail during the `check` task due to validation in `ValidateProductDependenciesTask`.

==COMMIT_MSG==
Remove `isValid` call in `ProductDependency` and removed `ConfigureProductDependenciesTask` for configuration cache compatibility

- Removed `isValid()` call from `ProductDependency` constructor to prevent configuration cache serialization issues during configuration time.
- Added `ValidateProductDependenciesTask` to perform validation of product dependencies during the execution phase, hooked into the `check` task.
- Completely removed `ConfigureProductDependenciesTask` and moved its manifest configuration logic into a `doFirst` action on the `Jar` task in `RecommendedProductDependenciesPlugin`.
- Captured the product dependencies `Provider` during configuration time to avoid serializing non-serializable objects like `Project`, ensuring lazy evaluation at execution time.
- Updated integration tests to use Gradle's configuration cache and added a test for invalid product dependency validation failure.
- Documented the removal of `ConfigureProductDependenciesTask` as a breaking change in RevAPI, noting it is safe as the task was not intended for external use.
==COMMIT_MSG==

## Possible downsides?
- **Breaking Change**: The removal of `ConfigureProductDependenciesTask` is a breaking change for any external users who might have depended on this task directly (although none could be found when searching). However, this task was not intended for external use, and its removal is documented in the RevAPI configuration with justification.
- **Timing Difference**: The manifest configuration now happens at the start of the `Jar` task's execution via `doFirst`, rather than in a separate preceding task. While this does not affect the final result, it could impact users who relied on the exact timing of when the dependencies were resolved.


.palantir/revapi.yml
```diff
@@ -23,3 +23,10 @@ acceptedBreaks:
     - code: "java.method.removed"
       old: "method java.io.File com.palantir.gradle.dist.CompileRecommendedProductDependencies::getProductDependenciesFile()"
       justification: "Changing a new API"
+  "7.76.0":
+    com.palantir.sls-packaging:gradle-recommended-product-dependencies:
+    - code: "java.class.removed"
+      old: "class com.palantir.gradle.dist.ConfigureProductDependenciesTask"
+      justification: "ConfigureProductDependenciesTask is not used externally so it\
```
felixdesouza, line 30: I don't think this is true? see https://pl.ntr/2v0

gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/RecommendedProductDependenciesPlugin.java
```diff
@@ -36,18 +45,56 @@ public final void apply(Project project) {
...
 
+    private void validateProductDependencies(Project project, RecommendedProductDependenciesExtension ext) {
+        TaskProvider<ValidateProductDependenciesTask> validateProductDependenciesTask = project.getTasks()
+                .register(
+                        "validateProductDependencies",
+                        ValidateProductDependenciesTask.class,
+                        task -> task.getRecommendedProductDependencies()
+                                .set(ext.getRecommendedProductDependenciesProvider()));
+
+        project.getTasks().named("check").configure(task -> task.dependsOn(validateProductDependenciesTask));
```
felixdesouza, line 60: perhaps this might be okay depending on how CI is set up, but this may not run on certain builds as the `check` task may not be run.

gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/RecommendedProductDependenciesPlugin.java
```diff
@@ -36,18 +45,56 @@ public final void apply(Project project) {
...
-                    cmt.setProductDependencies(ext.getRecommendedProductDependenciesProvider());
-                });
+        // Capture the provider during configuration time to avoid serializing non-serializable objects
+        final Provider<?> productDependenciesProvider = ext.getRecommendedProductDependenciesProvider();
 
-        // Ensure that the jar task depends on this wiring task
+        // Directly configure the Jar task's manifest using doFirst
         project.getTasks().withType(Jar.class).named(JavaPlugin.JAR_TASK_NAME).configure(jar -> {
-            jar.dependsOn(configureProductDependenciesTask);
+            jar.doFirst(new Action<Task>() {
```
felixdesouza, line 69: I think we can do a bit better here. 

The semantics are still different from before. But what I believe you want to do is, have your task create a `Manifest` and write it out to file.

Then you configure the jar task:
```java
jarTask.manifest(manifest -> manifest.from(yourManifestGeneratingTask.flatMap(YourManifestGeneratingTask::getOutputFile)));
```

felixdesouza: `Manifest#from` merges manifests from other places into the manifest for the Jar task. So you can define what you want, and then merge it in later. It's not as good as having the lazy attributes. But I think it probably works out quite well considering you want to serialise bits here and there and it's probably better as a task output.

felixdesouza:
It would be good to get a stacktrace, I'm not quite understanding why you can't do `isValid`. My initial hunch is that because it's not an immutables type, that we're not serialising things properly perhaps. But I don't know until I see more.

gradle-recommended-product-dependencies/src/main/java/com/palantir/gradle/dist/RecommendedProductDependenciesExtension.java
```diff
@@ -42,7 +42,6 @@ public final void productDependency(@DelegatesTo(ProductDependency.class) Closur
         recommendedProductDependencies.add(providerFactory.provider(() -> {
             ProductDependency dep = new ProductDependency();
             project.configure(dep, closure);
-            dep.isValid();
```
felixdesouza, line 45: surely this would be a legitimate place to put it?

felixdesouza: yeah, it would be good to get the stacktrace to see where the actual issue is.

felixdesouza: an alternative, which you're quite fortunate to have the old style way of getting things, is to do a `map(set -> set.forEach(dep::validate))`. That would mean that anyone who consumes it will have it validated already? Rather than having it done in a separate task.

felixdesouza: My issue with having it be in a separate task is that if validation is done externally, side effects may already have happened if the dependency isn't valid. It may not always be possible to run a task again without a manual clean for example. I would very much like to keep the validation as close to source as possible.

FinlayRJW: ah... I've been very silly here... the issue is not in `sls-packaging` at all - the issue is here in [gradle-conjure](https://github.com/palantir/gradle-conjure/blob/develop/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaServiceDependencies.java#L64) where we don't filter out deps without a minimum version I think I can revert everything related to `isValid`

FinlayRJW:
okay I think the real issue here (at least with isValid) lives in `gradle-conjure` so I'm going to park this for awhile and try and fix that up

PR #1841, opened 2025-07-31
MainClassResolver checks in generated sources as well

SinghKaranveer:
## Before this PR
MainClassResolver did not check generated sources for a main method meaning packages with generated main methods were incompatible. 

## After this PR

==COMMIT_MSG==
MainClassResolver checks in generated sources as well
==COMMIT_MSG==

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/util/MainClassResolver.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/util/MainClassResolver.java
@@ -47,7 +47,6 @@ public final class MainClassResolver {
                 .resolve("sources")
                 .toFile();
         if (generatedDir.exists() && generatedDir.isDirectory()) {
-            // Add all subdirectories of generated sources (e.g., annotationProcessor/java/main, etc.)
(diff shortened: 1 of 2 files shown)


gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/util/MainClassResolver.java
```diff
@@ -36,7 +38,22 @@ public static String resolveMainClass(Project project) {
...
+                .toPath()
+                .resolve("generated")
+                .resolve("sources")
+                .toFile();
+        if (generatedDir.exists() && generatedDir.isDirectory()) {
+            File[] subDirs = generatedDir.listFiles(File::isDirectory);
+            if (subDirs != null) {
+                sourceDirs.addAll(Arrays.asList(subDirs));
+            }
+        }
```
CRogers, line 54: See internal conversation about just using `distribution { mainClass 'className' }` to do this manually.

This code doesn't necessarily handle all generated code and also does not correctly pass on task dependencies from code generation tasks to consumers of this `mainClass` data (and this can be tricky to get right) - it would need some more work. I would just use the manual version for now and if other projects need to find the main class in generated we can look at getting this to work properly.

PR #1844, opened 2025-08-08
Use ZGC for hybrid profile on Java 21+

pkoenig10:
Based on internal testing, using ZGC has resulting in much improved safepoint and hiccup times, at the cost of an acceptable amount of CPU. Using ZGC is very likely a better tradeoff for the vast majority of services, especially those that run in cloud environments.

### DIFF
--- a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/gc/GcProfile.java
+++ b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/gc/GcProfile.java
@@ -71,6 +71,7 @@ public interface GcProfile extends Serializable {
                         "-XX:+ClassUnloadingWithConcurrentMark",
                         "-XX:+UseNUMA");
             }
+
             return ImmutableList.of(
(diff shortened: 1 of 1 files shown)


pkoenig10:
We have seen ZGC result in worse performance in environments with smaller heaps.

Another option is to allow sls-packaging to specify no GC (possible making that the default) and rely on [go-java-launcher](https://github.com/palantir/go-java-launcher) to choose a GC based on runtime heap size information.

This is not unprecedented - go-java-launcher [already uses cgroup memory limits to inform certain JVM options](https://github.com/palantir/go-java-launcher/blob/573a8dca1b36928a7211f938475d816f5d97f01c/launchlib/launcher.go#L375-L378). So this would just be a further extension of this existing behavior.

schlosna:
I think we should hold off on this pending some work in [go-java-launcher](https://github.com/palantir/go-java-launcher/) to better determine ZGC vs. G1 GC based on available resources, especially for more CPU & memory constrained environments & container limits

PR #1856, opened 2025-09-16
Add `module.v1` to `ProductType`

SpecificGenerality:
## Before this PR
The `createManifest` task could only be used to create services, daemons, helm charts, assets, or foundry products.

## After this PR
`createManifest` task supports creating modules.

## Possible downsides?
This PR is analogous to how Foundry products were added here: https://github.com/palantir/sls-packaging/pull/1703


nvelat:
👍

PR #1861, opened 2025-10-14
[Fix] Upgrade the "Fallback Version" to a `go-java-launcher` that is CVE Free

bmarcaur:
## Before this PR
`1.18.0` used an ancient version of golang which had a few vendored CVE's.

## After this PR
Bumps to the latest version (`1.215.0`) which is CVE free as of `10/2025`.


felixdesouza:
👍

PR #1868, opened 2025-10-23
Support Gradle 9!

### LINKED ISSUE #1862: Gradle 9 support
When moving to Gradle 9, we get this error:

```
Could not determine the dependencies of task ':service:resolveProductDependencies'.
> Could not resolve all dependencies for configuration ':service:runtimeClasspathForProductDependencies'.
   > Could not resolve project :service.
     Required by:
         project ':service'
      > A dependency was declared on configuration 'runtimeClasspathForProductDependencies' of 'project :service' but no variant with that configuration name exists.
```

This is because we're creating a configuration in a task input, which is way too late and quite an odd way to do it. Gradle 9 is getting stricter about `Configurations` and when they should exist.

## Copied from the [Gradle 9 GCV PR description](https://github.com/palantir/gradle-consistent-versions/pull/1443):

This is because sls-packaging [creates a configuration](https://github.com/palantir/sls-packaging/blob/54b7f034fc8a7c549797e5feb414866995ab12f5/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/DependencyDiscovery.java#L45) as part of calculating [task inputs for the `ResolveProductDependencies` task](https://github.com/palantir/sls-packaging/blob/288362a3d53863297f1358fc149606cf0b426430/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ProductDependencies.java#L78) which happens way too late. IMO this is bad behaviour for sls-packaging and feels correct to be an error here.

The timeline of things I think is happening in Gradle 9 is as follows:
* we do a resolution of `unifiedClasspath`
* gradle looks [for the candidates for graph variant selection](https://github.com/gradle/gradle/blob/master/platforms/software/dependency-management/src/main/java/org/gradle/internal/component/model/GraphVariantSelector.java#L98) within a project
* gradle looks [for all consumable variants for a project](https://github.com/gradle/gradle/blob/v9.1.0/platforms/software/dependency-management/src/main/java/org/gradle/internal/component/local/model/DefaultLocalComponentGraphResolveState.java#L108)
* [freezes and caches this conversion](https://github.com/gradle/gradle/blob/v9.1.0/platforms/software/dependency-management/src/main/java/org/gradle/internal/component/local/model/DefaultLocalComponentGraphResolveState.java#L98) of `Configuration` to internal format
* does not look at the `Configuration`s in `ConfigurationContainer` again i.e. [this](https://github.com/gradle/gradle/blob/v9.1.0/platforms/software/dependency-management/src/main/java/org/gradle/internal/component/local/model/DefaultLocalComponentGraphResolveState.java#L102) is never run again.

## Fix

We should just create the configuration once within the plugin configuration but not in a provider to a task that doesn't run until the task is executed

felixdesouza:
## Before this PR

As part of enabling [support for Gradle 9 in GCV](https://github.com/palantir/gradle-consistent-versions/pull/1443), there was a test that had to be disabled, because sls-packaging was just not gradle 9 compatible, independent of GCV. This PR is making it so.

sls-packaging did not support Gradle 9 due to the following:

### [Meta] Need Java 17 to run a Gradle 9 daemon
Bumped to Java 17 for this repo => Java 17 is required to use this plugin. I think this is fine, we've been minimum Java 17 for a lot of our other plugins and it's also 4 years old at this point.

[`String#stripIndent`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#stripIndent()) was added in Java 15, and it has different behaviour to the Groovy behaviour so in tests, you're forced to use 

```groovy
String#stripIndent(true /* forceGroovyBehaviour*/)
```

It manifests as really cryptic errors, but downstream of this is just poor handling of indented line files, but perhaps that's the point idk.

### `Project#exec` has been removed in Gradle 9
This was [deprecated in 8.11](https://docs.gradle.org/current/userguide/upgrading_version_8.html#deprecated_project_exec) and [removed in Gradle 9](https://docs.gradle.org/current/userguide/upgrading_major_version_9.html#removal_of_projectexec_projectjavaexec_and_script_level_counterparts). 

Simple replace with `ExecOperations`. 

It's looking a bit more difficult for scripts, as that method has also been removed there. Separate to this PR, but something to note.

### Some tests were using the non-Gradle 9 compatible version of GCV
Once I started running tests with Gradle 9, ran into familiar errors, but that was just due to GCV being super old. [Bumping to 3.3.0](https://github.com/palantir/gradle-consistent-versions/releases/tag/3.3.0) fixed this.

### `CopySpec#fileMode` has been removed
This was [deprecated in Gradle 8.8](https://docs.gradle.org/current/userguide/upgrading_version_8.html#unix_file_permissions_deprecated) and [removed in Gradle 9](https://docs.gradle.org/current/userguide/upgrading_major_version_9.html#removal_of_unix_mode_based_file_permissions)

The replacement `(Configurable)FilePermissions` was introduced [in Gradle 8.3 and stable in 8.8](https://docs.gradle.org/9.1.0/javadoc/org/gradle/api/file/FilePermissions.html).

This involves having to fallback to the old thing for Gradle 7. 

Really need to start writing that workaround library 😅 

### [Tests] `sourceCompatibility` from deprecated `java` plugin convention removed

Tests were setting `sourceCompatibility = '1.7'`, but that no longer works as the `java` convention blocks were [deprecated in Gradle 8.2](https://docs.gradle.org/current/userguide/upgrading_version_8.html#java_convention_deprecation) for removal in Gradle 9.0

This was just a simple replace to:

```gradle
java {
  sourceCompatibility = '1.7'
}
```

### `resolveProductDependencies` task cannot resolve the specially crafted `Configuration`

Problem statement in #1862, but here's the error message:

```
Could not determine the dependencies of task ':service:resolveProductDependencies'.
> Could not resolve all dependencies for configuration ':service:runtimeClasspathForProductDependencies'.
   > Could not resolve project :service.
     Required by:
         project ':service'
      > A dependency was declared on configuration 'runtimeClasspathForProductDependencies' of 'project :service' but no variant with that configuration name exists.
```

This is happening because we are creating a `Configuration` well after the time to create configurations is allowed in Gradle 9 (this is good restriction).

We were previously creating a configuration that "copied" the `Configuration` you wanted to set as the `productDependenciesConfig` inside an input property to a task and trying to use `finalizeOnRead` to stop it from being created twice 😱 

I tried to use `afterEvaluate`, which continued to work for Gradle 7/8 tests, but it was still happening too late + interactions with GCV and I wasn't super keen on `afterEvaluate`.

I settled on creating the "copy" when `setProductDependencyConfig` is called. This creates the configuration at an acceptable time for Gradle 9.

Why do we even need to "copy" this configuration?

This is because there's been no restriction on the type of configuration passed into `setProductDependencyConfig` i.e. it could be any of resolvable, declarable or consumable. As a result, we've had to hack around it to satisfy Gradle. If we actually restrict passing in anything other than a `consumable` configuration, we can get rid of the copy and everything is fine again.

That said, it *still* wasn't enough! The "copy" was both resolvable **_and_** consumable. Whilst this is fine in Gradle 7 and 8, [this is disallowed in Gradle 9](https://docs.gradle.org/current/userguide/upgrading_version_8.html#depending_on_root_configuration) as it both a resolution root and a variant!

So moving forward, the `ProductDependenciesPlugin` creates a *declarable* configuration `productDependencyDiscovery` akin to `api` and `implementation`. This is where consumers can add more dependencies that have product dependencies, or extend the configuration with another configuration.

Then for this plugin's own usage, a *resolvable* configuration that extends from `productDependencyDiscovery` and this is what is used within the `resolveProductDependencies`. 

So instead of making the "copy" of what's passed into `setProductDependencyConfig` be both resolvable and consumable, we can just drop it consumable, since we'll never resolve it directly, we'd only ever resolve the `productDependencyClasspath` configuration. The key piece is to add a direct project configuration dependency on this new *consumable* dependency. So all that's left is keeping track of the name of the new configuration, which we can store in a provider. No `afterEvaluate`, and simpler construction.

Ideally I'd like to get rid of `setProductDependencyConfig`, but there are a number of usages internally that make it hard to remove, and without this particular PR, there's no good way to do the right thing either. So this is the middle ground for the migration period. If we can't fully get rid of it, and we force the passed in configuration to be consumable, we don't need to do the "copy" and we can just depend on it directly!

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

closes #1862 

### Remove workarounds needed to support Gradle 5 😱 
Double tapping my above comment:

> Really need to start writing that workaround library 😅 

### Fix incorrect usage of `getProductDependenciesConfig` for `LaunchConfigTask`

The `LaunchConfig` task is what writes out `launcher-static.yml` and `launcher-check.yml` to be consumed by [go-java-launcher](https://github.com/palantir/go-java-launcher).

Part of that config, is specifying the classpath, and there's some code in `ModuleArgs` which extracts jvm module information from the classpath. We have to collect two types of classpath because there's an option for a classpath to be specified in a manifest jar instead. The `ModuleArgs` operates on the "realised" classpath. 

What's weird is that, when we're *not* using the manifest classpath, what we end up doing is using the `productDependenciesConfig` `Configuration` from the extension. For most cases this is actually equivalent i.e. it's the `runtimeClasspath`. However, there are cases where it's been set to something else so that excludes can be applied, or additional product dependencies can be inferred. This is incorrect as it's the wrong usage! Regardless of using the manifest classpath or not, it should "resolve" to the same set of jars, namely the `runtimeClasspath`. This just sets it up correctly. Perhaps this could be a separate PR.

## Possible downsides?

Should start a corresponding Gradle 9 errorprone migration akin to gradle-guide and configuraiton cache based changes.


felixdesouza:
Closing in favour of #1881 and smaller PRs

PR #1874, opened 2025-10-31
[Gradle 9] Java 17 + Gradle 8.14.3

felixdesouza:
## Before this PR

Started off in #1868, but that was a rather large PR with separate concerns.

### [Meta] Need Java 17 to run a Gradle 9 daemon

Bumped to Java 17 for this repo => Java 17 is required to use this plugin. I think this is fine, we've been minimum Java 17 for a lot of our other plugins and it's also 4 years old at this point. 

> N.B. it's not a strict requirement to be on Java 17 as a *plugin*, just that we need to run our tests and code with at least Java 17 for Gradle 9 to work. We get a lot of the nice benefits by moving to Java 17 source wise, so this is also an added benefit.

[`String#stripIndent`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#stripIndent()) was added in Java 15, and it has different behaviour to the Groovy behaviour so in tests, you're forced to use 

```groovy
String#stripIndent(true /* forceGroovyBehaviour*/)
```

It manifests as really cryptic errors, but downstream of this is just poor handling of indented line files, but perhaps that's the point idk.

### `Project#exec` has been removed in Gradle 9 and deprecated in 8.11
This was [deprecated in 8.11](https://docs.gradle.org/current/userguide/upgrading_version_8.html#deprecated_project_exec) and [removed in Gradle 9](https://docs.gradle.org/current/userguide/upgrading_major_version_9.html#removal_of_projectexec_projectjavaexec_and_script_level_counterparts). As we've also bumped Gradle to 8.14.3, it's now a deprecated usage, and that's failing compilation as we run with `-Werror`. It's an easy fix, so might as well include it here.

Simple replace with `ExecOperations`. This is fine because `ExecOperations` was introduced in Gradle 6 which our minimum supported Gradle version is far higher than.

It's looking a bit more difficult for scripts, as that method has also been removed there. Separate to this PR, but something to note.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

* Update to latest gradle 8 wrapper
* Run tests on Java 17
* Repo runs on Java 17
* Minimum Java version required is Java 17
* `String#stripIndent()` -> `String#stripIndent(true)`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -56,7 +56,7 @@ subprojects {
 }
 
 javaVersions {
-    libraryTarget = 11
+    libraryTarget = 17
 }
 
 jdks {
--- a/gradle-recommended-product-dependencies/src/test/groovy/com/palantir/gradle/dist/RecommendedProductDependenciesPluginIntegrationSpec.groovy
+++ b/gradle-recommended-product-dependencies/src/test/groovy/com/palantir/gradle/dist/RecommendedProductDependenciesPluginIntegrationSpec.groovy
@@ -32,7 +32,7 @@ class RecommendedProductDependenciesPluginIntegrationSpec extends IntegrationSpe
         }
         apply plugin: 'java'
         apply plugin: 'com.palantir.recommended-product-dependencies'
-        """.stripIndent()
+        """.stripIndent(true)
 
     }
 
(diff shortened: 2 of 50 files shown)


FinlayRJW:
👍

PR #1876, opened 2025-10-31
No more workspace persistence

felixdesouza:
## Before this PR
Workspace is being persisted, it's typically about 3.3GB  and takes just as long as the actual build which is wasteful. We also spend almost the same amount of restoring from the workspace as it takes to publish. It's faster to just redo the work!

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

Disable workspace persistence.


.circleci/template.sh
```diff
@@ -1,3 +1,4 @@
 #!/usr/bin/env bash
 export CIRCLECI_TEMPLATE=java-library-oss
-export JDK=11
+export JDK=17
+export PERSIST_WORKSPACE_BETWEEN_CHECK_AND_PUBLISH=false
```
FinlayRJW: nit: new line

FinlayRJW:
👍

PR #1912, opened 2025-11-19
Enable compact object headers on JDK 25+

parker89:
## Before this PR

Enable compact object headers on JDK 25+

In [JEP 519](https://openjdk.org/jeps/519), compact object headers were graduated from experimental. 

Technically compact object headers could be enabled in JDK 24, but given we don't use that JDK widely and it would require enabling experimental options, let's just enable the JVM flag for JDK 25+. 

A [future JEP](https://openjdk.org/jeps/8361187) plans to make compact object headers by default. At that point we can update sls-packaging to not add the flag for that JDK and up. 

## After this PR

==COMMIT_MSG==
Enable compact object headers on JDK 25+
==COMMIT_MSG==


bjlaub (approved):
lgtm

PR #1975, opened 2026-01-26
Fix Gradle 9 issue by using equals() instead of identity comparison in `getInRepoProductIds`

FinlayRJW:
## Before this PR

When using gradle-sls-packaging with Gradle 9, builds fail with:

`Must call this method with the root project: {project=:}`

The issue is that `getInRepoProductIds` uses identity comparison (`==`) to verify the passed project is the root project. In Gradle 9, during dependency resolution, project instances may be wrapped or decorated, causing `rootProject == rootProject.getRootProject()` to return false even when the project truly is the root project.

## After this PR

==COMMIT_MSG==
Use `.equals()` instead of identity check (`==`) for root project validation in `getInRepoProductIds`. This ensures the check works correctly in Gradle 9 where project instances may be decorated during resolution.
==COMMIT_MSG==


CRogers:
:+1:

PR #2054, opened 2026-04-17
Add product-id to product-dependencies.lock

ptaug:
## Description

Adds a `product-id: <groupId>:<artifactId>` line as the first line of the lock file:
```
product-id: com.palantir.group:service
com.palantir.foundry:foundry-api (1.234.0, 2.x.x)
```

Today, every bot that processes lock files independently resolves the owning product by parsing `build.gradle` (Java) or reading a sidecar `product.lock` (Go). This is fragile, duplicated, and costs extra GitHub API calls.

With the product ID in the lock file itself, bots can read it directly from the file they've already fetched. Existing parsers are unaffected as they match `group:name (min, max)` and will skip the `product-id:` line.

==COMMIT_MSG==
Add product-id to product-dependencies.lock
==COMMIT_MSG==

## Possible downsides?

Clients may fail to parse product-dependencies.lock. We've checked internal bots that they would be able to silently ignore the new line.


CRogers (dismissed):

CRogers:
:+1:

PR #2078, opened 2026-05-28
Add `minimumVersionFrom` to lock product-dependency minimum versions via GCV

FinlayRJW:
## Before this PR

Declaring a product dependency requires hard-coding `minimumVersion`, which then drifts from the actual version GCV resolves for the corresponding Maven coordinate. Keeping them in sync is manual and error-prone.

## After this PR

Adds a `minimumVersionFrom` option to the `distribution` extension's `productDependency` block, so the minimum version can be sourced from a locked Maven coordinate instead of being hard-coded:

```groovy
distribution {
    productDependency {
        productGroup = 'group'
        productName = 'name'
        minimumVersionFrom = 'group:name'
        maximumVersion = '1.x.x'
    }
}
```

> [!IMPORTANT]
> **Closing in favour of a GCV-side fix.** `minimumVersionFrom` does fix the product-dependency case, but it has a separate flaw. It requires building a `Provider<List<Dependency>>` from the declared `productDependencies` and locking that configuration via GCV's `versionsLock` (`lock.production { from(configurationName) }`). When `versionsLock` runs, it must evaluate that provider during **configuration time**, which in turn forces any *old-style* `getVersion('group:name')` in the product-dependency closures to run at configuration time and resolve `unifiedClasspath`

## Possible downsides?

N/A — closing unmerged.

### DIFF
--- a/gradle-recommended-product-dependencies/build.gradle
+++ b/gradle-recommended-product-dependencies/build.gradle
@@ -12,6 +12,8 @@ dependencies {
     api 'com.palantir.sls.versions:sls-versions'
     api 'com.palantir.safe-logging:preconditions'
 
+    compileOnly 'com.palantir.gradle.consistentversions:gradle-consistent-versions'
+
     testImplementation gradleTestKit()
     testImplementation 'com.netflix.nebula:nebula-test'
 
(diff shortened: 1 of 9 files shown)


FinlayRJW:
closing in favour of https://github.com/palantir/gradle-consistent-versions/pull/1664

PR #2079, opened 2026-05-28
Make service distribution runtime classpath consistent

FinlayRJW:
## Before this PR

Four places in the plugin built what should be the same runtime classpath, and had drifted:

- The manifest jar's `Class-Path` listed runtime deps before the project jar (reverse of the non-manifest order) and re-added itself, a no-op per the [JAR spec](https://docs.oracle.com/javase/8/docs/technotes/guides/jar/jar.html#classpath).
- `startScripts` (non-manifest) used `SourceSet.runtimeClasspath`, dragging in `build/classes` and `build/resources` dirs that aren't packaged into `service/lib/`. Accidental drift in [#385](https://github.com/palantir/sls-packaging/pull/385) (Groovy → Java migration): changed from [`project.configurations.runtimeClasspath`](https://github.com/palantir/sls-packaging/blob/1e68b08/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/CreateStartScriptsTask.groovy#L31) to [`sourceSets.main.runtimeClasspath`](https://github.com/palantir/sls-packaging/blob/1829651/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java#L133-L134).

## After this PR
==COMMIT_MSG==
- Project jar now appears first in the manifest `Class-Path`, matching the non-manifest ordering, and the redundant self-reference is gone.
- The manifest jar, `startScripts`, `launchConfigTask`, and `runTask` all share a single `serviceRuntimeClasspath` helper so they can't drift again.
==COMMIT_MSG==

## Possible downsides?

### DIFF
--- a/gradle-sls-packaging/src/test/java/com/palantir/gradle/dist/service/JavaServiceDistributionPluginTests.java
+++ b/gradle-sls-packaging/src/test/java/com/palantir/gradle/dist/service/JavaServiceDistributionPluginTests.java
@@ -39,10 +39,12 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.jar.Attributes;
+import java.util.jar.JarFile;
 import java.util.jar.JarOutputStream;
 import java.util.jar.Manifest;
 import java.util.zip.ZipOutputStream;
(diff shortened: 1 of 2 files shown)


gradle-sls-packaging/src/test/java/com/palantir/gradle/dist/service/JavaServiceDistributionPluginTests.java
```diff
@@ -930,8 +930,8 @@ void produces_manifest_classpath_jar_and_windows_start_script_with_no_classpath_
                 .replace("\r\n ", "");
         assertThat(zipManifest).contains("Class-Path: ");
         assertThat(zipManifest).contains("guava-19.0.jar");
-        assertThat(zipManifest).contains("root-project-manifest-classpath-0.0.1.jar");
         assertThat(zipManifest).contains("root-project-0.0.1.jar");
+        assertThat(zipManifest).doesNotContain("root-project-manifest-classpath-0.0.1.jar");
```
CRogers, line 937: Can we actually verify the order of these in the tests?

CRogers: And another test would be to generate a classpath from both using a manifest classpath and not using a manifest classpath, and ensuring they are the same.

gradle-sls-packaging/src/test/java/com/palantir/gradle/dist/service/JavaServiceDistributionPluginTests.java
```diff
@@ -929,9 +931,54 @@ void produces_manifest_classpath_jar_and_windows_start_script_with_no_classpath_
...
-        assertThat(zipManifest).contains("root-project-0.0.1.jar");
+        assertThat(zipManifest)
+                .as("the project's own jar should be listed before its runtime dependencies on the Class-Path")
+                .containsSubsequence("root-project-0.0.1.jar", "guava-19.0.jar");
+        assertThat(zipManifest).doesNotContain("root-project-manifest-classpath-0.0.1.jar");
+    }
+
+    @Test
+    void manifest_classpath_lists_jars_in_the_same_order_as_the_non_manifest_classpath(
+            GradleInvoker gradle, RootProject rootProject) throws Exception {
```
CRogers, line 942: Just to check  - did you verify this failed on the old code?

CRogers:
:+1:

gradle-sls-packaging/src/test/java/com/palantir/gradle/dist/service/JavaServiceDistributionPluginTests.java
FinlayRJW: not yet can do

FinlayRJW: yep is failing on develop

PR #2117, opened 2026-07-14
Add project-published OCI artifacts to SLS manifests

FinlayRJW:
## Before this PR

SLS distributions can declare a container image only when its URI is already known in the consuming project. There is no dependency-based way to add OCI coordinates published by another project, so integrations must resolve producer outputs themselves and risk realising values before the producing task has run.

## After this PR

SLS distributions can use `manifestOciArtifacts project(':images')` to include every published OCI coordinate from another project, or `manifestOciArtifacts project(':images'), 'bar'` to select one named artifact.

Coordinates are resolved lazily from the producer's `sls-manifest-oci-artifacts` variant, preserving producer task dependencies. Coordinates with `publish: false` are excluded, and Java producers can create coordinate documents using `OciArtifactCoordinates.of(...)`.

==COMMIT_MSG==
==COMMIT_MSG==

## Testing

- `./gradlew test`
- `./gradlew format compileAllErrorProne checkstyleMain checkstyleTest --continue -PerrorProneApply`

## Possible downsides?

None known. Projects must opt into the new `manifestOciArtifacts` dependency.

## Are Docs needed?

Yes. The README documents both dependency forms, the producer contract, and the public coordinate factory.

### DIFF
--- a/gradle-sls-packaging/src/test/java/com/palantir/gradle/dist/artifacts/ManifestOciArtifactsPluginTests.java
+++ b/gradle-sls-packaging/src/test/java/com/palantir/gradle/dist/artifacts/ManifestOciArtifactsPluginTests.java
@@ -178,7 +178,7 @@ class ManifestOciArtifactsPluginTests {
             """);
         addImage(producer, "image-a", "registry.example.com/group/image-a:1.0.0", true);
         consumer.buildGradle().append("""
-            distribution.configureArtifacts { artifact ->
+            distribution.artifacts.configureEach { artifact ->
                 artifact.uri.set(artifact.uri.get().replace('registry.example.com', 'mirror.example.com'))
             }
             """);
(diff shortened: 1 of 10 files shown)


gradle-sls-packaging/src/test/java/com/palantir/gradle/dist/artifacts/ManifestOciArtifactsPluginTests.java
```diff
@@ -0,0 +1,265 @@
...
+            distribution.artifacts
+                    .matching { artifact ->
+                        artifact.type != null &&
+                                artifact.type.getOrNull() == 'oci' &&
+                                artifact.uri != null &&
+                                artifact.uri.getOrNull() != null &&
+                                artifact.uri.getOrNull().startsWith('registry.example.com/')
+                    }
+                    .configureEach { artifact ->
+                        artifact.uri.set(artifact.uri.get().replaceFirst(
```
felixdesouza: what if you tried with doing a map

FinlayRJW: so initially I didn't think this would work - but if you drop the matching and roll it all into the map it seems to work see [b4e3528](https://github.com/palantir/sls-packaging/pull/2117/commits/b4e35287bd37417e22434b590e39b5abde030243)

FinlayRJW: sorry it's not dropping matching that fixed it but removing the eager getOrNull()/set()  matching only had to go because it must return a boolean

PR #2124, opened 2026-07-31
Revert "Increase default JFR max chunk size to 40m"

parker89:
## What

Reverts #2104, restoring the always-on JFR flag to `-XX:FlightRecorderOptions=stackdepth=256`. The JFR max chunk size goes back to the JVM default of 12m.

The increased chunk size contributed to increased memory in two ways: 

- For services using the JFR event streaming (eg rereading JFR events back into user code to drop metrics/logs based on JFR events), the constant maps are held in heap twice (once for the current, and once for the previous). Due to these being held in heap, they were also greatly expanded (Eg a chunk size of 12mb translates into well over 40mb of heap)
- the 'tracing' segment of NMT also increased based on the chunk size


mpritham (approved):

aldexis:
@parker89 Could you add some explanation in here for posterity about why we needed to revert?

